PR #67, opened 2012-10-25
Map author slugs to user_nicename

ericmann:
Entries in the author taxonomy were originally created using the user login.  Without spaces, this was fine, but _with_ spaces caused problems.  The user with login "john doe" was mapped to the author term with the slug "john-doe" ... WordPress was unable to retrieve authors from the taxonomy if their login contained spaces.

This patch uses the `user_nicename` instead.  Functionally, it is identical to the user login but, like the slugs in the author taxonomy, replaces spaces with hyphens.  `get_term_by( 'slug', $found_user->user_nicename, $this->coauthor_taxonomy );` will return matches both for users with and without spaces in their login names.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -246,11 +246,14 @@ class coauthors_plus {
 			case 'login':
 			case 'user_login':
 			case 'email':
+			case 'user_nicename':
 			case 'user_email':
 				if ( 'user_login' == $key )
 					$key = 'login';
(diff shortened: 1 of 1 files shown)


TylerDurdon:
Hi, really like the plugin, but users with space in user_login will still not be added when using
$coauthors_plus->add_coauthors.
Was there a workaround or is there a fix?
Thx & Regards

PR #124, opened 2013-06-27
fixed broken author+term query

wluo:
when co-author-plus plugin is enabled, we noticed that we were not getting results when searching for author and another taxonomy term. it looks like co-author-plus is actually not adding the join term it needs when a WP query also contains a taxonomy query, because the taxonomy query generates a join clause that's identical to what co-author-plus generates and co-author-plus cannot tell whether that clause is a duplicate or not.

the fix makes the co-author-plus-generated sql join clause slightly different (by using an alias for the wp_term_relationships table). i think this does not change the functionality of the plugin but avoids confusing sql clauses from different sources.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -544,8 +544,8 @@ class coauthors_plus {
 				return $join;
 
 			// Check to see that JOIN hasn't already been added. Props michaelingp and nbaxley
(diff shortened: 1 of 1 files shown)


trepmal:
Resolved in 6dbcd3cd48d29c7aa46a96269c733fb1f649dec6

PR #204, opened 2014-06-13
Guest author links should be filterable

thebrandonallen:
With two small, but significant, exceptions, the code in `CoAuthors_Guest_Authors:filter_author_link()` is the same as the function it's trying to filter, `get_author_posts_url()`.

The first difference, is where the `user_nicename` is set for guest authors. Without it, guest author links aren't appropriately created.

The other difference, is the lack of the `author_link` filter. Anyone trying to manipulate links returned by `get_author_posts_url()` or use a non-standard rewrite tag in WP_Rewrite::author_base, will find that they're filters fail. Found this while try to track down an issue a user was having with my [Edit Author Slug](https://github.com/thebrandonallen/edit-author-slug/commits/peter-debug) plugin.

This pull request keeps us DRY, and maintains the `author_link` filter on guest authors.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -149,7 +149,7 @@ class CoAuthors_Guest_Authors
 		global $post;
 
 		$guest_author = $this->get_guest_author_by( 'ID', $post->ID );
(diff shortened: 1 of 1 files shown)


thebrandonallen:
Closed in favor of #258

PR #236, opened 2014-11-20
Fix #215 "Cannot remove the last author once post is saved"

### LINKED ISSUE #215: Cannot remove the last author once post is saved
If you assign a co-author to a post, save it, and reload the page, the Remove link disappears. You can no longer remove that author.

paulschreiber:
Problem: "If you assign a co-author to a post, save it, and reload the page, the Remove link disappears. You can no longer remove that author."

Fix: Remove the comment // Hide the delete button if there's only one co-author and the two lines of code that follow. The lines in question hide the "X" button. Removing those lines means the X button stays, allowing users to remove the last author.

### DIFF
--- a/js/co-authors-plus.js
+++ b/js/co-authors-plus.js
@@ -16,10 +16,6 @@ jQuery(document).ready(function () {
 		var $coauthor_row = jQuery(elem).closest('.coauthor-row');
 		$coauthor_row.remove();
 
-		// Hide the delete button when there's only one Co-Author
-		if ( jQuery( '#coauthors-list .coauthor-row .coauthor-tag' ).length <= 1 )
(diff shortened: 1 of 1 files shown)


paulschreiber:
@danielbachhuber @joshbetz Can we include this fix? I received another request for it today.

bcampeau:
I think @danielbachhuber is right that it probably needs a unit test for this first. This particular update also  enables this feature for sites that may want it to continue to work as it does now. It would make more sense to me to do it as a filterable option.

joshbetz:
A filterable option sounds good. In any case, I think the delete button should still be hidden if there are no coauthors.

bcampeau:
@paulschreiber This script needs to be localized with the value of "force guest authors" and then this check can be disabled only in the scenario where that is false.

GaryJones:
Closing for lack of response.

PR #241, opened 2014-12-17
Give user capabilty if they don't have it

LinuxBozo:
We should add the capability if the user *does not* have it already (`empty`), 
not re-add it if they already do (`! empty`).

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1282,10 +1282,10 @@ class coauthors_plus {
 
 		$current_user = wp_get_current_user();
 		if ( 'publish' == get_post_status( $post_id ) &&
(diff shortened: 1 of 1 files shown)


joshbetz:
@LinuxBozo Can you rebase so we can merge this? Sorry it's taking so long.

LinuxBozo:
@joshbetz Done..

bcampeau:
The logic here is actually incorrect. It's checking to see if the user already has the permission to edit published or private posts. If they do, it will also grant them the same permission on posts where they are guest authors.

However, this update would grant ANY user the ability to edit published or private posts where they are a guest author even if they didn't normally have that permission. This could have unintended consequences in loosening permissions within the admin dashboard. Unfortunately, we can't merge this.

Issue #201, opened 2014-05-09
Guest Author Page


MikeNGarrett:
/author/[author_slug] is no longer working for guest authors. The template is returning a 404.

maxfenton:
Running into the same problem.

maxfenton:
I haven't been able to test @matthewspencer's fix, so I basically reverted https://github.com/Automattic/Co-Authors-Plus/commit/2ce2ce88f75d848440da2395cd3bbe375bdac537 by making a custom function in my theme `functions.php` with a modified version of the older `fix_author_page`

```
remove_filter( 'posts_selection', array( $coauthors_plus, 'fix_author_page' ) );
remove_action( 'the_post', array( $coauthors_plus, 'fix_author_page' ) );
add_filter( 'posts_selection', 'mf_fix_author_page' ); 
add_action( 'the_post', 'mf_fix_author_page' );
function mf_fix_author_page() { /* ... */ }
```

rebeccahum:
@MikeNGarrett  Are you still experiencing this issue? Have you tried @matthewspencer's fix?  That wp cli command should update the guest author's posts with the correct amount.

MikeNGarrett:
This was so long ago that I'm not sure the context of the original issue. I know we found a way to work around the issue way back then, but I'm not sure if the original issue is fixed.

jqueryalmeida:
Solution:
`function co_author_archive($query) {
    if ($query->is_author)
        $query->set( 'post_type', array('custom_post', 'post') );
    remove_action( 'pre_get_posts', 'co_author_archive' );
}
add_action('pre_get_posts', 'co_author_archive');
`

rebeccahum:
Hi @jqueryalmeida, thanks so much for your contribution. Would you be able to submit a PR for this?

trepmal:
I think this is resolved in #440.

PR #242, opened 2014-12-17
Removing check for is_user_member_of_blog 

bcampeau:
I've discussed this issue a few times, but this line of code really breaks the spirit of Co-Authors Plus. Guest authors don't have to be users, so why should they be required to be a member of the blog to have an author archive page? I also have legit use cases (i.e. NY Post sites) where this breaks their author pages.


maxfenton:
+1 — the requirement that the user is a member of the blog broke the logic on the author pages of a site I'm working on. Any chance this (or other) pull requests will get merged anytime soon? Looks like this has been in the queue for many months.

maxfenton:
@natebot Thank you so much. I'm new to this plugin (and community), so I'm delighted to see development is alive and well. Stoked to see this pull request after looking at issue #201.

joshbetz:
From: 2ce2ce8

> The author archive should be displayed when the user is a member of the blog, or the user has a previously published post on the blog.

I think we have to keep this. If a user is not a member of a site and doesn't have any posts, they shouldn't have an author page. Think about big multisite environments like WordPress.com. To use the NY Post example, http://nypost.com/author/betzster should be a 404 page rather than an empty author page.

maxfenton:
@joshbetz That makes perfect sense. I was having missing pages for guest authors with published posts, so maybe my problem lies somewhere else.

mattoperry:
Some thoughts:

- sometimes it's undesirable for guest authors to have an author page ... for example, if a public figure writes a one-time op-ed on a site, it's not clear that there should be an author archive for that author.
- on the other hand, if a guest author (even if not a member) has 1000 posts or something, you'd kind of expect an author page.
- we should avoid empty archive pages ... archives for normal WP authors with no posts 404 .. seems like the same thing should happen here (perhaps it already does)
- @joshbetz is there anything in particular you're worried about breaking if this is introduced?  One thing to note is that this is already merged on .com, just not in the public version of the plugin.

Just spoke to @natebot and we'd like to bump this from the release we're doing this week since it's just meant to be a maintenance/sync release, and we should probably a) only include very safe changes and b) consider this a bit more.  Doing this will allow us to sync .com and github, with the exception of this change, and then move on to a larger release that could include this in the coming weeks.

PR #246, opened 2014-12-23
fix "Inline control structures are not allowed" errors

paulschreiber:
required by WordPress coding standard


paulschreiber:
@joshbetz

PR #248, opened 2014-12-26
Fix bug where existing guest author cannot update or map to an existing ...

hauvong:
...wp user account because the same user_login already exist.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -688,12 +688,25 @@ class CoAuthors_Guest_Authors
 			wp_die( __( 'Guest authors cannot be created without display names.', 'co-authors-plus' ) );
 		$post_data['post_name'] = $this->get_post_meta_key( $slug );
 
+		// get linked account for validation
+		if ( !empty( $original_args['cap-linked_account'] ) ) {
+			$user = get_user_by( 'id', $original_args['cap-linked_account'] );
+			if ( !empty( $user ) ) {
+				$linked_account = $user->user_login;
+			}
+		}
+
+		// No link account passed, check existing linked account
+		if ( empty( $linked_account) ) {
+			$linked_account = get_post_meta( $original_args['ID'], $this->get_post_meta_key( 'linked_account' ), true );
+		}
+
 		// Guest authors can't be created with the same user_login as a user
 		$user_nicename = str_replace( 'cap-', '', $slug );
 		$user = get_user_by( 'slug', $user_nicename );
 		if ( $user
 			&& is_user_member_of_blog( $user->ID, get_current_blog_id() )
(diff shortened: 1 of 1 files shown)


joshbetz:
Confirmed that this is already working as intended, per https://github.com/Automattic/Co-Authors-Plus/commit/91c9df5e85329464491bfe4103a6d6b7a9b6482c.

> Guest author user_logins can't be the same as users. Instead, they'll need to "Create Profile" from the user

hauvong:
I think the issue is a bug, unless this is intentional to require manual fixing from these two cases:

Case #1: un-mapped existing guest author from Wordpress user
1. Edit an existing guest author with WordPress User Mapped.
2. Remove the WordPress User Mapping by select "-- Not mapped --"
3. Save/Update the guest author
4. Re-edit same author
5. Save/Update the guest author

Case #2: Create guest author before Wordpress user 
1. Create a new guest author with display name "test"
2. Create a new Wordpress user with name "test"
3. Edit guest author and map the "test" guest author with Wordpress user
4. Save/Update the guest author

Both cases are throwing error: "Guest authors cannot be created with the same user_login value as a user. Try creating a profile from the user on the Manage Users listing instead."
There is no way to fix the record at this point without deleting the guest author via the wp-admin UI.

joshbetz:
In case anybody else comes across this issue, the "Create Profile" link needs to e enabled with `coauthors_show_create_profile_user_link `. https://github.com/Automattic/Co-Authors-Plus/blob/c16da410f75036719215128c6a7fb1d37fa28a80/readme.txt#L70

benlk:
I've implemented that filter as a cheap plugin, which can be saved as `wp-content/plugins/coauthors_show_create_profile_user_link/coauthors_show_create_profile_user_link.php`:

``` php
<?php
/*
Plugin Name: Co-Authors Plus Show Create User Profile Link
Plugin URI:
Description: re-shows the missing "Create Profile" link
Version: 0.0.0
Author: Ben Keith of INN Labs
Author URI: https://labs.inn.org/
License: GPL Version 2 or later
License URI: https://www.gnu.org/licenses/gpl-2.0.html
*/

/**
 * Enable the "Create Profile" link in the list of WP users
 *
 * @link https://github.com/Automattic/Co-Authors-Plus/pull/248#issuecomment-103215238
 * @link https://developer.wordpress.org/reference/functions/__return_true/
 * @link https://github.com/Automattic/Co-Authors-Plus/blob/3.4.2/php/class-coauthors-guest-authors.php#L1415-L1417
 */
add_filter( 'coauthors_show_create_profile_user_link', '__return_true' );
```

After activating this plugin, I was able to create new Guest Authors from the list of WordPress users, with the correct mapping between each.

PR #254, opened 2015-01-30
Introduce a `coauthors_display_name` filter

johnbillion:
Introduce a `coauthors_display_name` filter so the final display name output for each author can be filtered. Passes the `$author` object for context.


pippercameron:
This was fixed downstream for WordPress.com VIP 12/17/14. Merged by @natebot in e9439a5. Closing this out.

PR #255, opened 2015-02-04
Default to the list of supported post types instead of just "post" when updating term counts

xmattus:


bcampeau:
@xmattus this is failing all unit tests. Is it possible that this is running when `$this->supported_post_types` isn't set yet? If not, this should include a unit test update. Thanks.

xmattus:
I've been informed the issue this was intended to address was fixed in another update anyway, so I'm cool to close this out.

PR #256, opened 2015-03-14
Fixed a bug in coauthors__echo() when adding separators

jamiebergen:
I found a bug when trying to use the coauthors() function as follows:
```
coauthors($between = '</div><div class="resource-meta author">', $betweenLast = '</div><div class="resource-meta author">', $before = '<div class="resource-meta author">', $after = '</div>', $echo = true);
```
The use of rtrim caused the second-to-last author name to truncate.  The updated code correctly appends separators and avoids the use of rtrim.


nickdaugherty:
Thanks for the fix.

PR #257, opened 2015-03-17
Fix get_guest_author_by returning empty objects.

### LINKED ISSUE #212: get_coauthor_by() with ID returning empty objects
On a new WordPress install, if I run `$coauthors_plus->get_coauthor_by( 'id', 1 )`, I would expect to get the coauthor object for my site admin. Instead, I get an empty guest author object.

If I disable guest authors, then I get the coauthor object.

If I'm reading it correctly, this occurs when `get_coauthor_by()` calls `get_guest_author_by()` using my original arguments. I intend to check a user ID, but `get_guest_author_by()` treats it as a post ID. `get_guest_author_by()` also constructs an object out of whatever post ID it finds, regardless of type or status.

Have I misunderstood the purpose of `get_coauthor_by()` or what it's doing?

jjeaton:
When querying by ID, get_guest_author_by only looks for whether a post_ID exists, this also adds a post_type check to ensure it's only checking within the 'guest-author' post type.

Fixes #212.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -831,7 +831,7 @@ class CoAuthors_Guest_Authors
 		switch ( $key ) {
 			case 'ID':
 			case 'id':
-				$query = $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE ID=%d", $value );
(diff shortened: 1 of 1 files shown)


jjeaton:
Thank you!

PR #258, opened 2015-03-19
Guest author links should be filterable

thebrandonallen:
Originally posted as #204, but recommitted for current master due to merge conflicts.

With two small, but significant, exceptions, the code in CoAuthors_Guest_Authors:filter_author_link() is the same as the function it's trying to filter, get_author_posts_url().
The first difference, is where the user_nicename is set for guest authors. Without it, guest author links aren't appropriately created.

The other difference, is the lack of the author_link filter. Anyone trying to manipulate links returned by get_author_posts_url() or use a non-standard rewrite tag in WP_Rewrite::author_base, will find that they're filters fail. Found this while trying to track down an issue a user was having with my Edit Author Slug plugin.


maxfenton:
+1

bcampeau:
I'm fine with line 152 to use `get_author_posts_url` inside `filter_post_updated_messages`. That makes sense to me.

The removal of most of the logic inside `CoAuthors_Guest_Authors:filter_author_link()` doesn't make sense. This would revert back to the default author link format which would definitely break some things in the plugin. Also you're calling `get_author_posts_url` which calls the `author_link` filter inside a function that's already hooking that filter, which would potentially create an infinite loop.

If you want to update this to only include line 152, we could definitely consider merging that. You could then hook `author_link` later than Co-Authors Plus to make any modifications that you need.

thebrandonallen:
I haven't been able to reproduce this for a while, and there's issues with my original PR. Closing.

PR #259, opened 2015-04-08
Bugfix - Use $this->args instead of local variable

dougaxe1:
Use $this->args instead of local $args. The mixing of variables caused unpredictable results in subsequent iterations.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -56,7 +56,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 		$authors = array();
 		$author_terms = array();
 
-		$args = array(
+		$this->args = array(
 				'order'            => 'ASC',
 				'orderby'          => 'ID',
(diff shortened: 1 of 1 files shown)


jjeaton:
+1 to this, using the variable instead of `$this->args` actually prevents the loop from continuing to paginate through posts meaning that only the first 100 posts get terms created.

nickdaugherty:
Commit 5fad6ace7481f28bbd4d88bffe5fb7b343e82b58 actually fixed the pagination bug, by going the other direction and from `$this->args` to a local variable.

I'd prefer to keep `$args` local (the other instances of it as a class property ideally being removed over time), as their scope really only makes sense in the command function being called. Storing them in a class member is unnecessary and I would argue can cause more unpredictable results and behavior.

If other helper functions besides the command needed access to the args, the needed arg should be passed into that function as a function argument, rather than relying on the class property.

jjeaton:
@nickdaugherty even better thanks!

PR #263, opened 2015-05-08
Fixing undefined class variable

jonathanbardo:
The wp-cli command wasn't working because of an undefined class variable. I replaced it with the local variable instead and it's working again as usual.


nickdaugherty:
Thanks for the fix :+1:

PR #267, opened 2015-06-16
MP6ify

0aveRyan:
screen shot 2015-06-15 at 11 25 02 pm

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -385,7 +385,9 @@ class coauthors_plus {
 		?>
 
 		<div id="coauthors-edit" class="hide-if-no-js">
-			<p><?php _e( 'Click on an author to change them. Drag to change their order. Click on <strong>Remove</strong> to remove them.', 'co-authors-plus' ); ?></p>
+			<?php add_thickbox(); ?>
+			<a class="button thickbox" href="<?php echo get_bloginfo('url'); ?>/wp-admin/users.php?&page=view-guest-authorsTB_iframe=true&width=1200&height=640"><small><?php _e( 'Guest Authors', 'co-authors-plus' ); ?></small></a>
+			<p><em><?php _e( 'Click to edit. Drag to order.', 'co-authors-plus' ); ?></em></p>
 		</div>
 
 		<?php wp_nonce_field( 'coauthors-edit', 'coauthors-nonce' ); ?>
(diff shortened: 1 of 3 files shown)


co-authors-plus.php
```diff
@@ -385,7 +385,9 @@ public function coauthors_meta_box( $post ) {
 		?>
 
 		<div id="coauthors-edit" class="hide-if-no-js">
-			<p><?php _e( 'Click on an author to change them. Drag to change their order. Click on <strong>Remove</strong> to remove them.', 'co-authors-plus' ); ?></p>
+			<?php add_thickbox(); ?>
+			<a class="button thickbox" href="<?php echo get_bloginfo('url'); ?>/wp-admin/users.php?&page=view-guest-authorsTB_iframe=true&width=1200&height=640"><small><?php _e( 'Guest Authors', 'co-authors-plus' ); ?></small></a>
```
bcampeau: Should be spaces inside parentheses for `get_bloginfo` to be consistent with WordPress PHP coding standards.

For safe output escaping, you should assemble the URL and then `esc_url` it. Also use `esc_html_e` instead of `_e`. It's safer.

co-authors-plus.php
```diff
@@ -385,7 +385,9 @@ public function coauthors_meta_box( $post ) {
...
+			<p><em><?php _e( 'Click to edit. Drag to order.', 'co-authors-plus' ); ?></em></p>
```
bcampeau: Use `esc_html_e` here too.

co-authors-plus.php
dryanmedia: @bcampeau thank you for the feedback!

mjangda: Since the "Guest Authors" button isn't directly related to updating the styles, can we please remove it from this PR and open a new issue to discuss its addition? (I'm not a fan of using Thickbox here.)

bcampeau: I'm not a fan of adding thickbox either. I do have concern this will potentially break more than it would "fix".

natebot: Does Thickbox support small screens well? What is the benefit of using an
overlay?

Nathan // @natebot // natebot@gmail.com

dryanmedia: lets nix thickbox but perhaps hyperlink to guest authors page instead? 

Currently the UX requires producers to click away from the page to add guests. 

Longterm would make more sense to think about a toggle of some kind (thickbox is obviously a functional hack, apologies) to quickly add new guests without clicking away from a post edit screen**. If you've tried the current flow on mobile or tablet it's particularly cumbersome.

GaryJones:
Given that the look and integration into wp-admin has moved on, and the PR has conflicts, I'm going to close this one out. If you see an opportunity to improve the design of the latest release, then fresh PRs are welcome!

PR #275, opened 2015-07-28
use Yoda conditions in comparisons

paulschreiber:


co-authors-plus.php
```diff
@@ -332,7 +332,7 @@ public function coauthors_meta_box( $post ) {
 		// @daniel, $post_id and $post->post_author are always set when a new post is created due to auto draft,
 		// and the else case below was always able to properly assign users based on wp_posts.post_author,
 		// but that's not possible with force_guest_authors = true.
-		if ( ! $post_id || $post_id == 0 || ( ! $post->post_author && ! $coauthors_plus->force_guest_authors ) || ( $current_screen->base == 'post' && $current_screen->action == 'add' ) ) {
+		if ( ! $post_id || 0 === $post_id || ( ! $post->post_author && ! $coauthors_plus->force_guest_authors ) || ( 'post' === $current_screen->base && 'add' === $current_screen->action ) ) {
```
mjangda: I'm not sure we want an strict check here. It's unlikely that `$post_id` will be a string here, but better to be safe?

mjangda:
The yoda changes are good! Although, it would have been better to separate about the strict checks into a separate PR (since they're technically changing logic).

One small comment and we can merge.

paulschreiber:
@mjangda Are we good to merge, or do you need any changes?

mjangda:
@paulschreiber We should change the `===` (strict equality check) for post_id to `==` (<a href="https://github.com/Automattic/Co-Authors-Plus/pull/275#discussion_r36014008">#</a>).

paulschreiber:
Okay, I've updated the branch/PR.

What value would cause `! $post_id` to be false but `0 == $post_id` to be true?

paulschreiber:
@mjangda @joshbetz can you merge this?

mjangda:
> What value would cause ! $post_id to be false but 0 == $post_id to be true?

Probably none, but PHP is known to do unexpectedly magical things :)

Issue #209, opened 2014-07-19
Co-Authors Plus adds too many new image sizes


kevinlisota:
The default install of Co-Authors Plus uses add_image_size to add 4 new thumbnail sizes, namely 32px, 64px, 96px and 128px, based on some common avatar sizes.

Our site does not use avatars and also has a large post/image count. We currently have 15k+ posts with 35+ images in the media library. This plugin adds 4 new image files for every image that we upload on every single post, not just author avatars, which causes a large excess of image files on our install that are not needed or used.

Add extra image sizes should be some sort of option to prevent the massive image file replication this causes.

maxfenton:
Note the `apply_filters()` being run on the list of sizes:
```
$this->avatar_sizes = array( 32, 64, 96, 128, );
$this->avatar_sizes = apply_filters( 'coauthors_guest_author_avatar_sizes', $this->avatar_sizes );
foreach ( $this->avatar_sizes as $size ) {
	add_image_size( 'guest-author-' . $size, $size, $size, true );
}
```

If your site doesn't use avatars, you could put this in your `functions.php` (for example)
```
function my_remove_coauthors_guest_author_avatar_sizes() {
  $empty_array = array();
  return $empty_array;
}
add_filter( coauthors_guest_author_avatar_sizes, my_remove_coauthors_guest_author_avatar_sizes );
```

or less elegantly (and because these aren't guaranteed to stay the same name in future version of the plugin):
```
remove_image_size( 'guest-author-32');
remove_image_size( 'guest-author-64');
remove_image_size( 'guest-author-96');
remove_image_size( 'guest-author-128');
```

Note: it might be worth adding a failsafe to the foreach in the plugin:
```
$this->avatar_sizes = array( 32, 64, 96, 128, );
$this->avatar_sizes = apply_filters( 'coauthors_guest_author_avatar_sizes', $this->avatar_sizes );
if ( !empty( $this->avatar_sizes ) ) {
	foreach ( $this->avatar_sizes as $size ) {
		add_image_size( 'guest-author-' . $size, $size, $size, true );
	}
}
```

mattoperry:
We'll be closing this for now since (as @maxfenton points out) the list of thumbnail sizes can be filtered.

PR #276, opened 2015-08-15
sync with wpcom changes

natebot:
This pulls in changes made in the http://vip-svn.wordpress.com/plugins repo. There are two commits in this PR worth noting: 

- 05c1251 adds a new avatar size, which may not be useful since subscriptions may not be supported, or even considered harmful ( see #209. )
- 6c830f0 assumes #242 which may be controversial.


mattoperry:
I'm not particularly worried about #209 for now since as @maxfenton notes the list can be filtered.  We might want to politely close that issue.

PR #285, opened 2015-09-03
Don't filter post_updated_messages when != guest author

danielbachhuber:
Otherwise, you get extra queries and PHP notices


danielbachhuber:
@mattoperry #reviewmerge — could you deploy this after the fact too?

PR #286, opened 2015-09-03
Make Travis notifications less noisy

danielbachhuber:
Only send emails on failures


danielbachhuber:
@mattoperry #reviewmerge

Issue #291, opened 2015-09-15
Notice: Trying to get property of non-object in /vagrant/wordpress/wp-content/plugins/Co-Authors-Plus/php/class-coauthors-guest-authors.php on line 157


mauteri:
Code is assuming guest author even when there are none.

castlehouse:
Looks like this was fixed in https://github.com/Automattic/Co-Authors-Plus/commit/0a19a1238e590a6ef3035edd301f4c11066b101a so should no longer be an issue.

paulschreiber:
Did this fix not make it to the WPCOM/VIP repo? I'm still seeing the warning.

stevenslack:
I am also seeing this notice. Just updated from the VIP repo as well.

PR #292, opened 2015-09-15
#291: Code is assuming guest author even when there are none.

mauteri:


php/class-coauthors-guest-authors.php
```diff
@@ -149,6 +149,11 @@ function filter_post_updated_messages( $messages ) {
 		global $post;
 
 		$guest_author = $this->get_guest_author_by( 'ID', $post->ID );
+
+		if( !$guest_author ) {
```
bcampeau: We should follow WordPress PHP coding standards for spacing: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage

PR #304, opened 2015-09-16
Return incoming filter value in filter_update_post_metadata

### LINKED ISSUE #303: Return incoming filter value in filter_update_post_metadata
https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-coauthors-guest-authors.php#L1109 should return the incoming filter value, `$retnull`, otherwise a filter that tries returning a non-null value for any post type other than CAP is overridden.

joshlevinson:
Closes #303


mjangda:
Good catch; I'm glad `null` is passed to the method otherwise this would break everything :)

PR #305, opened 2015-09-17
use esc_html_e() instead of _e()

paulschreiber:


mjangda:
Some of the strings have HTML in them (i.e. `<strong>`) and this change would encode those tags when outputting.

paulschreiber:
Good catch. I replaced a few of those calls with `wp_kses_post()`.

paulschreiber:
@mjangda

paulschreiber:
Updated to use `__()` and `wp_kses()`.

PR #308, opened 2015-09-17
use yoda condition in WP CLI

paulschreiber:
This also address a strict equality check. It's safe, as `$count` is always an integer.


paulschreiber:
@mjangda

PR #309, opened 2015-09-25
Moving coauthors_loading to global scope

bcampeau:
coauthors_loading is being referenced in `show_loading`, `hide_loading` and `move_loading` but it's being declared in `coauthors_initialize`. 

It needs to be defined globally in the script file or it will not exist in the other functions.

### DIFF
--- a/js/co-authors-plus.js
+++ b/js/co-authors-plus.js
@@ -11,6 +11,8 @@ jQuery( document ).ready(function () {
 		return false;
 	};
 
+	var $coauthors_loading;
+
 	function coauthors_delete( elem ) {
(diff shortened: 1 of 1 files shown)


bcampeau:
Not sure why this never got merged since we discussed this at our hack day, but this is a fairly significant bug and I think potentially the root cause of some issues with the autocomplete functionality. I'm going to go ahead and merge this.

pbearne:
video showing the bug

https://drive.google.com/file/d/0B5QczxsmCk1vTFIxc1BheUViT0E/view?usp=sharing

PR #311, opened 2015-10-07
add missing spaces in german translation

Drivingralle:
The "and" string (line 475) for the front end contains white space. This where missing in the German translation.


mjangda:
Thanks @Drivingralle!

PR #313, opened 2015-10-08
Updated delete_user_action function to fix issues #268 and #312

### LINKED ISSUE #268: Flaw in delete_user_action function for wp_delete_term
Hello,

This is a two part request/issue as I have put a fix for the Change Username plugin (https://github.com/Section214/Username-Changer) so upon username change the coauthorship is also transitioned. To do this I have to first remove the term then call add_coauthors function.
1. Would love a remove_coauthors function that allowed for the removal of coauthors dynamically through php.
2. Trying to use the wp_delete_term as seen on line 884 doesn't actually work as the user_login isn't a term_id. So I feel this is incorrect within the delete_user_action function.

What I ended up having to do was lookup the term_id and utilize it as follows;
`$current_term = get_term_by( 'name', $current_username, $coauthors_plus->coauthor_taxonomy );
wp_delete_term( $current_term->term_id, $coauthors_plus->coauthor_taxonomy );`
I believe this approach is necessary.

Thanks

### LINKED ISSUE #312: Delete User Action sql post lookup broken so reassignment not working
Hello,

I found if you reassign a user on delete that they aren't actually made into a coauthor if the original user was a coauthor on a post. Investigating into this turns out the sql call doesn't return any post ids;
```php
				$post_ids = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author = %d", $delete_id ) );
```

If I log the post_ids it returns empty, would only return true if the user is the original author any secondary authors won't be on the post as a post_author.

Cheers

garretthyder:
I've updated the delete_user_action to correct two issues (#268 and #312).

To fix #268 I've pulled the coauthor_term to utilize it's term_id in the wp_delete_term so as to properly delete the author term. Previously was using the user_login which does nothing. To reproduce the original issue you can delete a coauthor through the user_delete mechanism, then if you add a user with the exact same username they'll show up again on the post as a coauthor. This fix fully removes the relation to the post.

To fix #312 I've updated the posts lookup from the broken sql statement to a get_posts using the tax_query. Previously delete user would only reassign if the deleted user was the primary author as they'd be set as post_author, any secondary authors were overlooked. Now by using the get_posts and tax_query we cover all bases and properly reassign ALL posts/cpt/etc. to the reassign user.

Thoughts/Feedback? Happy to discuss.
Cheers

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -3,7 +3,7 @@
 Plugin Name: Co-Authors Plus
 Plugin URI: http://wordpress.org/extend/plugins/co-authors-plus/
 Description: Allows multiple authors to be assigned to a post. This plugin is an extended version of the Co-Authors plugin developed by Weston Ruter.
-Version: 3.2
+Version: 3.2.1
 Author: Mohammad Jangda, Daniel Bachhuber, Automattic
 Copyright: 2008-2015 Shared and distributed between Mohammad Jangda, Daniel Bachhuber, Weston Ruter
 
@@ -24,7 +24,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 */
 
-define( 'COAUTHORS_PLUS_VERSION', '3.1.2' );
+define( 'COAUTHORS_PLUS_VERSION', '3.2.1' );
 
 require_once( dirname( __FILE__ ) . '/template-tags.php' );
 require_once( dirname( __FILE__ ) . '/deprecated.php' );
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -857,17 +857,27 @@ class CoAuthors_Plus {
 	function delete_user_action( $delete_id ) {
 		global $wpdb;
 
+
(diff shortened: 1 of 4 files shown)


GaryJones:
Only the [last commit](https://github.com/Automattic/Co-Authors-Plus/pull/313/commits/4cb5cb499b86c9dbd7fd1e09da495be849bf2b12) is relevant, though #792 was merged which fixed one of the issues as well.

GaryJones:
Thanks for this, @garretthyder, and apologies for the very long delay.

Both issues this targeted, #268 and #312, were closed back in 2015, and the #268 fix — deleting the co-author term by its term ID in `delete_user_action` — has long since been part of the plugin. The branch itself has also drifted a long way from current `develop` and is no longer cleanly applicable, so we're going to close this one out.

For completeness: the broader concern behind #312 — reassigning a deleted user's *secondary* bylines, not only posts where they are the primary `post_author` — isn't fully handled today. If that's still useful to you, a fresh issue describing the scenario would be very welcome. Thank you again for the original contribution.

PR #315, opened 2015-10-14
improved search results for backend suggestion

diegocavalletti:
It looks like that backend suggestion searching is not fully accurate. For example we have multiple users that have "re" in their data but whenever we try to use that search string we just get a few results while as i said they should be many more.

In this commit i fetch the users you got to keep backwards compatibility and add them to the valid result set which will be filtered using the user capabilities.

Doing this in my environment (3.5k + users) provides a much more accurate results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,3 +7,4 @@ lint:
 
 phpunit:
 	phpunit
+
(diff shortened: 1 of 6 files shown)


GaryJones:
There are a lot of changes here , and it's not clear what it's meant to be doing. A change like this would also need tests.

As it's been 8 years without movement and the PR has conflicts, I'm going to close this one out.

Issue #206, opened 2014-06-24
Cannot edit guest authors when 38,000+ are present


Lexicality:
On a large WP install with 38,613 users, each of which have at least one guest author the "Edit Profile" link causes a page timeout. In addition various other parts of the WP-Admin stop responding during the timeout.
![2014-06-24_09-26-52](https://cloud.githubusercontent.com/assets/847422/3368779/79c9a1ca-fb79-11e3-88a6-875c7d51f475.png)

maxfenton:
+1

maxfenton:
@Lexicality Can you post a longer text snippet with exactly where to add that delay?

Lexicality:
@maxfenton voilà: #316

PR #316, opened 2015-10-25
Increase the typing detection timeout in the AJAX lookup to 1 second

Lexicality:
In response to https://github.com/Automattic/Co-Authors-Plus/issues/206#issuecomment-136903354, this boosts the typing detection threshold to 1000ms.
On the one hand this means that the suggestions won't appear magically as you type, but on the other hand it also means that poor server setups won't start to chug horribly if the user is a slow typist.

I'll admit that this could probably be fine-tuned; when I picked this value last year it was a high number that happened to solve the problem we were having at the time and there wasn't much scope for testing the limits of what would or wouldn't break the site.

### DIFF
--- a/js/co-authors-plus.js
+++ b/js/co-authors-plus.js
@@ -169,7 +169,8 @@ jQuery( document ).ready(function () {
 			})
 			.appendTo( $coauthors_div )
 			.suggest( coAuthorsPlus_ajax_suggest_link, {
-				onSelect: coauthors_autosuggest_select
+				onSelect: coauthors_autosuggest_select,
+				delay: 1000
 			})
 			.keydown( coauthors_autosuggest_keydown )
 			;


maxfenton:
Thank you!

joshbetz:
@Lexicality Do you happen to know what the default is? I think 500 or maybe even 250ms would be better, but I'm not sure how much of an increase that is over the default value.

Lexicality:
Sorry, I mentioned this in the commit message but forgot to add it to the
pull request. The default is 100ms.

bcampeau:
The query for looking up authors is quite expensive with a large set of users, I agree. I'd consider making this a filterable option implemented via `wp_localize_script` so different implementations could adjust the tolerance.

bcampeau:
I actually deployed this branch yesterday to a site that was having an issue with a very large user table and it is working quite well to improve performance thus far.

mjangda:
Seems like a sane change to me. I think 500ms is probably better as Josh says but we can run with this for now and see how it feels.

:ship:

PR #318, opened 2015-11-02
Have strings reflect filtered version of singular author label

danielbachhuber:


danielbachhuber:
@joshbetz #reviewmerge

PR #301, opened 2015-09-15
Revert "Revert "Introduce template tag tests""

mattoperry:
Reverts Automattic/Co-Authors-Plus#299  re-instating -- tests are working again

Issue #279, opened 2015-09-01
Incompatibility when `coauthors_auto_apply_template_tags` set to true


jjeaton:
When enabling this filter `coauthors_auto_apply_template_tags` in order to get RSS feeds correctly filtered, it also has the side effect of duplicating author's names if any `coauthors_*` functions are used elsewhere on the site (as the filters get applied there as well).

Is this filter only meant to be used when none of the `coauthors_*` functions are used?

If so, would it be possible/beneficial to separate the `the_author` and `the_author_posts_link` filters in `CoAuthors_Template_Filters` from the RSS filters in the class, or a way to selectively turn one group on or the other?

It seems the only approach to support RSS feeds now would be either:
1. Enable the filter, but manually remove the first two filters to avoid duplicating author's names.
2. Don't enable the filter, but duplicate the RSS filters elsewhere in your code to reimplement the functionality.

Maybe there is a better way? Happy to submit a PR.

natebot:
@jjeaton what's the simplest way for me to replicate your issue?

1. enable `coauthors_auto_apply_template_tags` filter.
2. use a `coauthors_*` function on any template.
3. see the authors' names repeated in each item of a feed.

Any additional steps?

jjeaton:
Yes, I'll get back to you soon with full steps to reproduce. Thanks!

jjeaton:
@natebot 

I was able to reproduce this on a stock 4.3 install running CAP v3.1.1 and Twenty Fifteen. The issue isn't that the author's names are repeated in a feed, it's that they are repeated when using `coauthors_posts_links()` or `the_author_posts_link()` in a template. This is because the coauthors filters are doubled up on these functions.

1. Enable `coauthors_auto_apply_template_tags` filter.
2. Add multiple authors to a post and publish.
2. Use `coauthors_posts_links()` or `the_author_posts_link()` in a template (
3. See each author's name duplicated for each link in the output (screenshot below)

![2015-09-08 at 4 03 pm](https://cloud.githubusercontent.com/assets/835509/9746002/200efdee-5643-11e5-9474-67ca9dabeb7a.png)

There is no issue when using `the_author()`, it seems to only be when using one of the functions that outputs the link to the author's posts, as both `the_author` and `the_author_posts_link` are filtered by CAP.

natebot:
I can confirm this bug.  It is still an issue in v3.2
[...]

PR #330, opened 2016-02-04
[issue-186] adding unit tests for template-tags

castlehouse:
First attempt at creating unit tests, trying to see if Travis builds the same as local

### DIFF
--- a/tests/test-template-tags.php
+++ b/tests/test-template-tags.php
@@ -41,8 +41,6 @@ class Test_Template_Tags_CoAuthors extends CoAuthorsPlus_TestCase {
 
 		wp_reset_postdata();
 
-		$this->assertWPError( get_coauthors() );
-
 	}
 
 	public function test_coauthors_post_links() {
--- a/tests/test-template-tags.php
+++ b/tests/test-template-tags.php
@@ -1,0 +1,162 @@
+<?php
+
+class Test_Template_Tags_CoAuthors extends CoAuthorsPlus_TestCase {
+
+	/**
+	 * Test get Co-Authors function
+	 */
+	public function test_is_coauthor_for_post() {
+
+		// Test only sending user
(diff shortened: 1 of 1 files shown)


tests/test-template-tags.php
```diff
@@ -0,0 +1,171 @@
...
+
+class Test_Template_Tags_CoAuthors extends CoAuthorsPlus_TestCase {
+
+	/**
+	 * Test get Co-Authors function
+	 */
+	public function test_is_coauthor_for_post() {
+
+		// Test only sending user
+		$this->assertEquals( false, is_coauthor_for_post( '' ) );
```
nb: Why did you decide to have many asserts in the same method, instead of separate methods for each check?

tests/test-template-tags.php
```diff
@@ -0,0 +1,171 @@
...
+
+		// Test sending user but not post_id
+		$this->assertEquals( false, is_coauthor_for_post( $this->author1 ) );
+
+		// Test sending valid post but no user
+		$this->assertEquals( false, is_coauthor_for_post( '', 1 ) );
+
+		// Check if author1_post1 belongs to author1 as set in coauthorsplus-testcase.php
```
nb: What will happen if we run only this file without running `coauthorsplus-testcase.php`?

tests/test-template-tags.php
castlehouse: I was following the style started by the other unit tests, and it meant being able to group them per function being tested

tests/test-template-tags.php
castlehouse: The test class extends from coauthorplus-test case so you would have to change this to be able to do so.

trepmal:
Related #301

philipjohn:
@trepmal Does this conflict with, or compliment, #301?

GaryJones:
Between #253, #279, and others, many tests for template tags now exist.

Issue #332, opened 2016-02-12
get_the_coauthor_meta does not always return successfully


castlehouse:
When you call the function the correct data is only returned when an author is assigned to a post who isn't linked to a guest author.

So when a guest author is assigned with a wp users associated or when a guest author with no wp user associated the data return is either empty or incorrect.

I have reproduced this on a VIP quickstart environment as well as a new multi site install.

trepmal:
For reference/reproduction:
I have a simple loop, with 3 posts, each with a different kind of user
```
while( $q->have_posts() ) {
	$q->the_post();
	echo '<ul>';
	echo '<li><strong>'. get_the_title() . '</strong></li>';
	echo '<li>Authors: '. implode( ', ', wp_list_pluck( get_coauthors(), 'display_name' ) ) . '</li>';
	echo '<li><pre>';
		var_dump( get_the_coauthor_meta( 'email' ) );
	echo '</pre></li>';
	echo '</ul><hr />';
}
```

Output:

* **Mapped Author**  
Authors: auntie  
```
array(1) {
  [965]=>
  string(0) ""
}
```
* **Just Guest Author (no mapping)**  
Authors: Tillamook  
```
array(1) {
  [885]=>
  string(0) ""
}
```

* **Just WP User**  
Authors: Ms Editor  
```
array(1) {
  [3]=>
  string(24) "editoremail@example.com"
}
```

Top 2 confirm that if the post author is either a Guest Author, or a Mapped co-author, no meta is displayed

PR #333, opened 2016-02-15
Fix for get_coauthor_meta function

castlehouse:
Allows function to work with wp user, guest author and wp user linked to guest author. See #332


rebeccahum:
This one doesn't quite work for guest authors linked to users.

rebeccahum:
This one doesn't quite work for guest authors linked to users.

GaryJones:
Looks like the underlying issue was fixed by #484.

PR #334, opened 2016-02-16
REST API Support

hbarroso:
This PR will allows the Co-Authors-Plus plugin to serve REST services through the new WordPress 4.4 API feature.

Please refer to `API.md` file for the full documentation.


mjangda:
> coauthors/v1/author/{author_id}/{post_id}

Why do we include the author_id in the route rather than as a separate param? Should we have a matching GET endpoint to get authors for a post?

hbarroso:
@mjangda That is a very rough route idea of what I wanted to do when I started. But I agree, it shouldn't need the `author_id`. Also I think it would be a good idea to also rename that route to `post/{post_id}` , in which we could add, get, remove multiple users from a post as parameter of array of elements.

co-authors-plus.php
```diff
@@ -126,6 +127,7 @@ function coauthors_plus() {
 	 * and the custom post type to store our author data
 	 */
 	public function action_init() {
+        global $wp_version;
```
mdawaffe: There's some whitespace issues here.  Also, this variable is not used.

mdawaffe:
Search Endpoint:

Why is this a POST request? A sort of similar WP-API request (listing posts by an author) is a GET:
`GET /wp-json/wp/v2/posts?author={author_id}`

What does `existing_authors` mean? Don't I want to search for existing authors?

mdawaffe:
POSTS - GET endpoint:

Why does this require authorization? The authorship of posts is public.

mdawaffe:
Post - POST endpoint:

The response for a POST endpoint should the the same as the response for the GET endpoint: a representation of the object being modified/retrieved.

Is the `append` option important?

mdawaffe:
Post - DELETE endpoint:

Can you look into how deletions are done in the rest of WP-API.  In general, we shouldn't be returning status messages, we should be returning representations of objects (or error codes/messages).

mdawaffe:
Guest endpoints:

What's the difference between POST and PUT?

mdawaffe:
I general, it seems like there are some inconsistencies in the endpoints: use of POST, PUT, append, and what is returned (messages, arrays, objects).

In REST, the resources are all objects, so URLs are typically constructed with nouns and look like: collection/id/collection/id/...:
/posts/
/posts/:post_id
/posts/:post_id/comments/
/users/
/users/:user_id/social-accounts/
...

Here we have:
/search ← a verb
/post ← a noun, but not a collection
/guest ← a noun, but not a collection

Do you have any thoughts on making these endpoints more consistent?

mdawaffe:
When I go to http://.../author/foobar on my site, I get a list of posts by that author (whether they are the "real" author or a co-author).  When I make a WP-API request like:
`GET /wp-json/wp/v2/posts?author={user_id_of_foobar}`
(The WP-API endpoint, which will be in core some day)
the posts for which foobar is a co-author do not show up.  Should they?

tests/test-api.php
```diff
@@ -0,0 +1,392 @@
+<?php
+
+/**
+ * Test Co-Authors Plus' REST API
+ */
+global $wp_version;
+if (version_compare($wp_version, '4.4', '>=')) {
```
mdawaffe: Doing a big conditional around a class definition is a bit ugly. Is there a better way to achieve the goal?

tests/test-api.php
```diff
@@ -0,0 +1,392 @@
...
+                $request->set_param($key, $value);
+            }
+            return $this->server->dispatch( $request );
+        }
+
+        /**
+         * Clears any persisted authentication
+         */
+        protected function logout() {
+            wp_set_current_user( -1 );
```
mdawaffe: How did you pick user_id -1?

co-authors-plus.php
```diff
@@ -947,24 +949,27 @@ function filter_count_user_posts( $count, $user_id ) {
 	/**
 	 * Checks to see if the current user can set authors or not
 	 */
-	function current_user_can_set_authors( $post = null ) {
+	function current_user_can_set_authors( $post = null, $is_api_request = false ) {
```
mdawaffe: In 271bce4720509aa8fe21abc37c438dcf813964cd, you mentioned that you added this `$is_api_request` parameter, but you didn't say why.  Why is it important/useful?

mdawaffe:
There are some code styling issues (mostly whitespace stuff - tabs v. spaces and the like). Is there a quick way to clean that all up?

mdawaffe:
There are some anonymous PHP functions (a.k.a. "closures"), which require PHP 5.3. The minimum requirement for WordPress is PHP 5.2.4, though I'm not sure if this plugin imposes stricter requirements.

mattoperry:
Just noting that dropping PHP 5.2 support is okay -- many of our maintained plugins don't support it and I think we'll officially stop supporting it with CAP in an upcoming release (maybe 3.2 or 3.3)

hbarroso:
Hey guys,
I've just updated the documentation.
I've addressed most issues in the comments. I believe the routes a lot cleaner and more usable. But feedback is welcome.

mdawaffe:
Documentation:
Can the documentation live in the plugin? Perhaps in the readme.txt file or some API.md file?

I believe the search endpoint is a GET endpoint, not POST as above.  If the documentation were with the plugin, it'd be easier to keep up to date :)

mdawaffe:
Sanitation: I have several questions about how data is sanitized.

* Why does `sanitize_array()` use `esc_attr()`?
* Why use `sanitize_key` for:
  * search `q` (which I believe searches email addresses, display names, etc.)
  * `linked_account` - is this always a numeric ID?
  * `aim`, `yahooim`, `jabber` - I'm not sure what characters are allowed in these fields (not that they are used much). We should probably be more flexible.
  * `id` - is this always a numeric ID?

Can you look over the rest of the sanitation callbacks and make sure they make sense for each field?

mdawaffe:
Guest Authors:

In a PUT request, what happens if a request to the guest author endpoint is made with some input that does not correspond to a `type=guest-author` post?

Can all the `get_args()` definitions be simplified with `get_guest_author_fields()`?

When reassigning with `reassign-another`, should `leave-assigned-to` be a nicename or an ID or something else?

In `prepare_params_for_database()`, is the `$ignore_user_login` parameter needed?

I think `search_guests()` needs a call to `$wpdb->esc_like()`.

mdawaffe:
I like the new routes :)

There's a few outstanding questions from earlier.  Could you check them out?

* https://github.com/Automattic/Co-Authors-Plus/pull/334#issuecomment-189037843: Why does [the post GET endpoint] require authorization? The authorship of posts is public.
* https://github.com/Automattic/Co-Authors-Plus/pull/334#issuecomment-189048694: "core" WP-API request question.
* https://github.com/Automattic/Co-Authors-Plus/pull/334#discussion_r54188556: How did you pick user_id -1?
* https://github.com/Automattic/Co-Authors-Plus/pull/334#discussion_r54188870: In 271bce4, you mentioned that you added this $is_api_request parameter, but you didn't say why. Why is it important/useful?

hbarroso:
> Documentation:
> Can the documentation live in the plugin? Perhaps in the readme.txt file or some API.md file?
> 
> I believe the search endpoint is a GET endpoint, not POST as above. If the documentation were with the plugin, it'd be easier to keep up to date :)

I've just made a commit sorting this.

> Sanitation: I have several questions about how data is sanitized.
> 
> Why does sanitize_array() use esc_attr()?
> Why use sanitize_key for:
> search q (which I believe searches email addresses, display names, etc.)

You're right, I'm just a little paranoid about input data. I will change this to a more flexible function `sanitize_text_field` seems like a good candidate. 
> linked_account - is this always a numeric ID?

No, this is actually a string (login). I tried to keep it compatible with the current code inside the `class-coauthors-guest-authors.php`. Why did you got the impression it was an numeric ID?

> aim, yahooim, jabber - I'm not sure what characters are allowed in these fields (not that they are used much). We should probably be more flexible.

I'll do the same thing here as I'll do with the `q` parameter on search.

> id - is this always a numeric ID?

Yes.

> Can you look over the rest of the sanitation callbacks and make sure they make sense for each field?

Absolutely.

hbarroso:
> Guest Authors:
> 
> In a PUT request, what happens if a request to the guest author endpoint is made with some input that does not correspond to a type=guest-author post?

Currently nothing, the code is checking if the post is a `guest-author`, but I think I should probably add an error message in case it isn't. Good catch.

> 
> Can all the get_args() definitions be simplified with get_guest_author_fields()?
>

Yes, now that I look a it, I think I can at least improve the `sanitize_callback` key, so it doesn't get always repeated between `post` and `put_item`.  

> When reassigning with reassign-another, should leave-assigned-to be a nicename or an ID or something else?

I opted for the nicename, since it's what the previous code was using, I tried to maintain the same input types as close as possible. The search for authors actually returns both. So if you think this would benefit from having both or just the ID I can easily change this.

> 
> In prepare_params_for_database(), is the $ignore_user_login parameter needed?

Nope, you're right. I actually added the logic to only include fields from a particular context. So this can get removed.

> 
> I think search_guests() needs a call to $wpdb->esc_like().

Yes, definitely, I didn't knew about this one.

hbarroso:
> I like the new routes :)
> 
> There's a few outstanding questions from earlier. Could you check them out?
> 
> #334 (comment): Why does [the post GET endpoint] require authorization? The authorship of posts is public.

Yes, I missed that at the time, but I've already changed in a previous commit. It's not public.

> #334 (comment): "core" WP-API request question.

I don't think they should, since you are calling the CoAuthors API it should only show data that concernes it. It shouldn't be the plugin responsibility to handle or treat other types of data not directly related to it.

> #334 (comment): How did you pick user_id -1?

I just decided to use an integer that would never exist in a PK table.

> #334 (comment): In 271bce4, you mentioned that you added this $is_api_request parameter, but you didn't say why. Why is it important/useful?

When developing the API I tried to reuse as much code from the core has possible. One of the methods I used was the `current_user_can_set_authors`, however,  that method allowed a post to be null, and would fetch the current get post instead. But from the API perspective it would be useful to use that just to check the current user capabilities without any post. So I added that $is_api_request so it would ignore the `get_post()` and just check the users permissions.

mdawaffe:
> No, this is actually a string (login). I tried to keep it compatible with the current code inside the class-coauthors-guest-authors.php. Why did you got the impression it was an numeric ID?

It looks like an ID from the dropdown in the form, but I see now that there's some user_login <-> ID translation in place. Thanks.

mdawaffe:
> Documentation

It looks like the /posts/ endpoint documentation is wrong.  I don't think it returns email addresses? Is there anything else that needs to be audited?

mdawaffe:
> I opted for the nicename, since it's what the previous code was using, I tried to maintain the same input types as close as possible. The search for authors actually returns both. So if you think this would benefit from having both or just the ID I can easily change this.

In some places, we use a nicename/user_login (reassigning). In some places (delete), an ID. It seems like we should be internally consistent (or accept both?).  What do you think?

mdawaffe:
> > Why does [the post GET endpoint] require authorization? The authorship of posts is public.
> Yes, I missed that at the time, but I've already changed in a previous commit. It's now public.

The documentation still says it requires auth.

mdawaffe:
> > "core" WP-API request question.
> I don't think they should, since you are calling the CoAuthors API it should only show data that concernes it. It shouldn't be the plugin responsibility to handle or treat other types of data not directly related to it.

OK - I can see both sides but don't have an opinion either way.  Let's work with this as is and see if there are any use cases in the future.

mdawaffe:
> > How did you pick user_id -1?
> I just decided to use an integer that would never exist in a PK table.

Using `0` is the normal WordPress way. If you follow the code, you'll find it doesn't matter (-1 and 0 both work fine), but I think `0` is a bit more robust, just because that's what people normally do.  It's probably fine as is.

mdawaffe:
> `$is_api_request`

Two things:
1. Let's try to avoid the [boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html).  Perhaps `->current_user_can_set_authors( $post_id, array( 'api' => true )` would be better?
2. "api" describes where the parameter is used, but it doesn't describe what it does.  What if some future feature needed the same behavior but had nothing to do with the API? Can we come up with a better name?

mdawaffe:
Sanitation:

Thanks for cleaning that all up in ab0c76bfc84.  The `'q'` parameter in the guests API, though, still uses `sanitize_key()`, and then "manually" calls `sanitize_text_field()` later in '->get()`. Which one is "right"? Or are both needed?

mdawaffe:
`does_coauthor_exists()` and `does_coauthor_email_exists()` would read better as `does_coauthor_exist()` and `does_coauthor_email_exist()`.

mdawaffe:
@mattoperry Above are my last thoughts (I think :)). Do you have any feedback?

hbarroso:
> It looks like the /posts/ endpoint documentation is wrong. I don't think it returns email addresses? Is there anything else that needs to be audited?

You're right, since this route is now public I decided to remove the emails to avoid spammers from stealing contacts, but forgot to update the documentation. I've also updated other parts of the API documentation  where email is not used . Commit: https://github.com/Automattic/Co-Authors-Plus/pull/334/commits/8d2dd44a52b641dfbf8eed801611909746706263

hbarroso:
> The documentation still says it requires auth.

It's now in the commit in https://github.com/Automattic/Co-Authors-Plus/pull/334/commits/8d2dd44a52b641dfbf8eed801611909746706263 commit I mistaken by another route.

hbarroso:
> In some places, we use a nicename/user_login (reassigning). In some places (delete), an ID. It seems like we should be internally consistent (or accept both?). What do you think?

Yeah, it can get a little confusing. I'd say we go with nicename/user_login. Having both I don't see any benefit, and it would require changing some core functions like `get_coauthor_by` to fetch the DB for an id and username.

hbarroso:
> Using 0 is the normal WordPress way. If you follow the code, you'll find it doesn't matter (-1 and 0 both work fine), but I think 0 is a bit more robust, just because that's what people normally do. It's probably fine as is.

Updated in https://github.com/Automattic/Co-Authors-Plus/pull/334/commits/65ae179a94a09d3b9b35dab80695f2813d21488d

hbarroso:
> Two things:
> 1. Let's try to avoid the boolean trap. Perhaps ->current_user_can_set_authors( $post_id, array( 'api' => true ) would be better?
> 2. "api" describes where the parameter is used, but it doesn't describe what it does. What if some future feature needed the same behavior but had nothing to do with the API? Can we come up with a better name?

I suggest we take a different approach. If we look carefully, this method is actually breaking the Single Responsibility Principle. It states that it's for checking the user permission, but it's also checking for post_type. Worse is, it's even declaring variables that are not even being used like `$post_type_object` and inserting a `global $typenow` which isn't also being used anywhere in the scope.

I suggest we refactor this method into two different method. One for the checking the post type and another for the actual permissions. This way the API can safely use the correct one without having to use extra parameters.

Let me know what you think.

hbarroso:
> Sanitation:
> 
> Thanks for cleaning that all up in ab0c76b. The 'q' parameter in the guests API, though, still uses sanitize_key(), and then "manually" calls sanitize_text_field() later in '->get()`. Which one is "right"? Or are both needed?

Yup, this was redundant. I replaced it with the correct and removed the second one in commit https://github.com/Automattic/Co-Authors-Plus/pull/334/commits/311e3445f4b88d2358eeda50e66040992c0a3da0

hbarroso:
> does_coauthor_exists() and does_coauthor_email_exists() would read better as does_coauthor_exist() and does_coauthor_email_exist().
> 
> EDIT: actually, the name of does_coauthor_email_exists() is confusing. If it's passed a coauthor object, it can return false even when the email exists :) Is there a better name that describes what the function does?
> 
> Since it's only ever called like that once, should we just not bother with the second parameter? It could return false if the email doesn't exist, or the coauthor ID of the coauthor with that email if it does exist. Then we could do that secondary check in ->put_item() itself (or some other function).
> 
> I'm not sure if that's the best solution. Do you have any another suggestions?

I like your suggestion. I think, if we used that logic in more than one place it would be useful to put it inside a new method. But using the logic inside the->put_item() is suffice IMO. I've also updated the methods names to *_exist() in https://github.com/Automattic/Co-Authors-Plus/pull/334/commits/0c4599ae17f16e05479db43b66fc0bf4cea23680

mdawaffe:
>  I've also updated the methods names to *_exist() in 0c4599a

Thanks. Is `does_coauthor_email_exists()` still a good name, then?

hbarroso:
> Thanks. Is does_coauthor_email_exists() still a good name, then?

Yes, I think changing to something like `is_email_unique()` would be more meaningful now.

mdawaffe:
> I suggest we refactor this method into two different methods. One for checking the post type and another for the actual permissions. This way the API can safely use the correct one without having to use extra parameters.

I agree, though it seems like one function to check if a user can set authors for a specific post, and a different one for if a user can set authors for posts of a specific type:
`current_user_can_set_post_authors( $post_id )`, which could be used almost everywhere
`current_user_can_set_post_type_authors( $post_type )`, which could be used in a few places.

Would that cover everything, do you think?

hbarroso:
> current_user_can_set_post_authors( $post_id ), which could be used almost everywhere
> current_user_can_set_post_type_authors( $post_type ), which could be used in a few places.
> 
> Would that cover everything, do you think?

I like that. But what about the part that checks for the user level? 
Do mean, leave it inside the current `current_user_can_set_authors` as it is and use those two new methods as helpers inside as well? Or should we also refactor that code into a new method?

I'm talking about this code
```php
		$current_user = wp_get_current_user();
		if ( ! $current_user ) {
			return false;
		}
		// Super admins can do anything
		if ( function_exists( 'is_super_admin' ) && is_super_admin() ) {
			return true;
		}

		$can_set_authors = isset( $current_user->allcaps['edit_others_posts'] ) ? $current_user->allcaps['edit_others_posts'] : false;

		return apply_filters( 'coauthors_plus_edit_authors', $can_set_authors );
```

This code doesn't depend on post or post types.

mdawaffe:
> I'm talking about this code

That's the code I'm talking about too. It looks as though it's supposed to take into account post-specific capabilities but doesn't.

It seems like it should be a wrapper for either `current_user_can( 'edit_post', $post_id )` and/or checking for the `$post_type->caps->edit_others_posts` capability, but it doesn't do either :) I'm not sure what the "right" thing to do here is - thoughts? We can leave the implementation to a different PR, though.

hbarroso:
@mdawaffe Yeah, I'm not sure either. For the sake of keeping the PR more clean (regarding the API implementation) I'd suggest we refactor just that code into a different method so we don't need the `$is_api_request` and just use that new method. And then on a future PR we deal with this. What do you think?

mdawaffe:
That makes sense to me.

php/api/class-coauthors-api-guests.php
```diff
@@ -0,0 +1,447 @@
...
+		$data  = $this->prepare_data( array( $guest ) );
+
+		return $this->send_response( $data );
+
+	}
+
+	public function delete_item( WP_REST_Request $request ) {
+		global $coauthors_plus;
+
+		$coauthor_id = (int) sanitize_text_field( $request['id'] );
```
mattoperry: Probably just need either (int) or sanitize_text_field here (not both)

API.md
```diff
@@ -0,0 +1,289 @@
+This is the documentation support for the REST API feature.
+
+Route | Method | Description
+-------- | --------------- | ----------
+/**authors** | GET | Search for available authors.
+/**posts**/`:post_id`/**authors** | GET |  Get all authors from a post
+/**posts**/`:post_id`/**authors** | PUT |  Appends a new author to a post
+/**posts**/`:post_id`/**authors**/`:author_id` | DELETE |  Removes an author from a post
+/**guests** | GET |  Lists of filters all the currently available guests
```
mattoperry: can you clarify this line?

API.md
```diff
@@ -0,0 +1,289 @@
...
+
+Route | Method | Description
+-------- | --------------- | ----------
+/**authors** | GET | Search for available authors.
+/**posts**/`:post_id`/**authors** | GET |  Get all authors from a post
+/**posts**/`:post_id`/**authors** | PUT |  Appends a new author to a post
+/**posts**/`:post_id`/**authors**/`:author_id` | DELETE |  Removes an author from a post
+/**guests** | GET |  Lists of filters all the currently available guests 
+/**guests** | POST |  Creates a new guest
+/**guests**/`:guest_id` | PUT |  Updates a new guest
```
mattoperry: Probably updates an existing guest?

mattoperry:
Hey guys -- I've had a look over this and added a couple very minor notes, but it looks generally good and I'm not sure I have a ton of additional notes beyond what @mdawaffe has added above.  I think what I'd like to do here is to add this to the 3.3 milestone and test this in a couple of real-world situations ... there are also a couple of people on the VIP team I'd like to show this to.

php/api/class-coauthors-api-guests.php
hbarroso: Yes, it's redundant . I'll use (int) instead.

API.md
hbarroso: This should be "Lists OR filters current available guests". It's a typo. Going to update it the docs.

API.md
hbarroso: Yes. I'm going to correct this too.

djouonang:
Hello I have been trying to use the wp rest api to pull data from local site to live wordpress site with use of shortcode. When i try to do the reverse that is to pull posts from live to local it works but when i change links to retrieve from local to live it does not display anything on the page. I already have the wp rest api installed on both sites. Below is my code:

function my_recent_posts_shortcode($atts){

$response = wp_remote_get( 'http://localhost/wordpress/wp-json/wp/v2/posts' );

if( is_wp_error( $response ) ) {

return;

}

$posts = json_decode( wp_remote_retrieve_body( $response ) );

if( empty( $posts ) ) {

return;

}

if( !empty( $posts ) ) {

$list = '';

foreach( $posts as $post ) {

$list .='

link. '">' . $post->title->rendered . '
';
}

return $list . '';

}

}

add_shortcode('recent-posts', 'my_recent_posts_shortcode');

misfist:
I'm really confused about what the status of this is. The REST API code isn't in the being discussed here doesn't seem to be in any existing branch. Could someone shed some light?

misfist:
Now that the REST API is in core, this has become even more important. Is there any update on this?

emhoracek:
@misfist the branch seems to be here: https://github.com/hbarroso/Co-Authors-Plus/tree/add/rest-api-endpoints

mikeselander:
@philipjohn @mjangda what's the plan for this PR and functionality? We have a client need for REST endpoints for coauthors plus and browsing through this PR it looks like most work has already been completed on this, just not merged?

philipjohn:
This isn't a priority right now, but we will have a bunch of work starting on the plugin soon so might be able to get this looked at as part of that.

misfist:
@philipjohn - How/why is making Co-authors Plus work with the REST API, which is now in core not a priority?

philipjohn:
@misfist Having API endpoints would definitely be nice, but it's not crucial to the functioning of this plugin which would work just as well without them. There are benefits to having these API endpoints, but the priority is fixing existing bugs and other more frequently requested enhancements.

PR #335, opened 2016-02-19
Add filter for meta box title

romancandlethoughts:


mjangda:
The apply_filters should go outside the translation function.

PR #337, opened 2016-02-27
Guest authors with special characters

### LINKED ISSUE #198: Guest authors with non-ASCII characters can't be used as co-authors
[Reported in the forums](http://wordpress.org/support/topic/non-ascii-guest-author-names-cannot-be-added-to-post)

Needs a fix with unit test coverage both ways so we don't accidentally break something.

benhutchins:
Fix for Automattic/Co-Authors-Plus#198 to support guest authors with special characters in display names. On the client side we receive URL encoded nicename values, so need to decode them before passing them to jQuery which was double encoding them. Additionally there was an inconsistent use of sanitize_title versus sanitize_text_field used when saving guest authors and then when looking them up when saving a regular post, resolving that allowed you to actually save the authors array when an author had a special character.

Fixes #198

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -772,7 +772,7 @@ class coauthors_plus {
 				check_admin_referer( 'coauthors-edit', 'coauthors-nonce' );
 
 				$coauthors = (array) $_POST['coauthors'];
-				$coauthors = array_map( 'sanitize_text_field', $coauthors );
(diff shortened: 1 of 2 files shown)


GaryJones:
As #198 was fixed by #473, and this PR now only contains whitespace changes, I'm closing it.

PR #340, opened 2016-03-26
PHPCS indentation fixes

paulschreiber:


co-authors-plus.php
```diff
@@ -1249,8 +1249,8 @@ public function js_vars() {
 				// AJAX link used for the autosuggest
 				var coAuthorsPlus_ajax_suggest_link = <?php
 				echo json_encode( add_query_arg( array(
```
mjangda: Probably better just to drop the `array(` to a newline instead?

PR #345, opened 2016-03-26
use underscores in class names for consistency

paulschreiber:
Most classes had underscores already; now they all do.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -6,9 +6,9 @@
  * @since 3.0
  * @see https://github.com/wp-cli/wp-cli
  */
-WP_CLI::add_command( 'co-authors-plus', 'CoAuthorsPlus_Command' );
+WP_CLI::add_command( 'co-authors-plus', 'CoAuthors_Plus_Command' );
 
-class CoAuthorsPlus_Command extends WP_CLI_Command {
+class CoAuthors_Plus_Command extends WP_CLI_Command {
 
 	/**
 	 * Subcommand to create guest authors based on users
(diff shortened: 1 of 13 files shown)


template-tags.php
```diff
@@ -77,7 +77,7 @@ function is_coauthor_for_post( $user, $post_id = 0 ) {
 	return false;
 }
 
-class CoAuthorsIterator {
+class CoAuthors_Iterator {
```
mjangda: We can't change this one because of back-compat (if someone is using the class directly, for example).

php/class-wp-cli.php
```diff
@@ -6,9 +6,9 @@
  * @since 3.0
  * @see https://github.com/wp-cli/wp-cli
  */
-WP_CLI::add_command( 'co-authors-plus', 'CoAuthorsPlus_Command' );
+WP_CLI::add_command( 'co-authors-plus', 'CoAuthors_Plus_Command' );
 
-class CoAuthorsPlus_Command extends WP_CLI_Command {
+class CoAuthors_Plus_Command extends WP_CLI_Command {
```
mjangda: Also a back-compat issue.

paulschreiber: How is this on a back-compat issue? People don't invoke the CLI class directly, do they?

joshbetz: @paulschreiber You could extend `CoAuthorsPlus_Command` to write your own commands.

paulschreiber: That is technically true. It doesn't see likely. (A google search didn't turn up any examples.)

PR #346, opened 2016-03-30
don't silence errors

paulschreiber:
Remove @ from in front of two functions.


co-authors-plus.php
```diff
@@ -1501,7 +1501,7 @@ function wp_notify_postauthor( $comment_id, $comment_type = '' ) {
 				return false;
 			}
 
-			$comment_author_domain = @gethostbyaddr( $comment->comment_author_IP );
+			$comment_author_domain = gethostbyaddr( $comment->comment_author_IP );
```
mjangda: Core does the same thing. I believe it's because this function can get pretty noisy with warnings when you don't pass a valid IP.

PR #347, opened 2016-03-30
change json_encode() to wp_json_encode()

paulschreiber:


mjangda:
We'll need to bump the `Requires at least:` version to 4.1.

PR #353, opened 2016-05-04
Correctly pass in author name to coauthors_create_autosuggest

joshlevinson:
Currently, the entire author object is passed into `coauthors_create_autosuggest`. However, `coauthors_create_autosuggest` expects just the author's display name. This commit correctly passes in the `author.name` property.


mjangda:
Looks good! Thanks!

PR #356, opened 2016-05-10
Syncing changes from wpcom

philipjohn:
Making sure that changes made on wpcom make their way here before we commit any new versions from GH to wpcom.


mjangda:
Hm, the commit message doesn't match the change (adding new avatar size).

mjangda:
Okay, just confirmed that it is indeed part of the wpcom commit. Good to merge.

PR #361, opened 2016-06-03
prevent | in author display name from breaking author assigning

lipemat:
When an author's display name contains a | character the assigning breaks due to the current js use of this.value.split( '|' ); within coauthors_autosuggest_select().

I was able to resolve this with a simple convert to &#108; which display almost identical to | but does not break the js logic.

### DIFF
--- a/co-authors-plus-lipemat.php
+++ b/co-authors-plus-lipemat.php
@@ -1,6 +1,6 @@
 <?php
 /*
-Plugin Name: Co-Authors Plus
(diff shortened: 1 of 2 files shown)


philipjohn:
Thanks for trying to tackle this! It's a known issue that we need to sort out. Rather than this approach, though, we're going to focus on upgrading our use of Select2 (#389) to deal with it. Contributions welcome on that!

PR #363, opened 2016-06-24
Send moderation email even if guest author has no email address

### LINKED ISSUE #364: Moderation notification emails not generated if guest author has no email address set
If a guest author does not have any email address configured, then when someone comments on posts by that guest author, no moderation notification emails are sent to any users (even admins who have emails set).

I've created a pull request to fix this: https://github.com/Automattic/Co-Authors-Plus/pull/363

RobjS:
Currently, if a guest author has no email address set, then no moderation emails are sent to any users if someone comments on a post by that guest author.

This fixes the issue, so that other users still receive emails even if the guest author has no email address set.

fixes #364


mjangda:
Nice, thanks for the fix!

PR #366, opened 2016-07-13
Add filter to get_coauthors

jasonbahl:
Adds a filter to get_coauthors to allow hooking in and providing coauthor objects that might not exist in the db. Use case: we have several places in our site where we output authors based on get_coauthors (templates, some custom POST requests we send out to some API's, etc). We have the need to "fake" some coauthors in some instances, and want to be able to dynamically insert them in when a post meets certain criteria. Being able to filter the data that gets returned when calling get_coauthors will allow us to take advantage of all existing code where we use get_coauthors, but will allow us to dynamically inject "fake" authors when needed.

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -36,7 +36,7 @@ function get_coauthors( $post_id = 0 ) {
 			}
 		} // the empty else case is because if we force guest authors, we don't ever care what value wp_posts.post_author has.
 	}
-	return $coauthors;
+	return apply_filters( 'get_coauthors', $coauthors, $post_id );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mslinnea:
I also would have found this useful recently.  In the use case I encountered, we had hidden authors that should not be displayed on template pages.  If this filter existed, the solution would have been super easy.

rebeccahum:
Do you mind resolving the conflicts?

rebeccahum:
Thanks for the contribution!  Do you mind resolving the conflicts before we can proceed?

jasonbahl:
for whatever reason I can't get .git to fix the merge conflict on this PR, so I opened a new PR here https://github.com/Automattic/Co-Authors-Plus/pull/646

PR #367, opened 2016-07-19
Update CLI to use cached functions

jasonbahl:


mjangda:
Thanks for the PR, jasonbahl. Can you please provide more context around the change?

jasonbahl:
As part of a VIP GO performance review, VIP GO pointed out several places in our codebase that weren't using cached functions, and this was part of it. I don't think these really affect our performance, but we agreed that we would adjust our controlled codebase and do our best to contribute back to plugins we don't control.

bcampeau:
Yeah @mjangda, @emrikol requested this.

mjangda:
👍

mjangda:
@philipjohn I think we may need to revert this (because of #390) and explicitly pass `orderby => termorder` to the `get` calls.

PR #368, opened 2016-07-19
Update to use cached functions

jasonbahl:
As part of a VIP GO code review, we were asked to update our codebase to use cached functions. We updated our controlled code, and are working to submit PRs to codebases we don't directly control

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -523,7 +523,11 @@ class CoAuthors_Plus {
 		$term_ids = $wpdb->get_results( "SELECT term_id FROM $wpdb->term_taxonomy WHERE term_taxonomy_id IN ($tt_ids)" );
 
 		foreach ( (array) $term_ids as $term_id_result ) {
-			$term = get_term_by( 'id', $term_id_result->term_id, $this->coauthor_taxonomy );
+			if ( function_exists( 'wpcom_vip_get_term_by' ) ) {
+				$term = wpcom_vip_get_term_by( 'id', $term_id_result->term_id, $this->coauthor_taxonomy );
+			} else {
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -784,8 +792,9 @@ function coauthors_update_post( $post_id, $post ) {
 	}
 
 	function has_author_terms( $post_id ) {
-		$terms = wp_get_object_terms( $post_id, $this->coauthor_taxonomy, array( 'fields' => 'ids' ) );
-		return ! empty( $terms ) && ! is_wp_error( $terms );
+		$term_objects = get_the_terms( $post_id, $this->coauthor_taxonomy );
+		$terms = ( ! is_wp_error( $term_objects ) ) ? wp_list_pluck( $term_objects, 'term_id' ) : array();
```
mjangda: The list pluck is unnecessary here. We can just do `return ! is_wp_error( $term_objects ) && ! empty( $term_objects )`

mjangda:
I think we should drop the uses of `wpcom_vip_get_term_by` (i.e. VIP-specific functions) here (the `id` lookups use `get_term` anyway, which should be cached already?)

bcampeau:
For the instances where we're using `id` as the first param, we don't even really need `get_term_by` at all. `get_term` would be sufficient there and that's definitely a cached function.

For the other usage by slug, I'm fine with the check for the existence of `wpcom_vip_get_term_by` since there really isn't another way to make it a cached function.

jasonbahl:
Going through some old open PRs of mine. . .is this worth merging with some changes or should I close this?

rebeccahum:
Closing as `wpcom_vip_get_term_by()` is deprecated now, see: https://vip.wordpress.com/functions/wpcom_vip_get_term_by/

PR #378, opened 2016-09-19
Add offset and number to create-guest-authors

aaronjorbin:
Currently, create-guest-authors is an all or nothing proposition. What happens though when you are 50% complete with a 30,000 user import and then accidently close your terminal window?  You have to start over.  By allowing chunking of create-guest-authors, this can be continued from the number that have been completed rather than starting from scratch.  

Anecdotally, running it in chunks seems to be quicker.

### DIFF
--- a/features/create-guest-author.feature
+++ b/features/create-guest-author.feature
@@ -23,3 +23,29 @@ Feature: Guest authors can be created
       - 0 guest author profiles were created
       - 1 users already had guest author profiles
       """
+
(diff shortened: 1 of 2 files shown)


GaryJones:
Only the [last commit](https://github.com/Automattic/Co-Authors-Plus/pull/378/commits/f6bb3e2b062735ebef746d32be10ae12788e4393) is relevant here. Simple but useful addition.

Issue #374, opened 2016-08-17
CoAuthors Plus potential incompatibility with WordPress 4.6


kevinlisota:
Upgraded to WordPress 4.6 last night, and now our install is receiving the following error when viewing our author archive pages.

We use a custom query on our author archives like this:

```
$curauth = $coauthors_plus->get_coauthor_by( 'user_nicename', $author_name );

$featured_posts_author = new WP_Query(
array(
    'post_type' => 'post',
    'post_status' => 'publish',
    'posts_per_page' => 4,
    'author_name' => $curauth->user_nicename,
    'tax_query' => [
        [
            'taxonomy' => 'display_options',
            'field' => 'slug',
            'terms' => 'featured'
        ]
    ],
    'orderby' => 'date',
    'order' => 'DESC'
));
```

This throws the following database error. Did not happen prior to WP 4.6. Any thoughts on why this is happening?

`WordPress database error Not unique table/alias: 'wp_term_relationships' for query SELECT SQL_CALC_FOUND_ROWS  wp_posts.ID FROM wp_posts  LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id) LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id) LEFT JOIN wp_term_taxonomy ON ( wp_term_relationships.term_taxonomy_id = wp_term_taxonomy.term_taxonomy_id ) WHERE 1=1  AND ( 
wp_term_relationships.term_taxonomy_id IN (2569)\n) AND ((wp_posts.post_author = 9962 OR (wp_term_taxonomy.taxonomy = 'author' AND wp_term_taxonomy.term_id = '8109'))) AND wp_posts.post_type IN ('post') AND ((wp_posts.post_status = 'publish')) GROUP BY wp_posts.ID HAVING MAX( IF ( wp_term_taxonomy.taxonomy = 'author', IF (  wp_term_taxonomy.term_id = '8109',2,1 ),0 ) ) <> 1  ORDER BY wp_posts.post_date DESC LIMIT 0, 4 /*`

kevinlisota:
Specifically, I think the plugin is adding a duplicate INNER JOIN. There is code in the `posts_join_filter` function that checks for this, but appears to no longer work properly on WP 4.6

mjangda:
Looks like this happens when you combine a taxonomy query (tag, category, etc.) and an author query.

pstonier:
[...]

PR #381, opened 2016-11-10
Join term_relationships with alias

trepmal:
When joining term_taxonomy, also join term_relationships with alias to prevent unintended matches in case of another taxonomy query.
Addresses https://github.com/Automattic/Co-Authors-Plus/issues/374

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -614,7 +614,8 @@ class CoAuthors_Plus {
 			$term_relationship_inner_join = " INNER JOIN {$wpdb->term_relationships} ON ({$wpdb->posts}.ID = {$wpdb->term_relationships}.object_id)";
 			$term_relationship_left_join = " LEFT JOIN {$wpdb->term_relationships} ON ({$wpdb->posts}.ID = {$wpdb->term_relationships}.object_id)";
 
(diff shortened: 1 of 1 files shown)


mattoperry:
Hmmm ... maybe something wrong with the build tests here, or is there really a problem?

sboisvert:
@mattoperry The builds are failing on 5.2 IIRC we agreed to stop supporting 5.2 for any shared plugin and make the minimum PHP version supported 5.4. I think we should change the travis tests to reflect this.

mjangda:
> we agreed to stop supporting 5.2 for any shared plugin and make the minimum PHP version supported 5.4

This is tangential to this PR, but if we are doing this, we need to make sure, we need to make sure to add the necessary checks in place to avoid fatal-ing when the plugin is installed on < PHP5.4.

sboisvert:
Agreed.
@philipjohn having a standard way of doing this might be nice, is this something that would fit in your rotation? Not necessarily doing it as much as thinking about it and overseeing it.

mjangda:
Let's move discussion on the PHP5.4 thing to a new issue to avoid derailing the conversation here :)

https://github.com/Automattic/Co-Authors-Plus/issues/384

Issue #354, opened 2016-05-04
Patch for coauthors_plus->fix_author_page() _doing_it_wrong() error


onpubcom:
Co-Authors-Plus currently generates the following PHP Notice in the case where the $wp_query global is not yet defined:

`
PHP Notice:  is_author was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run. Before then, they always return false. Please see <a href="https://codex.wordpress.org/Debugging_in_WordPress">Debugging in WordPress</a> for more information. (This message was added in version 3.1.) in wp-includes/functions.php on line 3897,
PHP Stack trace:,
PHP   1. {main}() index.php:0,
PHP   2. require() index.php:17,
PHP   3. require_once() wp-blog-header.php:13,
PHP   4. require_once() wp-load.php:42,
PHP   5. require_once() wp-config.php:108,
PHP   6. do_action($tag = *uninitialized*, $arg = *uninitialized*) wp-settings.php:277,
PHP   7. call_user_func_array:{wp-includes/plugin.php:525}(*uninitialized*, *uninitialized*) wp-includes/plugin.php:525,
PHP   8. cycloneslider_init(*uninitialized*) wp-includes/plugin.php:525,
PHP   9. CycloneSlider_Plugin->run() plugins/cyclone-slider-2/cyclone-slider.php:160,
PHP  10. CycloneSlider_Plugin->offsetGet($offset = *uninitialized*) plugins/cyclone-slider-2/cyclone-slider.php:35,
PHP  11. cycloneslider_service_export_page($plugin = *uninitialized*) plugins/cyclone-slider-2/src/CycloneSlider/Plugin.php:24,
PHP  12. CycloneSlider_Data->get_sliders($args = *uninitialized*) plugins/cyclone-slider-2/cyclone-slider.php:369,
PHP  13. get_posts($args = *uninitialized*) plugins/cyclone-slider-2/src/CycloneSlider/Data.php:263,
PHP  14. WP_Query->query($query = *uninitialized*) wp-includes/post.php:1763,
PHP  15. WP_Query->get_posts() wp-includes/query.php:4006,
PHP  16. do_action($tag = *uninitialized*, $arg = *uninitialized*) wp-includes/query.php:3409,
PHP  17. call_user_func_array:{wp-includes/plugin.php:525}(*uninitialized*, *uninitialized*) wp-includes/plugin.php:525,
[...]

PR #383, opened 2016-11-15
Ensure $wp_query is as expected

trepmal:
Some plugins may run a query earlier than expected causing CAP's is_author() check to send up a 'called incorrectly' notice.
Patch really just does a sanity check to make sure `$wp_query` is as expected before relying on it.
Addresses #354

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .DS_Store
 .svn
-wpcom-helper.php
 .idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,34 @@
 language: php
 
 php:
-    - 5.3
-    - 5.5
+    - "nightly"
 
 env:
     - WP_VERSION=latest
-    - WP_VERSION=4.1
-    - WP_VERSION=4.0
+
+matrix:
+  include:
+# nightly+latest already included above as first build.
+#  - php: "nightly"
+#    env: WP_VERSION=latest
+   - php: "nightly"
+     env: WP_VERSION=4.6
+   - php: "nightly"
+     env: WP_VERSION=4.5
+   - php: "5.2"
+     env: WP_VERSION=latest
+   - php: "5.2"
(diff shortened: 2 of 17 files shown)


mjangda:
This does fix the initially reported issue (i.e. the notice), but I wonder if we should actually fix the root of the problem, which feels like the plugin may not be hooking in at the ideal spot with the `fix_author_page` callback.

trepmal:
Do you have an inkling of where that ideal spot might be? :)

It doesn't seem like there's a later hook ready and waiting that the callback can be switched to with no effort. Of course, there may be a better hook that _does_ require a little more effort...

I'll take a deeper look at what `fix_author_page` does to see where else it might work

mjangda:
> Do you have an inkling of where that ideal spot might be? :)

I'm not totally sure :) It's a bit of a bummer that `posts_selection` doesn't pass through the query object. One of the filters like `posts_request` or `posts_results` might work, although, it would break this logic for any queries with `suppress_filters == true` (which may be okay if we document it clearly).

sboisvert:
>although, it would break this logic for any queries with suppress_filters == true (which may be okay if we document it clearly).

I'm not sure I'm following this clearly. Would this mean that any query with suppress_filters == true would fail for is_author()?
I'm trying to think of a context where this would be used. Since get_posts() suppresses filters that's what I'd think would be the most used way that suppress_filters == true, but then would someone use is_author()?

Am I following correctly?

mjangda:
Actually, CAP query mods don't fire when `supress_filters === true` anyway since filters like `posts_where` don't trigger.

Plus, for the `get_posts()` use case in this context, we don't really care that much anyway, since the queried object isn't accessible.

lkraav:
Hmm, this might be superceded by https://github.com/Automattic/Co-Authors-Plus/pull/801 by now.

nielslange:
> Hmm, this might be superceded by #801 by now.

Thanks for pointing this out, @lkraav.

Issue #375, opened 2016-08-23
preg_replace all in posts_where_filter


douglas-johnson:
I am experiencing an issue with #240 

It seems that changing the limit on the preg_replace from 1 to -1 can cause unintended consequences when a logged in user views an author archive. Namely, it shows private posts to any logged in user, regardless of capabilities. 

The pattern which looks for `.post_author` in the where clause is matched twice because [WP_Query->get_posts](https://core.trac.wordpress.org/browser/tags/4.6/src/wp-includes/query.php#L3178) adds an OR for the current user's private posts.

Adding `OR (wp_2_term_taxonomy.taxonomy = 'author' AND wp_2_term_taxonomy.term_id = '799')` to the query for the current user's id means that the author archive will display private posts from the author taxonomy. 
## Example

Pattern:
`/(\b(?:wp_2_posts\.)?post_author\s*=\s*(\d+))/`

Replacement:
`($1 OR (wp_2_term_taxonomy.taxonomy = 'author' AND wp_2_term_taxonomy.term_id = '799'))`
### Non Logged in User

Subject: 
`AND (wp_2_posts.post_author = 1) AND wp_2_posts.post_type = 'post' AND (wp_2_posts.post_status = 'publish')`

Outcome:
`AND ((wp_2_posts.post_author = 1 OR (wp_2_term_taxonomy.taxonomy = 'author' AND wp_2_term_taxonomy.term_id = '799'))) AND wp_2_posts.post_type = 'post' AND (wp_2_posts.post_status = 'publish')`
### Logged in User ( Contributor )

Subject: 
`AND (wp_2_posts.post_author = 1) AND wp_2_posts.post_type = 'post' AND (wp_2_posts.post_status = 'publish' OR wp_2_posts.post_author = 180 AND wp_2_posts.post_status = 'private')`

Outcome:
`AND ((wp_2_posts.post_author = 1 OR (wp_2_term_taxonomy.taxonomy = 'author' AND wp_2_term_taxonomy.term_id = '799'))) AND wp_2_posts.post_type = 'post' AND (wp_2_posts.post_status = 'publish' OR (wp_2_posts.post_author = 180 OR (wp_2_term_taxonomy.taxonomy = 'author' AND wp_2_term_taxonomy.term_id = '799')) AND wp_2_posts.post_status = 'private')`

trepmal:
Okay, I've been discecting this query...

[...]

PR #386, opened 2016-11-21
Fix private post viewing on front-end

trepmal:
Lets coauthors see private posts in author feeds, and prevents users from viewing other's private posts that shouldn't be visible (i.e. contributor viewing author's private posts)

Addresses #375


sboisvert:
@rebeccahum wanna give this a test and if it works (which I suspect it does) we'll merge it?

PR #391, opened 2016-12-10
Bug Fix: Co-Author Order Not Retained in WordPress 4.7

mslinnea:
After upgrading to WordPress 4.7, the order of the co-authors always displayed alphabetical.  Changing the order in the admin interface stopped working.  

It looks like `get_the_terms` is defaulting to order by name.  Unfortunately, `get_the_terms` does not accept an orderby parameter, but the function it calls, `wp_get_object_terms` does.

Since  `wp_get_object_terms` is an uncached function, I've added caching.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -116,6 +116,9 @@ class CoAuthors_Plus {
 		// Support infinite scroll for Guest Authors on author pages
 		add_filter( 'infinite_scroll_js_settings', array( $this, 'filter_infinite_scroll_js_settings' ), 10, 2 );
 
+		// Delete CoAuthor Cache on Post Save
+		add_action( 'save_post', array( $this, 'clear_cache') );
+		add_action( 'delete_post', array( $this, 'clear_cache') );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


co-authors-plus.php
```diff
@@ -1469,6 +1472,15 @@ public function filter_jetpack_open_graph_tags( $og_tags, $image_dimensions ) {
...
 		return apply_filters( 'coauthors_open_graph_tags', $og_tags );
 	}
+
+	/**
+	 * Callback to clear the cache on post save and post delete.
+	 *
+	 * @param $post_id The Post ID.
+	 */
+	public function clear_cache( $post_id ) {
+		wp_cache_delete( 'coauthors_post_' . $post_id );
```
goldenapples: I believe that this call to `wp_cache_delete` will need to pass the cache group `'co-authors-plus'` as its second argument, in order to ensure that it is cleared properly - since the cache item was added in the `'co-authors-plus'` cache group.

mslinnea: Good spot!

philipjohn:
Thanks @mslinnea ! I'm going to merge this now and add the cache group param.

danielbachhuber:
> After upgrading to WordPress 4.7, the order of the co-authors always displayed alphabetical. Changing the order in the admin interface stopped working.

This issue is actually https://core.trac.wordpress.org/ticket/40496

Issue #390, opened 2016-12-10
Changing sort order of coauthors broken in WordPress ^4.7.x


dsifford:
Hello there,

It appears that ever since the update to WordPress 4.7, changing sort order of coauthors no longer works properly.

I confirmed this on a clean install of WordPress with only co-authors-plus installed. It's most definitely an issue with this plugin.

Happy to help if needed. Thanks so much.

blackwood:
Indeed, can replicate. However, changing line 17 in `get_coauthors` in the file `template-tags.php` from:

     $coauthor_terms = get_the_terms( $post_id, $coauthors_plus->coauthor_taxonomy );

to: 

    $coauthor_terms = wp_get_post_terms( $post_id, $coauthors_plus->coauthor_taxonomy, array( 
      'orderby' => 'term_order', 
    ) );

...fixes the issue for me.

The untested theory we have over here is that the default args provided by `wp_get_object_terms` could have changed in 4.7, which is how `get_the_terms` updates the cache if there's no cached terms available -- so whenever `get_coauthors` is called, it ignores the custom `term_order`, using `name` instead.

blackwood:
...but it could be something else entirely. It would definitely be great to get the plugin author's input here.

blackwood:
#392 and #391 look like they address the same issue?

dsifford:
Thanks for the comments @blackwood -- I'll give your fix a whirl for the time being while waiting for the maintainers' input on this 😄

dsifford:
@blackwood Your fix isn't doing the trick here for me. Hmm.. Looks like the two open PRs involve clearing object cache?

Both PRs break the build so I'm curious what negative implications doing it @mslinnea's way has (if any?)

mslinnea:
Thanks for linking to my PRs.  I don't know what's up with Travis, as the errors seem unrelated to my code changes.

I tried @blackwood's fix and it does work for me, but `wp_get_post_terms` is [listed as an uncached function](https://vip.wordpress.com/documentation/caching/uncached-functions/), which is why I steered clear of using it.

dsifford:
@mslinnea Interesting. I wasn't aware of the cached/uncached functions. Thanks for the resource!

Also, perhaps @blackwood's fix will work for me... I SSH'ed into the server and made the change extremely quickly with vim so chances are I could have made a typo! I'll try it again in a few.

blackwood:
[...]

PR #392, opened 2016-12-12
Bug Fix: Co-Author Order Not Retained in WordPress 4.7 (filter get_terms_orderby)

mslinnea:
This is the fix I'm currently using for issue #390.  Since this is a filter, it can be added to your theme code, if you want to solve this issue before the plugin gets fixed.  Just remember to prefix the function if you're using it outside of a class :smile:


philipjohn:
Closing this in favour of #391

xcentric-solutions:
This bug has made a come-back in WordPress 4.8.2.

PR #400, opened 2017-01-17
Reduce amount of sleep

sboisvert:
This command was taking multiple days to run for some sites. I think reducing the amount of sleep here is reasonable. No need to pause for 5s at each 20 queries.


trepmal:
Merged master to try and clean up some of the noise in the tests, but I think there's other stuff going on there...

PR #404, opened 2017-03-11
Prepare v3.2.2 release

philipjohn:
Full changeset for this release: https://github.com/Automattic/Co-Authors-Plus/compare/69f58e2b84927643fc382dba45d935f0477bbf7d...bd133868490f014733ebb9b7b37be65cdf17963b

Issue #398, opened 2016-12-21
Failing test: `test__author_queried_object_fix`


philipjohn:
> There was 1 failure:
> 
> 1) Test_Author_Queried_Object::test__author_queried_object_fix
> is_archive, is_author is expected to be true. is_404 is expected to be false.
> Failed asserting that false is true.
> 
> /srv/www/wordpress-develop/tests/phpunit/includes/testcase.php:616
> /srv/www/wpmu-subdirectory/wp-content/plugins/co-authors-plus/tests/test-author-queried-object.php:73

This test has been failing since it's introduction. It fails at this point:

```
// Remove the user from the blog
remove_user_from_blog( $author2, $blog2 );

/**
 * Author 2 was removed from the blog, but still a published author
 */
$this->go_to( get_author_posts_url( $author2 ) );
$this->assertQueryTrue( 'is_author', 'is_archive' );
```

The issue is that `is_404` is true. It should be false, because the author archive page should still display one post.

A manual check of the flow in this test is successful - the author archive displays the one post as expected.

As an additional check I deleted the user from the blog using `remove_user_from_blog()` via WP CLI instead of through the admin, just to check if there was a difference. There was none.

philipjohn:
There's a bunch of DB errors:

[...]

PR #406, opened 2017-03-20
More term ordering fixes

philipjohn:
Following on from #390 we also needed to revert the change introduce in #367 which meant that the CLI commands would also be affected by the bug.

This change moves the fix from #391 into it's own helper function - `cap_get_coauthor_terms_for_post()` to avoid unnecessary code duplication for the cache getting/setting. It updates the existing uses of the cache to use the helper, and fixes the CLI commands' use of `get_the_terms()` to use our new helper instead.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1643,7 +1643,7 @@ function cap_get_coauthor_terms_for_post( $post_id = false ) {
 	global $coauthors_plus;
 
 	$cache_key = 'coauthors_post_' . $post_id;
-	$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' )
+	$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' );
 
 	if ( false === $coauthor_terms ) {
 		$coauthor_terms = wp_get_object_terms( $post_id, $coauthors_plus->coauthor_taxonomy, array(
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -119,6 +119,7 @@ class CoAuthors_Plus {
 		// Delete CoAuthor Cache on Post Save & Post Delete
 		add_action( 'save_post', array( $this, 'clear_cache') );
 		add_action( 'delete_post', array( $this, 'clear_cache') );
+		add_action( 'set_object_terms', array( $this, 'clear_cache_on_terms_set' ), 10, 6 );
 	}
 
 	/**
@@ -1481,6 +1482,23 @@ class CoAuthors_Plus {
 	public function clear_cache( $post_id ) {
 		wp_cache_delete( 'coauthors_post_' . $post_id, 'co-authors-plus' );
 	}
+
+	/**
+	 * Callback to clear the cache when an object's terms are changed.
+	 *
+	 * @param $post_id The Post ID.
+	 */
+	public function clear_cache_on_terms_set( $object_id, $terms, $tt_ids, $taxonomy, $append, $old_tt_ids ) {
+
+		// We only care about the coauthors taxonomy
+		if ( $this->coauthor_taxonomy !== $taxonomy ) {
+			return;
+		}
+
+		wp_cache_delete( 'coauthors_post_' . $object_id, 'co-authors-plus' );
+
+	}
+
 }
 
 global $coauthors_plus;
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1475,6 +1475,38 @@ class CoAuthors_Plus {
 	}
 
 	/**
+	 * Retrieve a list of coauthor terms for a single post.
+	 *
+	 * Grabs a correctly ordered list of authors for a single post, appropriately
+	 * cached because it requires `wp_get_object_terms()` to succeed.
+	 *
+	 * @param int $post_id ID of the post for which to retrieve authors.
+	 * @return array Array of coauthor WP_Term objects
+	 */
+	public function get_coauthor_terms_for_post( $post_id ) {
+
+		if ( ! $post_id ) {
+			return array();
+		}
+
+		$cache_key = 'coauthors_post_' . $post_id;
+		$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' );
+
+		if ( false === $coauthor_terms ) {
+			$cached = wp_get_object_terms( $post_id, $this->coauthor_taxonomy, array(
+				'orderby' => 'term_order',
+				'order' => 'ASC',
+			) );
+			// Cache an empty array if the taxonomy doesn't exist.
+			$coauthor_terms = ( is_wp_error( $cached ) ) ? array() : $cached;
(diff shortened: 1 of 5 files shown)


co-authors-plus.php
```diff
@@ -1624,3 +1642,37 @@ function cap_filter_comment_moderation_email_recipients( $recipients, $comment_i
...
+	$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' );
+
+	if ( false === $coauthor_terms ) {
+		$coauthor_terms = wp_get_object_terms( $post_id, $coauthors_plus->coauthor_taxonomy, array(
+			'orderby' => 'term_order',
+			'order' => 'ASC',
+		) );
+		wp_cache_set( $cache_key, $coauthor_terms, 'co-authors-plus' );
+	} else {
+		$coauthor_terms = array();
```
mjangda: We should remove the `else` and just do the variable init + assignment before the `if`

co-authors-plus.php
```diff
@@ -1624,3 +1642,37 @@ function cap_filter_comment_moderation_email_recipients( $recipients, $comment_i
...
+/**
+ * Retrieve a list of coauthor terms for a single post.
+ *
+ * Grabs a correctly ordered list of authors for a single post, appropriately
+ * cached because it requires `wp_get_object_terms()` to succeed.
+ *
+ * @param int $post_id ID of the post for which to retrieve authors.
+ * @return array Array of coauthor WP_Term objects
+ */
+function cap_get_coauthor_terms_for_post( $post_id = false ) {
```
mjangda: Why do we have a default of `false` if we just return early in that case?

co-authors-plus.php
```diff
@@ -1624,3 +1642,37 @@ function cap_filter_comment_moderation_email_recipients( $recipients, $comment_i
...
+
+	$cache_key = 'coauthors_post_' . $post_id;
+	$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' );
+
+	if ( false === $coauthor_terms ) {
+		$coauthor_terms = wp_get_object_terms( $post_id, $coauthors_plus->coauthor_taxonomy, array(
+			'orderby' => 'term_order',
+			'order' => 'ASC',
+		) );
+		wp_cache_set( $cache_key, $coauthor_terms, 'co-authors-plus' );
```
mjangda: I think we need to catch (and not cache) when a `WP_Error` is returned by `wp_get_object_terms` as that could lead to some "interesting" bugs.

co-authors-plus.php
```diff
@@ -1624,3 +1642,37 @@ function cap_filter_comment_moderation_email_recipients( $recipients, $comment_i
...
+	} else {
+		$coauthor_terms = array();
+	}
+
+	return $coauthor_terms;
```
mjangda: Thoughts on moving the logic for this function inside the coauthors_plus class (and this function can just call that)?

co-authors-plus.php
philipjohn: I just like being explicit :) but I'll remove it.

co-authors-plus.php
philipjohn: Hmm, I guess we missed this in the original fix PR too. We'll get a `WP_Error` when the taxonomy doesn't exist, so I'll cache an empty array in this case. Otherwise, we could end up with lots of `wp_get_object_terms()` and the resultant unpleasant queries.

co-authors-plus.php
philipjohn: You're right, I considered that and got lazy.

co-authors-plus.php
philipjohn: Yeah, I did think about putting this into the class but thought it silly to force everyone to do `global $coauthors` just to make a simple function call. Your idea solves that though, so I'll do that!

philipjohn:
Okay, all good. The failing test is still just that one from the earlier `JOIN` change.

trepmal:
For the failed PHP7 tests: http://stackoverflow.com/a/42561590/7618202, https://core.trac.wordpress.org/ticket/39822

trepmal:
Failing test comments: https://github.com/Automattic/Co-Authors-Plus/issues/398#issuecomment-288559694

trepmal:
934a261 updates the test to prevent subsite tables from being created as temporary, thus allowing the test to pass.

ping @mjangda, @sboisvert, @dartiss 

related: #398, #404

philipjohn:
Thanks @trepmal !

Looks like those remaining issues are thanks to [this core bug](https://core.trac.wordpress.org/ticket/39822). Given that, are we safe to merge this @mjangda ?

co-authors-plus.php
```diff
@@ -1474,13 +1475,62 @@ public function filter_jetpack_open_graph_tags( $og_tags, $image_dimensions ) {
...
+		$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' );
+
+		if ( false === $coauthor_terms ) {
+			$cached = wp_get_object_terms( $post_id, $this->coauthor_taxonomy, array(
+				'orderby' => 'term_order',
+				'order' => 'ASC',
+			) );
+			// Cache an empty array if the taxonomy doesn't exist.
+			$coauthor_terms = ( is_wp_error( $cached ) ) ? array() : $cached;
+			wp_cache_set( $cache_key, $coauthor_terms, 'co-authors-plus' );
```
mjangda: I don't think we should be caching when there is an error.

co-authors-plus.php
```diff
@@ -1474,13 +1475,62 @@ public function filter_jetpack_open_graph_tags( $og_tags, $image_dimensions ) {
...
+
+		if ( ! $post_id ) {
+			return array();
+		}
+
+		$cache_key = 'coauthors_post_' . $post_id;
+		$coauthor_terms = wp_cache_get( $cache_key, 'co-authors-plus' );
+
+		if ( false === $coauthor_terms ) {
+			$cached = wp_get_object_terms( $post_id, $this->coauthor_taxonomy, array(
```
mjangda: Minor: `$cached` is not quite the best name here. We should probably swap the var names `$coauthor_terms` and `$cached`.

```
 + $cache_coauthors = wp_cache_get( $cache_key, 'co-authors-plus' );
 +
 +		if ( false === $coauthor_terms ) {
 +			$coauthor_terms = wp_get_object_terms( $post_id, $this->coauthor_taxonomy, array(
...
```

co-authors-plus.php
mjangda: Addressed in fe48900

mjangda: @philipjohn let me know if that changeset makes sense to you and we can merge.

philipjohn: My worry about not caching errors is that a simple typo could easily result in `wp_get_object_terms()` triggering a lot more than we'd like, and the uncached queries that comes with it.

mjangda:
Remaining test failures are because the version of Travis running for CAP doesn't have PHP5.2 and PHPCS errors.

PR #412, opened 2017-05-22
It should be 'continue' instead of return false at some situation.

### LINKED ISSUE #411: wp_notify_postauthor() doesn't notify authors in some situations
The function does a `foreach` loop over all of the authors of the post. But inside the `foreach` loop, `return false` is used instead of `continue` which means that subsequent authors are not notified.

https://github.com/Automattic/Co-Authors-Plus/blob/525f211c38b56e6b52bbb1e8b88e8f408dc411be/co-authors-plus.php#L1555-L1574

Changing these three `return false;`s to `continue;` would fix it.

See also: #403

emgk:
fixes #411


GaryJones:
[Last commit](https://github.com/Automattic/Co-Authors-Plus/pull/412/commits/48341b2396af8b98a55c07282a246f68a8fb1dbf) is relevant. This looks good to review.

Issue #53, opened 2012-06-12
Create new user or guest author while assigning to a post


danielbachhuber:
[Originally from the forums](http://wordpress.org/support/topic/plugin-co-authors-plus-create-author-instantly), it would be neat if you could create a user or guest author on the fly while assigning to a post. For example, if you were to do a search for a user that doesn't exist, there would be a small prompt to create a new user/ guest author in a modal or the like.

emgk:
Hello @philipjohn and @danielbachhuber as per the enhancement request I added the code to create a new guest author if doesn't exist while assigning to the post on post edit page.

Please check the PR above and do let me know if any changes needed! :)

GaryJones:
Picking this up after a long pause. PR #413 was the only attempt at this in the tracker; I've just closed it because it could no longer be rebased onto current `develop` (the autosuggest pipeline has been substantially rewritten since 2017), but the underlying user problem is still very much worth addressing.

Some thoughts on shape, in case anyone picks this up.

**Where the friction actually is**

The current flow to add a brand-new guest author is roughly: leave the post screen → Users → Guest Authors → Add new → fill the full form → save → return to the post → retype → select → save. That is a lot of context switching for what feels like one action, and the cost is paid most by editorial workflows that handle frequent one-off contributors.

**"No match" isn't always "create"**

The naive framing is "if the search returns nothing, offer to create". But "no match" can also mean:

1. The user typed it wrong.
2. The person exists as a regular WordPress user but has not been added as an author/guest author yet.
3. The person exists as a guest author under a slightly different spelling.

A good solution should distinguish at least case 2 from a true "create new", because creating a fresh, unlinked guest author when a matching WP user already exists is the wrong shape of data, and CAP already supports linking a guest author to a WP user from the guest-author edit screen.

**A few options, in ascending effort**

1. **Deep link with prefill.** When the autosuggest returns nothing, render a "Create new guest author" link in the dropdown that opens `users.php?page=view-guest-authors&action=add-new&display_name=<typed>` in a new tab. Reuses every existing piece of validation, capability gating, and "link to WP user" logic. Probably 30 lines of JS and a tiny PHP prefill on the new-guest-author screen. Resolves most of the friction for a tiny surface area.

[...]

PR #413, opened 2017-05-23
Added enhancement to create user instantly if not exists.

emgk:
Implemented functionality to create user while assigning author on edit post page. #53


GaryJones:
Thank you for this contribution, @emgk, and apologies for the very long wait on a response.

The underlying user problem you were solving is real and still worth addressing: creating a guest author currently requires leaving the post screen, filling out a full form, then coming back to assign them, which is a lot of context switching for what feels like a single action.

Unfortunately, the patch itself can no longer be rebased onto `develop` in any meaningful way. Since 2017 the autosuggest pipeline has been substantially rewritten:

- `CoAuthors_Plus` has moved out of the root file into `php/class-coauthors-plus.php`.
- `ajax_suggest()` now returns JSON via `wp_send_json()` rather than pipe-delimited strings, so the `"New | ..."` sentinel approach no longer fits the response shape.
- The classic-editor metabox JS has been rewritten around jQuery UI autocomplete, and a React-based author panel is now the path forward for new UI work.
- A REST controller (`php/api/endpoints/class-coauthors-controller.php`) is the modern home for write operations like this.

There are also a few aspects of the original implementation we'd want to revisit before merging anything similar: a capability check (not just a nonce) on the create endpoint, sanitising the derived `user_login` via `sanitize_user()`, replacing the native `prompt()` and `alert()` calls with a proper modal for accessibility, handling collisions visibly in the UI, and engaging with the existing "link to an existing WP user" affordance rather than always creating a fresh, unlinked guest author.

For these reasons I'm going to close this PR. I'll open a follow-up issue describing the user problem and a few options for solving it cleanly on the current architecture, ranging from a small "deep link to the new guest author screen with the typed text pre-filled" to a fuller modal-based create flow over REST. If you'd like to take another run at it on top of the current code, you'd be very welcome.

Thanks again for raising the idea in the first place.

Issue #69, opened 2012-10-26
Deleting a user can mistakenly assign new bylines


danielbachhuber:
To reproduce:
- User A creates a new post but publishes it under Guest Author A byline
- Admin deletes User A and assigns all of their posts to User B
- Post with Guest Author A byline will now get User B added to the byline too.

The culprit is:

`$post_ids = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author = %d", $delete_id ) );`

This gets post IDs where the user doesn't necessarily have the byline.

rebeccahum:
This is also because guest author account A (previously linked to user A) still exists.

When we delete users linked to guest author accounts, perhaps we can add a default option to keeping existing post bylines mapped to the guest author account.

rebeccahum:
This is also because guest author account A (previously linked to user A) still exists.

When we delete users linked to guest author accounts, perhaps we can add a default option to keeping existing post bylines mapped to the guest author account.

rebeccahum:
This is also because guest author account A (previously linked to user A) still exists.

When we delete users linked to guest author accounts, perhaps we can add a default option to keeping existing post bylines mapped to the guest author account.

rebeccahum:
This is also because guest author account A (previously linked to user A) still exists.

When we delete users linked to guest author accounts, perhaps we can add a default option to keeping existing post bylines mapped to the guest author account.

philipjohn:
Fixed by #414

PR #414, opened 2017-05-24
Remove associated guest user when mapped user id deleted.

### LINKED ISSUE #120: Remove user association from guest author when user is deleted / removed
When a user is deleted or removed from a site, any guest author associations they have should be deleted.

mdchiragpatel:
Fixes #120


rebeccahum:
This also fixes issue #69.

cc: @philipjohn

Issue #410, opened 2017-04-05
(CLI) `create-terms-for-posts` fails if a post doesn't have any author terms


trepmal:
If you run the WP-CLI command `wp co-authors-plus create-terms-for-posts` it'll fail once it runs into an author-less post.

In https://github.com/automattic/Co-Authors-Plus/blob/master/php/class-wp-cli.php#L52, there's this:

```
$terms = cap_get_coauthor_terms_for_post( $single_post->ID );
if ( empty( $terms ) ) {
	WP_CLI::error( sprintf( 'No co-authors found for post #%d.', $single_post->ID ) );
}
```
`cap_get_coauthor_terms_for_post()` will return an empty array for either 1) no author terms for post or 2) WP_Error

So when there are no terms, the CLI returns an error, which exits and kills the script. Not sure if the intent of the `error()` was to serve as a notice and continue.

I see three possible fixes

1. Use `wp_get_post_terms()` instead of the `cap_get_coauthor_terms_for_post()` wrapper, for better or worse, this skips caching. e.g.  

  ```
	$terms = wp_get_post_terms( $single_post->ID, $coauthors_plus->coauthor_taxonomy );
	if ( is_wp_error( $terms ) ) {
		WP_CLI::error( $terms->get_error_message() );
	}
  ``` 
2. Change `error()` to `line()` so it's non-blocking
3. Add an option in `cap_get_coauthor_terms_for_post()` to return the WP_Error instead of an empty array

mjangda:
> Change error() to line() so it's non-blocking

Or `WP_CLI::warning`

trepmal:
I'm not sure `warning` is appropriate in context, at least not in the case of `$terms` simply being empty. But possibly in combination with checking `is_wp_error()`.

david-binda:
Neither the `line()` nor `warning()` is an appropriate fix without other changes, as the the command would then simply do nothing, as it would read following:

```
if ( empty( $terms ) ) {
    WP_CLI::line( ... );
}
if ( ! empty( $terms ) ) {
WP_CLI::line( ... );
    continue;
}
```

Thus I would propose using the `wp_get_post_terms()` and `is_wp_error` check as it used to be in place before the https://github.com/Automattic/Co-Authors-Plus/commit/4aea6b153941b34da5e9eb848283ffb0885288d2 . Or allowing the `cap_get_coauthor_terms_for_post ` returning WP_Error, in case it's appropriate.

[...]

PR #420, opened 2017-05-30
Fixed WP CLI create-terms-for-posts if no co-authors found

rebeccahum:
As per #410,  script was stopping after no co-authors was found. Now script should display error message and continue to the next post.


rebeccahum:
Ok, I removed it.

philipjohn:
@trepmal Given the comments on #410 can you clarify whether this is necessary or we should revert the merge? Thanks!

Issue #401, opened 2017-03-03
No longer available for Pages under the Quick Edit link


catchmyfame:
I'm testing WP 4.7.2 and discovered that when I look at the list of pages, if I click "Quick Edit" for any page, the co-authors plugin isn't there -- it's just the standard WP downdown list of authors. It works fine for posts and custom post types of type post, just not pages.

If I open a page for editing the co-authors plus author options are there and work as expected.

philipjohn:
This is expected as the plugin currently doesn't support setting co-authors from the Quick Edit panel. Open to Feature/Pull Requests though!

philipjohn:
Oh, ha - completely didn't see the PR!

catchmyfame:
@philipjohn But it did in recent versions, so I'd call that a bug no?

philipjohn:
Not that I'm aware of. If you see it in a previous version, please do provide details. Otherwise, see #422 which adds Quick Edit support.

catchmyfame:
@philipjohn  Oh, it totally did used to exist. Then back around the beginning of March (maybe earlier) I believe an update to WP caused it to no longer be on the Quick Edit page. I was just posting a bug report about it, not asking for a new feature.

PR #422, opened 2017-05-31
Pages archive now displays coauthors and quick edit works

rebeccahum:
As per #401, admin can now use quick edit for co-authors on pages and co-authors on pages archive display correctly.

get_post_type() was returning false on pages archive which broke is_post_type_enabled and current_user_can_set_authors()

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -293,11 +293,9 @@ class CoAuthors_Plus {
 	public function is_post_type_enabled( $post_type = null ) {
 
 		if ( ! $post_type ) {
-			if ( is_admin() ) {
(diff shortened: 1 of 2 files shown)


trepmal:
When you save after using Quick Edit, and the table row refreshes via ajax, the author column reverts to Core's column, instead of Co-Author's

quick-edit

A page refresh shows that the update worked, it's just the UI that's not reflecting the change in the moment.

rebeccahum:
@trepmal Fixed -- should work after you save on quick edit.

rebeccahum:
@trepmal Fixed -- should work after you save on quick edit.

Issue #429, opened 2017-06-23
create-guest-authors should show some sense of progress


trepmal:
The `create-guest-authors` CLI command is silent till the end. I'm running it on a site now, and I get the feeling it's going to be a while, but I have know idea if it'll actually be 5 minutes or 5 hours.

It should show either a progress bar or line-by-line success/failure output (even if with a `--verbose` flag) to give the user some indication of finish time.

trepmal:
(By the way, the site had 2500+ users, and it took over 6 hours for the command to complete)

sboisvert:
Fixed in #431

PR #431, opened 2017-06-28
Added progress to create-guest-authors so users have an idea of how long it will take

rebeccahum:
As per #429, WP CLI command `create-guest-authors` does not display any sense of progress until the end, so the user has no idea how long the command will take to run.  I've added a progress notification for every user processed with % complete.


trepmal:
It looks like this will be very verbose. I think we should look at https://make.wordpress.org/cli/handbook/internal-api/wp-cli-utils-make-progress-bar/

rebeccahum:
Oh, that is a better one.  I revised the changes as per your suggestion!

Issue #77, opened 2012-11-08
Guest Authors: Indicate the number of posts to be modified when deleting


danielbachhuber:
As an improvement to the user experience, the view for deleting a guest author should indicate how many posts will be affected by deleting a guest author. A plus would be showing some of the titles.

If there aren't any posts associated with the guest author, make it easier to delete the profile

philipjohn:
Fixed by #432

PR #432, opened 2017-06-29
Deleting guest authors is less confusing

rebeccahum:
As per #77, if there are 0 posts, it won't give the user any options on confirming the deletion of a guest author.  It also displays the amount of posts associated with a guest author on deletion, so that the user knows how many posts the guest author has.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -479,19 +479,24 @@ class CoAuthors_Guest_Authors
 			echo '<p>#' . esc_html( $guest_author->ID . ': ' . $guest_author->display_name ) . '</p>';
 			// display wording differently per post count
 			if ( 0 === $count ) {
-				$post_count_message = '<p>' . esc_html(  sprintf( __( 'There are no posts associated with this guest author.', 'co-authors-plus' ), strtolower( $this->labels['singular'] ) ) ) . '</p>';
+				$post_count_message = '<p>' . sprintf( __( 'There are no posts associated with this guest author.', 'co-authors-plus' ), strtolower( $this->labels['singular'] ) ) . '</p>';
 			}
 			else {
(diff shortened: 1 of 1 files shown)


php/class-coauthors-guest-authors.php
```diff
@@ -463,36 +463,70 @@ function view_guest_authors_list() {
...
+				$note = '<p class="description">' . esc_html( sprintf( __( "Note: If you'd like to delete the %s and all of their posts, you should delete their posts first and then come back to delete the %s.", 'co-authors-plus' ), strtolower( $this->labels['singular'] ), strtolower( $this->labels['singular'] ) ) ) . '</p>';
+				if ( 1 === $count ) {
+					$post_count_message = '<p>' . esc_html(  sprintf( __( 'There is %d post associated with this guest author. What should be done with the post assigned to this %s?', 'co-authors-plus' ), $count, strtolower( $this->labels['singular'] ) ) ) . '</p>';
+				}
+				else {
+					$post_count_message = '<p>' . esc_html(  sprintf( __( 'There are %d posts associated with this guest author. What should be done with the posts assigned to this %s?', 'co-authors-plus' ), $count, strtolower( $this->labels['singular'] ) ) ) . '</p>';
+				}
+				$post_count_message .= $note;
+			}
+			echo $post_count_message;
```
philipjohn: It's always best to [late escape](https://vip.wordpress.com/documentation/vip/best-practices/security/validating-sanitizing-escaping/#always-escape-late) to avoid future changes making the output unsafe. This also simplifies the above code by removing the need for multiple `esc_html()` calls.

rebeccahum: Done. I ended up using `wp_kses()` because I wanted to keep the `<p>` tags for formatting purposes.

Issue #118, opened 2013-04-01
Add an author type argument to coauthors_wp_list_authors()


danielbachhuber:
Some users would like to be able to only display users or guest authors

http://wordpress.org/support/topic/can-coauthors_wp_list_authors-list-only-guest-authors-and-not-wp-users?replies=2

PR #436, opened 2017-06-30
coauthors_wp_list_authors() has option to list only guest authors now

rebeccahum:
As per #118, with the template tag `coauthors_wp_list_authors()` it included all users and guest authors.  Some wanted the feature to only list guest authors because the core function `wp_list_authors()` already lists all users with posts.  Now with this fix, if you pass in the `$args` of `guest_authors_only` to be true, then the template tag will list only guest authors.

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -448,6 +448,7 @@ function coauthors_wp_list_authors( $args = array() ) {
 		'style'            => 'list',
 		'html'             => true,
 		'number'           => 20, // A sane limit to start to avoid breaking all the things
+		'guest_authors_only' => false
 	);
 
 	$args = wp_parse_args( $args, $defaults );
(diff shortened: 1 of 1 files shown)


template-tags.php
```diff
@@ -448,6 +448,7 @@ function coauthors_wp_list_authors( $args = array() ) {
 		'style'            => 'list',
 		'html'             => true,
 		'number'           => 20, // A sane limit to start to avoid breaking all the things
+		'guest_authors_only' => false
```
philipjohn: We should keep the spacing consistent here, and the doc block will need updating to include this new argument. (The formatting of the doc block is also not following the standards, so we can sort that at the same time.

template-tags.php
```diff
@@ -472,6 +473,17 @@ function coauthors_wp_list_authors( $args = array() ) {
 	}
 
 	$authors = apply_filters( 'coauthors_wp_list_authors_array', $authors );
```
philipjohn: Should the filter be before or after we remove the WP authors?

template-tags.php
```diff
@@ -472,6 +473,17 @@ function coauthors_wp_list_authors( $args = array() ) {
...
+	
+	// only show guest authors if the $args is set to true
+	if ( $args['guest_authors_only'] ) {
+		$guest_authors = [];
+		foreach ( $authors as $author ) {
```
philipjohn: Rather than do another foreach here, could we instead skip adding authors to the array within the foreach at [line 464](https://github.com/rebeccahum/Co-Authors-Plus/blob/9b57a4344c75e0825e5a30118721ae6a9f94cef9/template-tags.php#L464)

template-tags.php
rebeccahum: done

template-tags.php
rebeccahum: yes

template-tags.php
rebeccahum: done

Issue #228, opened 2014-10-02
Recent changes to bylines/author assignments don't get reflected properly in post counts in admin UI


ryanmarkel:
Whenever changes are made to authors or user account links are created, counts get messed up and can cause users to panic. :)

You can reproduce this somewhat easily:
1. Have a user account (no guest author/byline) that has authored posts. You'll see a count in the Users panel.
2. Create a new byline for that user.
3. Link that byline to that user account.
4. The counts for both the byline and the user account for number of posts will read zero.

It's not immediately clear how to adjust or fix this. Updating terms using the CLI tool has no effect.

rebeccahum:
Re-opening because reverted #438 per https://github.com/Automattic/Co-Authors-Plus/commit/a645d596357c0b1c84c57a5a5ab9925f669050ee so this issue still exists.

PR #438, opened 2017-07-05
Accurate Guest Author post count on linked accounts

rebeccahum:
As per #228, once you link a guest author account to a user with pre-existing posts, the Guest Authors "Posts" column does not update to reflect an accurate count.  This patch fixes the issue.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1344,11 +1344,17 @@ class CoAuthors_Plus {
 		if ( false !== ( $term = wp_cache_get( $cache_key, 'co-authors-plus' ) ) ) {
 			return $term;
 		}
-
-		// See if the prefixed term is available, otherwise default to just the nicename
-		$term = get_term_by( 'slug', 'cap-' . $coauthor->user_nicename, $this->coauthor_taxonomy );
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -1344,11 +1344,17 @@ public function get_author_term( $coauthor ) {
...
 		}
-
-		// See if the prefixed term is available, otherwise default to just the nicename
-		$term = get_term_by( 'slug', 'cap-' . $coauthor->user_nicename, $this->coauthor_taxonomy );
-		if ( ! $term ) {
-			$term = get_term_by( 'slug', $coauthor->user_nicename, $this->coauthor_taxonomy );
+		
+		// use linked user for accurate post count
+		if ( ! empty ( $coauthor->linked_account ) ) {
+			$term = get_term_by( 'slug', 'cap-' . $coauthor->linked_account, $this->coauthor_taxonomy );
```
trepmal: Should this check for a non-prefixed term if `$term` is false? (like in the `else` section)

rebeccahum: Good catch! Added the check, @trepmal

Issue #351, opened 2016-04-18
"Author Archives" Name Incorrect


srguglielmo:
I have Co-Authors-Plus from GitHub (cloned about a week ago) running on WP 4.4. I've modified the theme to use coauthors_posts_links() and all seems to be working well except for one thing.

Consider this post here: https://i.imgur.com/2J9bBIv.png

When clicking a name from there (either guest-authors or regular author), it brings me to a page that lists all the posts by that user. It works correctly and the proper posts are displayed.

However, at the top of that page, it says `Author Archives: Stephen Robert Guglielmo` regardless of which author I click (Jane Doe, Albert Einstein, or myself).

Additionally, when multiple real users create posts on a site, the `Author Archives` line displays the name of the user **who submitted the most recent post**, regardless of the name you click.

To explain this more:
Real **UserA** makes a post with various co-authors. `Author Archives` shows **UserA**, regardless of what name is clicked (and regardless of what post it's clicked on from).

Now, real **UserB** makes a post with various co-authors. `Author Archives` now shows **UserB**, regardless of what name is clicked (and regardless of what post it's clicked on from).

It shows the name of the real user who **most recently posted an item** on the site. I expected it to display the name of the user (real or co-authors) who was clicked.

If I disable Co-Authors-Plus, the "Author Archives" works properly by showing the name of the user I clicked on.

philipjohn:
Can you share the code that prints out `Author Archives: Stephen Robert Guglielmo` please?

PR #441, opened 2017-07-07
Filter author archive

rebeccahum:
As per #351, the author archive posts page title did not reflect the correct author.  This patch solves the issue!

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -128,6 +128,9 @@ class CoAuthors_Plus {
 		add_action( 'save_post', array( $this, 'clear_cache') );
 		add_action( 'delete_post', array( $this, 'clear_cache') );
 		add_action( 'set_object_terms', array( $this, 'clear_cache_on_terms_set' ), 10, 6 );
+
+		// Change Author Archive header to plural
+		add_filter( 'get_the_archive_title', array( $this, 'filter_author_archive_title'), 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


SeanDS:
I think this is actually not fixing the underlying problem. The problem is that the query doesn't contain the author requested. I think the problem is in `fix_author_page`. Sometimes the query is for an author id (`?author=[id]`), sometimes it is for an author name (`/author/authorname`). Both must be checked.

I have forked CAP for my own purposes, so my code is slightly different and will need adapted, but this works for me. In `fix_author_page`, replace...
```php
		$author_name = sanitize_title( get_query_var( 'author_name' ) );
		if ( ! $author_name ) {
			return;
		}
```

...with
```php
		$author_id = absint( get_query_var( 'author' ) );
		$author_name = sanitize_title( get_query_var( 'author_name' ) );

		if ( isset( $author_id ) ) {
			// get author by id
			$author = get_user_by( 'id', $author_id );
		} elseif ( isset( $author_name ) ) {
			// get author by specified name
			$author = get_user_by( 'login', $author_name );
		} else {
			// no query variable was specified; not much we can do
			return;
		}
```

(need to change `get_user_by` to whatever CAP uses to get coauthors)

This means that the `$authordata` variable is an author object and therefore the last block in the function works as intended. I think this should mean the `get_the_archive_title` filter won't be needed.

TheCrowned:
Hi Daniel,
thanks for reporting this! The issue was already fixed in #535 - it has not yet been released as a public update on WordPress.org, but the master branch of this repository contains the fix :)

PR #443, opened 2017-07-12
Prevent doubling up on filters which breaks the the_author_posts_link…

### LINKED ISSUE #279: Incompatibility when `coauthors_auto_apply_template_tags` set to true

trepmal:
Prevent doubling up on filters which breaks the `the_author_posts_link` template tag. Fixes #279 and subsequently the broken test in the pending merge #301

### DIFF
--- a/php/class-coauthors-template-filters.php
+++ b/php/class-coauthors-template-filters.php
@@ -20,6 +20,7 @@ class CoAuthors_Template_Filters {
 	}
 
 	function filter_the_author_posts_link() {
+		remove_filter( 'the_author', array( $this, 'filter_the_author' ) );
 		return coauthors_posts_links( null, null, null, null, false );
 	}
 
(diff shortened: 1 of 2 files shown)


trepmal:
May need to look for other places the author names may be doubled

rebeccahum:
@trepmal I'm still getting duplicate authors with `coauthors_posts_links()` in my theme
capture

philipjohn:
@trepmal Any chance you're able to look back at this?

philipjohn:
Closing in favour of #474

Issue #132, opened 2013-08-23
website vs url in coauthors_links_single function


frankar:
In template_tags.php, function coauthors_links_single
for co-authors works if I substitute get_the_author_meta('url') with
get_the_author_meta('website') on lines 327 and 329

If my author is a normal user is right get_the_author_meta('url')

philipjohn:
Can you elaborate on what's broken here please? Are you saying that `get_the_author_meta('url')` gives an unexpected result?

rebeccahum:
Hi @frankar, just following up if you are able to confirm what exactly the issue you are experiencing is?

trepmal:
1. Assign standard WP user to post
1a. Make sure WP user has Website field set in profile
2. Assign guest author to post
2a. Make sure guest author has Website set in guest author profile
3. Use `coauthors_links()`

With `get_the_author_meta('url')` only WP User will be linked. With `get_the_author_meta('website')` only guest author will be linked.

PR #444, opened 2017-07-13
Fix coauthors_links_single()

rebeccahum:
As per #132, the template tag `coauthors_links()` was not working for guest authors.  This patch addresses the issue.

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -378,14 +378,26 @@ function coauthors_emails( $between = null, $betweenLast = null, $before = null,
  * @return string
  */
 function coauthors_links_single( $author ) {
-	if ( get_the_author_meta( 'url' ) ) {
-		return sprintf( '<a href="%s" title="%s" rel="external">%s</a>',
-			get_the_author_meta( 'url' ),
-			esc_attr( sprintf( __( 'Visit %s&#8217;s website' ), get_the_author() ) ),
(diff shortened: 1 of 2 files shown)


template-tags.php
```diff
@@ -378,13 +378,23 @@ function coauthors_emails( $between = null, $betweenLast = null, $before = null,
  * @return string
  */
 function coauthors_links_single( $author ) {
-	if ( get_the_author_meta( 'url' ) ) {
-		return sprintf( '<a href="%s" title="%s" rel="external">%s</a>',
+	if ( 'guest-author' === $author->type ) {
+		if ( get_the_author_meta( 'website' ) ) {
+			return sprintf( '<a href="%s" title="%s" rel="external" target="_blank">%s</a>',
+				get_the_author_meta( 'website' ),
```
philipjohn: This should be escaped with `esc_url()` here.

template-tags.php
```diff
@@ -378,13 +378,23 @@ function coauthors_emails( $between = null, $betweenLast = null, $before = null,
...
+				esc_attr( sprintf( __( 'Visit %s&#8217;s website' ), get_the_author() ) ),
+				get_the_author()
```
philipjohn: This should be escaped with `esc_html()` or `wp_kses_post()`

PR #446, opened 2017-08-11
Added guest author hooks for create/delete

rebeccahum:
Added action hooks for everytime guest author is created and deleted


sboisvert:
The hooks are useful in being able to manage caches that take into account guest authors.

Issue #447, opened 2017-08-14
CAP's filtering of count_user_posts has side-effects on REST API usage


jacklenox:
At the moment, the core REST API uses `count_user_posts` in its permissions checks to see if a request has access to read a user. See the line in question [here](https://core.trac.wordpress.org/browser/tags/4.8.1/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php#L379).

CAP's filtering of `count_user_posts`, [as seen here](https://github.com/Automattic/Co-Authors-Plus/blob/c066be2/co-authors-plus.php#L943-L958), has side-effects on the REST API's functionality when it changes this count from the API's core expectation.

We should seek to make CAP REST API aware so that it functions as expected when used in combination with the REST API.

rebeccahum:
Closing because we've removed `wpcom-helper.php`.

PR #448, opened 2017-08-14
Fix author response in REST API

jacklenox:
As per #447, CAP's filtering of `count_user_posts` is having unintended side-effects on the embedded author in REST API responses.

Here's a temporary helper fix to restore expected functionality on WordPress.com.

### DIFF
--- a/wpcom-helper.php
+++ b/wpcom-helper.php
@@ -1,0 +1,227 @@
+<?php
+/**
+ * Auto-apply Co-Authors Plus template tags on themes that are properly using the_author()
+ * and the_author_posts_link()
+ * Auto-apply Co-Authors Plus in oembed endpoint
+ */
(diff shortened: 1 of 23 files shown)


wpcom-helper.php
```diff
@@ -236,3 +236,17 @@ function wpcom_vip_cap_replace_author_link( $link, $author_id, $author_nicename
 		add_filter( 'author_link', 'wpcom_vip_cap_replace_author_link', 99, 3 );
 	}
 }, 9 );
+
+/**
+ * Do not modify user counts during REST API requests pending CAP support for the API
+ *
+ * See more background on this issue here: https://github.com/Automattic/Co-Authors-Plus/issues/447
+ */
+add_action( 'init', function() {
```
philipjohn: How come this needs to be wrapped in an `init` hook?

PR #450, opened 2017-08-31
Fixes logic for DOING_AUTOSAVE check

mslinnea:
The logic for checking whether Autosave is occurring is reversed.  It doesn't look like this bug is very serious, as it doesn't prevent data from being saved during a normal post save.  I think mainly this fix would help with performance.  Autosave occurs every few seconds on the edit screen, so on a site with a lot of users, this could add up.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -729,7 +729,7 @@ class CoAuthors_Plus {
 	function coauthors_set_post_author_field( $data, $postarr ) {
 
 		// Bail on autosave
-		if ( defined( 'DOING_AUTOSAVE' ) && ! DOING_AUTOSAVE ) {
+		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
 			return $data;
 		}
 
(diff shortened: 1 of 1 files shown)


philipjohn:
Thanks for this! Could you detail the bug this fixes please, including steps to reproduce?

mslinnea:
@philipjohn The bug is that when WordPress is doing autosave, this function should return.  As stated in the comment on line 724, it says "Bail on Autosave".  As the logic is written currently, it does not bail on autosave.  Instead, it runs through the whole function.

You can reproduce this by putting a breakpoint later on in that function, for example on line 756.  Then open a WordPress edit screen, and you should find that your debugger stops at that line, and then a few seconds later, it happens again... And every time autosave is called, this logic will run.

It looks like the coauthors-nonce isn't set during autosave, so most of the function logic doesn't run, and what ends up running is just line 756, so this isn't too bad of a bug at all, but still incorrect in the intended purpose of this code -- which is to not save the author until the user has chosen to save the post.

Issue #449, opened 2017-08-25
add_coauthors.function doesn't work, when user_login has spaces?


TylerDurdon:
Hi, really like the plugin, but users with space in user_login will not be added when using
$coauthors_plus->add_coauthors.
Was there a workaround or is there a fix?

Thx & Regards

TylerDurdon:
is there anyone supporting this plugin?

rebeccahum:
Hi @TylerDurdon, can you give me more specific steps to reproduce this?  I'm seeing that it's possible to add users with a space in their user_login to a post when I type their name in the box under "Authors".

rebeccahum:
Hi @TylerDurdon, can you give me more specific steps to reproduce this?  I'm not sure what instance of `add_coauthors` you are experiencing this issue with.

rebeccahum:
I'm seeing that when I delete a user and try to re-assign the posts to a user with spaces in their `user_login`, it doesn't propagate as expected.  

There are probably more instances of this happening with `add_coauthors`.

rebeccahum:
I'm also seeing when you login as a user with spaces in their `user_login`, when you hit Posts > Add New, you receive the following the error message: 

```
Notice: Trying to get property of non-object in /wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 841

Notice: Trying to get property of non-object in /wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 851
```

TylerDurdon:
HI rebeccahum,

when i add the user in a page/post > author it works perfect.
When i add the user with the function in php:

> add_coauthors( $this->author1_post1, array( $editor1->user_login ), true );

it doens't work, when the user has a space in his user_login name.

TylerDurdon:
Hi @rebeccahum, i tried user-login and user_nicename, but there is no differnce in behaviour.
I always get the same users in the authors list after saving the site (using hook edit_form_after_title)
Without a space in user_login OR without "-" in user_nicename.

$coauthors_plus->add_coauthors($pageid , array($user->user_login), true); 
and
$coauthors_plus->add_coauthors($pageid , array($user->user_nicename), true); 

i dump my array, so the users to be added are clear to me, and the four users have the data:
1. user_login: admin 
user_nicename: admin
state: Works

2. user_login: Test User
user_nicename: test-user
state: Don't work

3. user_login: tester2
user_nicename: tester2
state: Works

4. user_login: torsten tester
[...]

PR #453, opened 2017-09-13
user_login spaces problem when using add_coauthors

rebeccahum:
As per #449, `add_coauthors` should be passing in the `user_nicename` instead of `user_login` because it won't work if `user_login` has spaces.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,30 @@
 language: php
 
 php:
-    - "nightly"
+    - 7.1
 
 env:
     - WP_VERSION=latest
 
 matrix:
+matrix:
   include:
-# nightly+latest already included above as first build.
-#  - php: "nightly"
-#    env: WP_VERSION=latest
-   - php: "nightly"
-     env: WP_VERSION=4.6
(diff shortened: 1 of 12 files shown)


TylerDurdon:
Hi @rebeccahum, 
referencing issue #449 

I tried user-login and user_nicename, but there is no differnce in behaviour.
I always get the same users in the authors list after saving the site (using hook edit_form_after_title)
Without a space in user_login OR without "-" in user_nicename.

$coauthors_plus->add_coauthors($pageid , array($user->user_login), true);
and
$coauthors_plus->add_coauthors($pageid , array($user->user_nicename), true);

i dump my array, so the users to be added are clear to me, and the four users have the data:

    user_login: admin
    user_nicename: admin
    state: Works

    user_login: Test User
    user_nicename: test-user
    state: Don't work

    user_login: tester2
    user_nicename: tester2
    state: Works

    user_login: torsten tester
    user_nicename: torsten-tester
    state: Don't work

after saving i have the authors admin and tester2 in the author-area in edit mode. any further idea?

Issue #425, opened 2017-06-02
0 posts for Guest Author different than 0 posts for WP User


trepmal:
Related #201, #409

**Setup**

I have created 2 users:
 * `lazy-author`: A standard WP User (role: author) with 0 posts
 * `lazy-guest`: A Guest Author with 0 posts

Author archive for the first - http://local.wordpress.local/author/lazy-author/ - returns a 200, with "no posts found"

Author archive for the second - http://local.wordpress.local/author/lazy-guest/ - returns a 404, with an "oops" message

I'm not sure we should return a 404, as this doesn't match core behavior. I think we might just be dealing with cruft left over after copy-pasta rotted a bit.

rebeccahum:
+1 to this for consistency with core.  It would also cause less confusion for those building custom author templates with this plugin.

philipjohn:
Agreed, let's make this consistent with core.

PR #457, opened 2017-10-22
Remove redundant test for 404 on Author Archive

philipjohn:
Since we made CAP consistent with core in it's handling of author archives where there are no posts (see #425) we no longer expect there to be a 404 response on the author archive for a guest author.


philipjohn:
Fixes #398

SeanDS:
I guess the [temporary table workaround](https://github.com/Automattic/Co-Authors-Plus/blob/432d6b0a12d96bd97d5fcb740f8ad90a25241962/tests/test-author-queried-object.php#L13) is now no longer needed?

PR #471, opened 2017-12-22
Delete is_array() condition

RoyTheRoyalBoy:
Since our function already defines/declares all_posts as an array in line 18, we shouldn't need to check it it's an array since we're guaranteed that it'll be an array.


RoyTheRoyalBoy:
That's true, I was just looking for potential optimizations while I was applying back in late December for internship, so I didn't really want to commit any changes. But I can make the changes!

PR #473, opened 2017-12-25
Fix the issue where guest authors with non-ASCII characters can't be used as co-authors

sanketio:
## See #198 

- Used `urlencode()` to encode special characters for coauthor to compare names
- Used `urldecode()` for display purpose in autocomplete with special characters in `user_nicename`
- Replaced `unescape()` with `decodeURIComponent()` in JS because `unescape()` is deprecated.
- Added unit test case for the fix and also covered additional unit test cases.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -766,7 +766,10 @@ class CoAuthors_Plus {
 
 		// This action happens when a post is saved while editing a post
 		if ( isset( $_REQUEST['coauthors-nonce'] ) && isset( $_POST['coauthors'] ) && is_array( $_POST['coauthors'] ) ) {
-			$author = sanitize_text_field( $_POST['coauthors'][0] );
+
+			// urlencode() is for encoding coauthor name with special characters to compare names when getting coauthor.
+			$author = urlencode( sanitize_text_field( $_POST['coauthors'][0] ) );
+
 			if ( $author ) {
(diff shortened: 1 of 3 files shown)


co-authors-plus.php
```diff
@@ -766,7 +766,10 @@ function coauthors_set_post_author_field( $data, $postarr ) {
 
 		// This action happens when a post is saved while editing a post
 		if ( isset( $_REQUEST['coauthors-nonce'] ) && isset( $_POST['coauthors'] ) && is_array( $_POST['coauthors'] ) ) {
-			$author = sanitize_text_field( $_POST['coauthors'][0] );
+
+			// urlencode() is for encoding coauthor name with special characters to compare names when getting coauthor.
+			$author = urlencode( sanitize_text_field( $_POST['coauthors'][0] ) );
```
philipjohn: Only legacy systems should use `urlencode()`. Instead, `rawurlencode()` should be used whenever a URL needs to be properly encoded.

philipjohn:
Oh, nearly forgot to mention the failing tests - are those related?

co-authors-plus.php
sanketio: Replaced `urlencode/urldecode` with `rawurlencode/rawurldecode` in 9806b14f6e2169a50b43082e73aa9445f2ce1ea3

PR #474, opened 2017-12-25
Fix the issue where incompatibility when `coauthors_auto_apply_template_tags` set to true

sanketio:
## See #279 

- Removed `the_author` filter from `coauthors_posts_links()` & `coauthors_links()` because that filter is called in loop in these functions, so it is creating this issue.
- Added unit tests

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -239,12 +239,15 @@ function coauthors_posts_links( $between = null, $betweenLast = null, $before = 
 
 	global $coauthors_plus_template_filters;
 
-	/**
-	 * Removing "the_author" filter so that it won't get called in loop and append names for each author.
-	 *
-	 * Ref : https://github.com/Automattic/Co-Authors-Plus/issues/279
-	 */
-	remove_filter( 'the_author', array( $coauthors_plus_template_filters, 'filter_the_author' ) );
+	if ( ! empty( $coauthors_plus_template_filters ) && $coauthors_plus_template_filters instanceof CoAuthors_Template_Filters ) {
+
+		/**
+		 * Removing "the_author" filter so that it won't get called in loop and append names for each author.
+		 *
+		 * Ref : https://github.com/Automattic/Co-Authors-Plus/issues/279
+		 */
+		remove_filter( 'the_author', array( $coauthors_plus_template_filters, 'filter_the_author' ) );
+	}
 
 	$coauthors_posts_links = coauthors__echo( 'coauthors_posts_links_single', 'callback', array(
 		'between'     => $between,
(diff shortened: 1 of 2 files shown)


template-tags.php
```diff
@@ -236,12 +236,31 @@ function coauthors( $between = null, $betweenLast = null, $before = null, $after
...
+	$coauthors_posts_links = coauthors__echo( 'coauthors_posts_links_single', 'callback', array(
+		'between'     => $between,
 		'betweenLast' => $betweenLast,
-		'before' => $before,
-		'after' => $after,
+		'before'      => $before,
+		'after'       => $after,
 	), null, $echo );
+
+	if ( ! empty( $coauthors_plus_template_filters ) && $coauthors_plus_template_filters instanceof CoAuthors_Template_Filters ) {
```
philipjohn: Rather than duplicate the conditional here, would it be more efficient to set a variable with the result of the conditional and use that instead?

E.g. 
```
$modify_filter = ! empty( $coauthors_plus_template_filters ) && $coauthors_plus_template_filters instanceof CoAuthors_Template_Filters;
...
if ( $modify_filter ) {
...
```

sanketio: Done in c8539e637f511590f4e178ab77a6c6676cfdc073

philipjohn:
@sanketio In testing this I'm seeing two things; 1) the original issue appears regardless of the use of the `coauthors_auto_apply_template_tags` filter and 2) the fix here doesn't alter the output.

The steps I took were:

1. Create two Guest Authors - "Guest Author A and Guest Author B"
2. Modify the Twenty Seventeen theme function `twentyseventeen_posted_on()` within `inc/template-tags.php` to set `$byline` to the output of `coauthors()`. The byline on a single post now reads "GUEST AUTHOR A AND GUEST AUTHOR BPOSTED ON MARCH 2, 2018 GUEST AUTHOR A AND GUEST AUTHOR B"
3. Add `add_filter( 'coauthors_auto_apply_template_tags', '__return_true' );` to the theme functions.php
4. Reload the page and the byline remains the same.
5. Apply the fix in this PR, and reload - the output remains the same.

Can you let me know what steps you used to verify the issue in #279 and what steps you took to test that this change fixes that issue?

Thanks!

sanketio:
@philipjohn 

I think you have directly used `coauthors()` as it is, so it will directly echo authors names. You'll need to use `coauthors( null, null, null, null, false )`, this will return authors names and store into `$byline` variable.

Let me know if you have any doubts.

philipjohn:
@sanketio could you take another look at this please? The `test_coauthors_posts_links_single` test is failing.

sanketio:
@philipjohn : I've fixed #499 and sent PR.

Issue #458, opened 2017-10-22
Tests failing with PHP 7


philipjohn:
> PHP Fatal error: Class 'PHPUnit_Framework_TestCase' not found in /tmp/wordpress-tests-lib/includes/testcase.php on line 15

https://travis-ci.org/Automattic/Co-Authors-Plus/jobs/291114352

Tests are passing in 5.6 but it looks like something to do with our setup means that the testing framework isn't properly available in PHP 7. It's probably an old code problem!

mjangda:
This is likely related: https://core.trac.wordpress.org/changeset/40547

philipjohn:
Fixed by #482

jeherve:
Recent PRs appear to have the same problem. Would it make sense to move to WP 4.7 instead of WP 4.6 in the tests? I believe that may solve the problem, as per this ticket: 
https://core.trac.wordpress.org/ticket/39822

WP 4.7 was released in December 2016, so maybe it is fair to test for that version now, 2 major releases later?

TheCrowned:
I am running tests with WP latest (at this time, 4.9.6) and PHP 7.2.5, and everything runs smoothly. Is this still an issue?

nielslange:
When running the unit tests using WP 5.3 and WP 5.7 with PHP 7.4, I see the following warnings and errors:

```
// Running PHPUnit using WordPress 5.3
$ phpunit
Installing...
Installing network...
Running as multisite...
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 13:
  - Element 'testsuite': The attribute 'name' is required but missing.

  Test results may not be as expected.


................................................................. 65 / 98 ( 66%)
............................FF...                                 98 / 98 (100%)

Time: 8.94 seconds, Memory: 32.00 MB

There were 2 failures:

1) Test_Template_Tags::test_coauthors_get_avatar_default
Failed asserting that 1 matches expected 0.

/Users/nielslange/wpvip/plugins/Co-Authors-Plus/tests/test-template-tags.php:991

2) Test_Template_Tags::test_coauthors_get_avatar_when_guest_author
Failed asserting that 1 matches expected 0.

[...]

PR #478, opened 2018-01-02
Update .travis.yml file for PHPUnit tests

sanketio:
## See #458 

Somehow PHPUnit is loading from `/home/travis/.phpenv/shims/phpunit` path instead of composer, so it was causing the issue.

I have updated below points:
- Added `dist: precise` for PHP v5.2
- Changed PHPCS path from `$PHPCS_DIR/scripts/phpcs` to `$PHPCS_DIR/bin/phpcs`, because in https://github.com/squizlabs/PHP_CodeSniffer, `phpcs` is moved to `bin` directory.
- Added condition check to use PHPUnit from composer first.


philipjohn:
Rather than setting the distro like this, let's copy the setup that's working for [Liveblog](https://github.com/Automattic/liveblog/blob/master/.travis.yml)

sanketio:
@philipjohn: Just wanted to ask if we want the same travis setup or just required settings from [Liveblog](https://github.com/Automattic/liveblog/blob/master/.travis.yml)?

philipjohn:
Just the relevant bits please, @sanketio - there's definitely some unique stuff in the Liveblog config that doesn't make sense for CAP.

PR #484, opened 2018-02-02
Changes to resolve issue #332 about missing coauthor meta 

nilzari:
I have made changes in the get_the_coauthor_meta function to resolve this issue #332 . The original function was not fetching meta data of the guest users. It was fetching data of only WP users from wp_users table. I have modified the function to fetch meta data of both WP users as well guest users associated with the post. 

Some of the meta fields associated with WP users are not available for guest users. So the function will return value for only those meta fields that are common for both type of users. 

Code 
```
while ( have_posts() ) : the_post();
     var_dump( get_the_coauthor_meta( 'email' ) ); 
endwhile;
```
Output for above code: 
 
For mapped guest user: Jeni
mapped_guest_user

For guest user without any mapping : George
not mapped guest user

For WP user: admin
new_user

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -422,43 +422,46 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
  *
  * @param string $field Required The user field to retrieve.[login, email, nicename, display_name, url, type]
  * @param string $user_id Optional The user ID for meta
+ *
+ * @return array $meta Value of the user field
  */
-function get_the_coauthor_meta( $field, $user_id = false ) { 
+function get_the_coauthor_meta( $field, $user_id = false ) {
 	global $coauthors_plus;
 
 	if ( ! $user_id ) {
 		$coauthors = get_coauthors();
-        } else {
-                $coauthor_data = $coauthors_plus->get_coauthor_by( 'id', $user_id );
-                $coauthors = array();
-                if ( ! empty( $coauthor_data ) ) {
-			$coauthors[] = $coauthor_data;
-		}
+    } else {
+        $coauthor_data = $coauthors_plus->get_coauthor_by( 'id', $user_id );
+        $coauthors = array();
+        if ( ! empty( $coauthor_data ) ) {
+            $coauthors[] = $coauthor_data;
         }
+    }
 
 	$meta = array();
-        
-        if ( in_array( $field, array( 'login', 'pass', 'nicename', 'email', 'url', 'registered', 'activation_key', 'status' ) ) )
-                $field = 'user_' . $field;
-        
-	foreach ( $coauthors as $coauthor ) {                
-		$user_id = $coauthor->ID;      
-                
-                if ( isset( $coauthor->type )  && 'user_url' === $field ) {
-                        if ( 'guest-author' === $coauthor->type) {
-                                $field = 'website';
-                        } 
-                } else if ( 'website' === $field ) {
-                        $field = 'user_url';
-                }
-                
-                if ( isset( $coauthor->$field ) ) {
-                        $meta[ $user_id ] = $coauthor->$field;
-                } else {
-                        $meta[ $user_id ] = '';
-                }                                
-	}        
-        
+
+    if ( in_array( $field, array( 'login', 'pass', 'nicename', 'email', 'url', 'registered', 'activation_key', 'status' ) ) ) {
+        $field = 'user_' . $field;
+    }
+
+	foreach ( $coauthors as $coauthor ) {
+		$user_id = $coauthor->ID;
+
+        if ( isset( $coauthor->type )  && 'user_url' === $field ) {
+            if ( 'guest-author' === $coauthor->type) {
+                $field = 'website';
+            }
+        } else if ( 'website' === $field ) {
+            $field = 'user_url';
+        }
+
+        if ( isset( $coauthor->$field ) ) {
+            $meta[ $user_id ] = $coauthor->$field;
+        } else {
+            $meta[ $user_id ] = '';
+        }
+	}
+
 	return $meta;
 }
 
@@ -466,9 +469,9 @@ function get_the_coauthor_meta( $field, $user_id = false ) {
 function the_coauthor_meta( $field, $user_id = 0 ) {
 	// TODO: need before after options
 	$coauthor_meta = get_the_coauthor_meta( $field, $user_id );
-        foreach ( $coauthor_meta as $meta ) {     
-                echo $meta; 
-        }
+    foreach ( $coauthor_meta as $meta ) {
+        echo $meta;
+    }
 }
 
 /**
--- a/template-tags.php
+++ b/template-tags.php
@@ -428,30 +428,26 @@ function get_the_coauthor_meta( $field ) {
 	$coauthors = get_coauthors();
 	$meta = array();
         
-	foreach ( $coauthors as $coauthor ) {
-                
-		$user_id = $coauthor->ID;
-                if ( isset($coauthor->data) ){
-                     $coauthor = $coauthor->data;
-                }
-                
-                if( $field == 'url' && isset ($coauthor->website) )
-                    $field = 'website';
-                
-                if( $field == 'website' && isset ($coauthor->user_url) )
-                    $field = 'user_url';
-                
-                if ( in_array( $field, array( 'login', 'pass', 'nicename', 'email', 'url', 'registered', 'activation_key', 'status' ) ) )
-                    $field = 'user_' . $field;
-
-                if ( isset( $coauthor->$field ) ){
-                    $meta[ $user_id ] = $coauthor->$field;
-                } else {
-                    $meta[ $user_id ] = '';
-                }
-                                
-	}
+        if ( in_array( $field, array( 'login', 'pass', 'nicename', 'email', 'url', 'registered', 'activation_key', 'status' ) ) )
+                $field = 'user_' . $field;
         
+	foreach ( $coauthors as $coauthor ) {                
+		$user_id = $coauthor->ID;
+                
+                if ( isset( $coauthor->type )  && 'user_url' === $field ) {
+                        if ( 'guest-author' === $coauthor->type) {
+                                $field = 'website';
+                        } 
+                } else if ( 'website' === $field ) {
+                        $field = 'user_url';
+                }
+                
+                if ( isset( $coauthor->$field ) ) {
+                        $meta[ $user_id ] = $coauthor->$field;
+                } else {
+                        $meta[ $user_id ] = '';
+                }                                
+	}        
 	return $meta;
 }
 
--- a/template-tags.php
+++ b/template-tags.php
@@ -426,11 +426,12 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
  * @return array $meta Value of the user field
  */
 function get_the_coauthor_meta( $field, $user_id = false ) {
-	global $coauthors_plus;
+    global $coauthors_plus;
 
-	if ( ! $user_id ) {
-		$coauthors = get_coauthors();
-    } else {
+    if ( ! $user_id ) {
+        $coauthors = get_coauthors();
+    }
+    else {
         $coauthor_data = $coauthors_plus->get_coauthor_by( 'id', $user_id );
         $coauthors = array();
         if ( ! empty( $coauthor_data ) ) {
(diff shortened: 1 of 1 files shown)


sboisvert:
@rebeccahum Great code review :)
@nilzari great code!

I had a few questions while looking at this. For example

>Some of the meta fields associated with WP users are not available for guest users. So the function will return value for only those meta fields that are common for both type of users.

Why did you make that decision? What are the advantages and disadvantages of it?

Reading only this PR I was confused about the code example given. I looked at it and I thought to myself "Well this looks like it works" which is what it is, but I guess my mind first went to that's the code example of what was broken. I think it could of been useful to duplicate the report and example from @trepmal in #332 to do a sort of "before and after" to show what didn't work and what works now :).

I agree with @rebeccahum's feedback re:
>Instead of setting all the fields dependent on the user type, it would be nice to set and return just one respective field once for both types at the same time (since we only pass in one field). To start, I would look into the similarities between the guest author object and the WP User.

nilzari:
I have updated the function get_the_coauthor_meta(). Now the function will return value for any meta field that is associated with either wp_user or guest_author. Function will return blank value if the field is not associated with the user.

**Screenshot before solving the issue:** Function was returning value of only wp_user. It was not fetching value for guest_author.
capture_before

**Screenshot after resolving issue:** Function returns value of meta field for both wp_user and guest_author.
capture_after

I have also tried to resolve the error generated for 'type' field in the previous commit as described by @rebeccahum

template-tags.php
```diff
@@ -417,19 +417,45 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+                } else {
+                    $meta[ $user_id ] = '';
+                }
+                                
 	}
+        
 	return $meta;
 }
 
+
```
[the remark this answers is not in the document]
sboisvert: It's good feedback and I like the way you phrased it :)

template-tags.php
```diff
@@ -417,19 +417,45 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+        
 	foreach ( $coauthors as $coauthor ) {
+                
 		$user_id = $coauthor->ID;
-		$meta[ $user_id ] = get_the_author_meta( $field, $user_id );
+                if ( isset($coauthor->data) ){
+                     $coauthor = $coauthor->data;
+                }
+                
+                if( $field == 'url' && isset ($coauthor->website) )
```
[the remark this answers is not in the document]
sboisvert: >So, to kill two birds with one stone, we can do a check to see if the type is set and if it is, do another check to make sure it's the guest author type. Once all those checks pass, we can set the field to website from an input of the field user_url. 

I'm not sure I follow. Perhaps an example would help? (Also, it's possible this is clear if you're more familiar with the code).

One suggestion on the above is I'd create a new paragraph for the yoda and strict comparison to denote it's separate from the first item.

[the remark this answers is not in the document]
rebeccahum: Something like starting the check off with `if ( isset( $coauthor->type ) && 'user_url' === $field )`, then in that if statement for failsafe, throw in another `if ( 'guest-author' === $coauthor->type )` check.  Once all checks pass, we can assign `$field = 'website'` since that's the cap property equivalent of `user_url`.

nilzari:
I have updated the code of get_the_coauthor_meta() function as per the comments provided by @rebeccahum. 

Thank you for the review @rebeccahum  and @sboisvert .

template-tags.php
```diff
@@ -417,19 +417,41 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+	foreach ( $coauthors as $coauthor ) {                
 		$user_id = $coauthor->ID;
-		$meta[ $user_id ] = get_the_author_meta( $field, $user_id );
-	}
+                
+                if ( isset( $coauthor->type )  && 'user_url' === $field ) {
+                        if ( 'guest-author' === $coauthor->type) {
+                                $field = 'website';
+                        } 
+                } else if ( 'website' === $field ) {
```
rebeccahum: Hi! Are you able to explain the reasoning behind this block?  If we called `get_the_coauthor_meta( 'website' )`,  guest authors would yield nothing seeing as their property is called `'website'` and we would be looking for `'user_url'`.  We should only set `'website'` to `'user_url'` if it is not a guest-author type or if the type is not set.

template-tags.php
```diff
@@ -417,19 +417,41 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
 	), null, $echo );
 }
 
-function get_the_coauthor_meta( $field ) {
+/**
+ * Outputs the co-authors Meta Data
+ *
+ * @param string $field Required The user field to retrieve.[login, email, nicename, display_name, url, type]
+ */
+function get_the_coauthor_meta( $field ) {
```
rebeccahum: Since `the_coauthor_meta` passes in a possible second parameter to `get_the_coauthor_meta` called `$user_id`, perhaps we should take that into account.

philipjohn:
Hi @nilzari we're currently testing for v3.3 and it'd be great to get this in if you're able to address @rebeccahum's feedback?

template-tags.php
nilzari: @rebeccahum : Are we allowing to call get_the_coauthor_meta( 'website' )? As per my knowledge, the function originally allowed only 'url' as the field parameter for both, guest_authors as well as wp_users. So, I have updated the function accordingly.

rebeccahum: @nilzari Based on your line of code, it is doing a check to see if it's passed in...so, based on that -- the only reason we would ever need to perform a check for `$field` equating to `'website'` is if we did a call to  `get_the_coauthor_meta( 'website' )`.
However, with that said, I don't think we need `'website'` to be passed in because I'm not seeing it in the list here for `get_the_author_meta` here: https://developer.wordpress.org/reference/functions/get_the_author_meta/ I'd presume we'd want `get_the_coauthor_meta` to mirror the core function as closely as possible.

nilzari: @rebeccahum : Below is the reason for condition ``` } else if ( 'website' === $field ) { ``` .

If the  ```$coauthors``` array has two objects, first one for guest_author and second one for wp_user (since the post has two authors). Now we make a call to ```get_the_coauthor_meta( 'url' )```. In this case, since the first object is guest_author, the ```$field``` is updated to 'website'. For the second object which is wp_user, the function will check if the ```$field``` is 'website', it will again change it to 'user_url'.

rebeccahum: @nilzari Ah okay, I see what you are trying to do now.  I don't think it will work as expected because in this case it's _actually_ checking to see if `'website'` was passed into `$field` once `isset( $coauthor->type )  && 'user_url' === $field` evaluates to `false`.  These lines you have here will be enough to do what you want:
```
if ( isset( $coauthor->type )  && 'user_url' === $field ) {
                        if ( 'guest-author' === $coauthor->type) {
                                $field = 'website';
                        } 
```
Because in the scenario you are describing, in the first one for guest_author, it will set the `$field` value to `'website'` and grab that property and then for a wp_user, it doesn't need to set anything since the `$field` value already at `user_url`.

nilzari: @rebeccahum : In the first iteration, for first object 'guest_author', the  ```$field``` values is set to 'website' by below lines of code .

```
if ( isset( $coauthor->type )  && 'user_url' === $field ) {
                        if ( 'guest-author' === $coauthor->type) {
                                $field = 'website';
                        } 
```

When the code will check for second object in the second iteration, it will find ```$field``` value as 'website' which is not applicable for wp_user. So I have to again set the value for ```$field``` as 'user_url'. I have to explicitly change the ```$field``` value for wp_user in this case since it is updated in first iteration and doesn't set back to 'user_url' unless i change it explicitly. Therefore, below condition is necessary.

```
else if ( 'website' === $field ) {
                        $field = 'user_url';
                }
```

rebeccahum: @nilzari Ahhh now I see the edge case that you are trying to account for.  I understand now, good work.  Carry on!

nilzari:
@rebeccahum : I have updated functions ```the_coauthor_meta()``` and ```get_the_coauthor_meta()``` to allow fetching data based on user_id. Could you please review it?

template-tags.php
```diff
@@ -417,22 +417,58 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
-function get_the_coauthor_meta( $field ) {
-	global $wp_query, $post;
+/**
+ * Outputs the co-authors Meta Data
+ *
+ * @param string $field Required The user field to retrieve.[login, email, nicename, display_name, url, type]
+ * @param string $user_id Optional The user ID for meta
+ */
+function get_the_coauthor_meta( $field, $user_id = false ) { 
+	global $wp_query, $post, $coauthors_plus;
```
rebeccahum: I _think_ we can safely remove the declaration of global vars `$wp_query` and `$post`, as I don't see it being used anywhere in this function.

nilzari: @rebeccahum : I have removed the unused global variables from the function as suggested by you.

rebeccahum:
@nilzari Looks good to me now!  Thank you for being receptive to feedback and for your hard work on this!

@sboisvert Do you want to do a glance-over on this as well?

template-tags.php
```diff
@@ -417,22 +417,61 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+        $coauthors = array();
+        if ( ! empty( $coauthor_data ) ) {
+            $coauthors[] = $coauthor_data;
+        }
+    }
 
-	$coauthors = get_coauthors();
 	$meta = array();
 
+    if ( in_array( $field, array( 'login', 'pass', 'nicename', 'email', 'url', 'registered', 'activation_key', 'status' ) ) ) {
```
philipjohn: Spacing is a little off here too. It'd also be good to split this into two or more lines.

template-tags.php
```diff
@@ -417,22 +417,61 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+ * @param string $user_id Optional The user ID for meta
+ *
+ * @return array $meta Value of the user field
+ */
+function get_the_coauthor_meta( $field, $user_id = false ) {
+	global $coauthors_plus;
+
+	if ( ! $user_id ) {
+		$coauthors = get_coauthors();
+    } else {
```
philipjohn: Spacing is a little off here - would be good to clean that up.

template-tags.php
```diff
@@ -417,22 +417,61 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
 
+    if ( in_array( $field, array( 'login', 'pass', 'nicename', 'email', 'url', 'registered', 'activation_key', 'status' ) ) ) {
+        $field = 'user_' . $field;
+    }
+
 	foreach ( $coauthors as $coauthor ) {
 		$user_id = $coauthor->ID;
-		$meta[ $user_id ] = get_the_author_meta( $field, $user_id );
+
+        if ( isset( $coauthor->type )  && 'user_url' === $field ) {
```
philipjohn: Spacing looks a little off from here to line 462 as well. Perhaps spaces vs tabs?

template-tags.php
```diff
@@ -417,22 +417,61 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+
 	return $meta;
 }
 
+
 function the_coauthor_meta( $field, $user_id = 0 ) {
 	// TODO: need before after options
-	echo get_the_coauthor_meta( $field, $user_id );
+	$coauthor_meta = get_the_coauthor_meta( $field, $user_id );
+    foreach ( $coauthor_meta as $meta ) {
```
philipjohn: Spacing again :)

template-tags.php
```diff
@@ -417,22 +417,61 @@ function coauthors_ids( $between = null, $betweenLast = null, $before = null, $a
...
+        echo $meta;
```
philipjohn: All output should be secured by appropriate escaping [as late as possible](https://vip.wordpress.com/documentation/best-practices/security/validating-sanitizing-escaping/#always-escape-late).

template-tags.php
nilzari: @philipjohn : I have made space alignment changes here. Could you please check and let me know if everything is ok? Also, could you please explain what do you mean by ```It'd also be good to split this into two or more lines.```?

template-tags.php
nilzari: @philipjohn : I have made space alignment changes here. Could you please check and let me know if everything is ok?

template-tags.php
nilzari: @philipjohn : I have made space alignment changes here. Could you please check and let me know if everything is ok?

template-tags.php
nilzari: @philipjohn : I have made space alignment changes here. Could you please check and let me know if everything is ok?

template-tags.php
nilzari: @philipjohn : I have added ```esc_html``` to output. Could you please check and let me know if everything is ok?

PR #487, opened 2018-02-08
Verify a user is reassigning posts during user delete

lipemat:
When deleting a user we are given the option to reassign or delete posts. No matter which option is chosen (radio) a user will always be selected in the 'reassign_user' drop-down.

This currently ignores the radio selections and always reassigns the posts to whatever user is either selected or the first on in the list.

This pull request adds a check for this radio value before blindly reassigning.


GaryJones:
As this PR is conflicted, I will close it out. Feel free to open a new targeted PR if the change is still desired.

GaryJones:
Actually, looks like the [last commit](https://github.com/Automattic/Co-Authors-Plus/pull/487/commits/e01b5301443f38829a324fbb9f436086ee7e497c) is small. Can't reopen as the original repository has been deleted.

PR #490, opened 2018-03-15
Skip irrelevant caps in filter_user_has_cap()

### LINKED ISSUE #489: Skip cap (capabilities) filtering on irrelevant cap checks

```
           'delete_user',
            'edit_users',
            'edit_user',
```

nickdaugherty:
Return early on caps that are not relevant to Co Authors Plus.

Fixes #489


nickdaugherty:
Probably more caps we should add to the list.

joshbetz:
Good catch. Could we only allow certain caps instead of allow all except this list?

rebeccahum:
Closed, as #543 has been merged now.

PR #493, opened 2018-03-19
Add correct post_type value to URL for "Mine" link on Pages screen

pdemarte:
Currently on the Pages screen there is an "invalid post type" error message because the post_type is being sent without a value.


rebeccahum:
Hey @pdemarte, are you able to elaborate or tell me how to reproduce?

Thanks!

pdemarte:
Hi, 
Sure, I should have included this in the first place!

The current code doesn't allow a user to view a list of only the Pages they have co-authored. It throws an error message instead.

### Steps to reproduce the issue:
1. Use the current master branch of Co-Authors-Plus on Wordpress 4.9.4 
2. Create a Page or use an existing Page
3. Assign a second author to the Page using the Co-Authors-Plus plugin (in this example, the second author I added is 'peted'.
4. Save the Page
5. Log in as one of the two authors of the Page.
6. Navigate to the All Pages screen `wp-admin/edit.php?post_type=page`
7. Click the "Mine" link
8. Instead of listing all of the current user's Pages, an error screen appears reading, "Invalid post type." 
co-authors-invalid-post-type Note the URL `wp-admin/edit.php?author_name=peted&post_type` in the screenshot.

    * The get_post_type() function in line 1301 results in a query string of just `?post_type`, instead of `post_type=page`, causing this error. If viewing a single Page, get_post_type() would return `page`, but on this screen it returns either `null` or `false`, so no value is appended to the `post_type` parameter.
8. It looks like a [similar fix](https://github.com/Automattic/Co-Authors-Plus/commit/f226e9c08d38e19e4de307956199a6a388abe842#diff-cf3698a69dea8725064e7ec6582e7a1eR297) was introduced in master since v3.2.2 was released, in the function `is_post_type_enabled()`.

When the pull request is applied, the url looks like  `wp-admin/edit.php?author_name=peted&post_type=page` and you should be able to see your own Pages after clicking "Mine":

co-authors-fixed-post-type

rebeccahum:
@pdemarte Thanks for the PR and providing the steps to reproduce!  Before we can merge this, do you mind resolving the conflicts?  Also, I noticed that when I see the "Mine" link, it should be "Mine (x number of posts" to mimic the posts behavior with a count:

i.e. on Posts:

image

then on Pages:

image

rebeccahum:
@pdemarte Thanks for pushing the merge conflicts.  Yes, the issue for the counts is now in #536.

pdemarte:
@rebeccahum I removed the count-related code from this branch, so it should be ready for merging.

rebeccahum:
@pdemarte Thanks so much for resolving the merge conflicts and the fixes.  However, it looks like the diff is off now since master wasn't merged directly into the PR.  Do you mind creating a new clean PR so that we can go ahead with the merge?  Thanks again!

pdemarte:
@rebeccahum Yes, and sorry about that.

Issue #384, opened 2016-11-16
Consider bumping min PHP requirement to 5.4?


mjangda:
Continuing discussion from https://github.com/Automattic/Co-Authors-Plus/pull/381#issuecomment-261065591

@sboisvert says: "we agreed to stop supporting 5.2 for any shared plugin and make the minimum PHP version supported 5.4."

If we move forward with this, we need to:
- [ ] Add some checks to prevent fatals when activating on < 5.4.
- [ ] Update the readme.
- [ ] Update the travis config to only test on PHP5.4+.

philipjohn:
I'm going to put this into v3.4!

What I'd like to suggest though is that we actually bump the minimum version to 5.6. What do you think?

nielslange:
Closing this issue as the corresponding rule set had been updated accordingly in #751.

PR #497, opened 2018-03-30
Gracefully require PHP 5.6.0

pyronaur:
Fills #384

I moved the whole Co-Authors-Plus plugin code from `co-authors-plus.php` to `co-authors-plus-core.php` to ensure that the plugin gracefully deactivates when the PHP Version requirements aren't met. This way PHP 5.6 code can be used throughout the plugin, except `co-authors-plus.php` which has to maintain PHP 5.2.4 compatibility.

If the current user visiting the site doesn't have permissions to deactivate plugins - no message will be shown, but the plugin won't load.

As soon as a user with `deactivate_plugins` privileges visits the dashboard the plugin will auto-deactivate and display a message that Co-Authors-Plus requires PHP 5.6.

I've also updated the readme `.md` and `.txt` files. This will also display the PHP version requirement on the WordPress.org repo due to the `Require PHP` tag [that's now available](https://make.wordpress.org/plugins/2017/08/29/minimum-php-version-requirement/)

**TL;DR**

☑️  Graceful degradation
☑️  Updated Readme for users
☑️  Updated Readme for WordPress.org "Requires PHP"
☑️  Travis Config is already using PHP 5.6+


class-cap-requirements.php
```diff
@@ -0,0 +1,85 @@
...
+		public function auto_deactivate() {
+
+			if ( function_exists( 'deactivate_plugins' ) && current_user_can( 'deactivate_plugins' ) ) {
+				deactivate_plugins( $this->plugin_name );
+			}
+
+		}
+
+	}
+}
```
GaryJones: Clear linespace needed at the end of the file here.

co-authors-plus.php
```diff
@@ -32,1752 +32,17 @@
...
- * @return array Array of coauthor WP_Term objects
- */
-function cap_get_coauthor_terms_for_post( $post_id ) {
-	global $coauthors_plus;
-	return $coauthors_plus->get_coauthor_terms_for_post( $post_id );
-}
+// Load the plugin if PHP Requirements are met
+if ( $cap_requirements->enviroment_meets_requirements() ) {
+	require_once dirname( __FILE__ ) . '/co-authors-plus-core.php';
+}
```
GaryJones: Clear line space needed here.

nielslange:
Closing this issue as the corresponding rule set had been updated accordingly in #751.

PR #504, opened 2018-04-15
Unexpected output of template tag 'coauthors_links' is resolved by updating conditions

### LINKED ISSUE #469: Function return author login name instead of display name
Hello,

The function `coauthors_links` return the login-name (slug) instead of the display name. By looking at the source, I noted that the callback `coauthors_links_single` will return nothing when no website is set for a guest author. The two conditionals seems to be the problem.

https://github.com/Automattic/Co-Authors-Plus/blob/b2bc1d3000fdb9b2de83601ccd26d14c9e023bf8/template-tags.php#L381

The conditional should check at the same time if the author type is guest-author _and_ if the author meta website is set.

```
if ( 'guest-author' === $author->type && get_the_author_meta( 'website' )) {
		return sprintf( '<a href="%s" title="%s" rel="external" target="_blank">%s</a>',
			esc_url( get_the_author_meta( 'website' ) ),
			esc_attr( sprintf( __( 'Visit %s&#8217;s website' ), esc_html( get_the_author() ) ) ),
			esc_html( get_the_author() )
		);
}
```

If no website is set, the function returns nothing and coauthors__echo returns the login-name.

https://github.com/Automattic/Co-Authors-Plus/blob/master/template-tags.php#L187

I'd be happy to submit a PR.

binodkalathil:
Fixes #469 
The template tag was outputting usernames of guest authors when they do not have a website set in their profile. It will now output guest authors display names as expected.
Also made corrections to PHPUnit test cases to work with the updated code.

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -420,14 +420,12 @@ function coauthors_emails( $between = null, $betweenLast = null, $before = null,
  * @return string
  */
 function coauthors_links_single( $author ) {
-	if ( 'guest-author' === $author->type ) {
-		if ( get_the_author_meta( 'website' ) ) {
-			return sprintf( '<a href="%s" title="%s" rel="external" target="_blank">%s</a>',
-				esc_url( get_the_author_meta( 'website' ) ),
-				esc_attr( sprintf( __( 'Visit %s&#8217;s website' ), esc_html( get_the_author() ) ) ),
-				esc_html( get_the_author() )
-			);
-		} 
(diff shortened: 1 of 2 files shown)


binodkalathil:
Closing this PR as travis.yml file is changed and I need to force a rebuild. I will submit another PR.

binodkalathil:
Adding before and after pics.

Before:
image

After:
image

sboisvert:
Looks good!

PR #507, opened 2018-04-19
Prevented duplication of Guest Authors when only their creation date is different

### LINKED ISSUE #326: Imported guest-author posts differing only in post_date should not be duplicated
During imports containing `guest-author` posts, (especially staged or chunked imports) these posts can sometimes fail the `get_post_exists` check for a trivial reason:  they differ only by `post_date`.  To replicate, import a `guest-author` ... then change the post date, produce new wxr and import the same author again.  A duplicate will appear.  These duplicates are difficult to remove, since their slugs are the same.

It's a not--entirely-uncommon use case (I've found) to import CAP posts that differ only in `post_date`.

CAP posts that differ only in post date should be considered to be duplicates, perhaps by filtering the results of the check using `wp_import_existing_post`.  

cc @ryanmarkel

binodkalathil:
Fixes #326 

Added a function to attach to the filter in Wordpress Importer plugin to modify the duplicate post detection process by avoiding the use of `post_date`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .DS_Store
 .svn
-wpcom-helper.php
 .idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
     include:
         # aliased to a recent 5.6.x version
         - php: "5.6"
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '5.6'
           env:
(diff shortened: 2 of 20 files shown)


GaryJones:
Only the [last commit](https://github.com/Automattic/Co-Authors-Plus/pull/507/commits/f2502769cef5006674b53e25b19bc2b628bedd78) is relevant here.

GaryJones:
Closing for age, conflicts and lack of update to narrow the scope to the right post type.

Issue #515, opened 2018-05-23
v3.3.0 Breaks translation of author name and 'Author: %s' string translation for get_the_archive_title() function


jxn:
It looks like this was introduced here: https://github.com/Automattic/Co-Authors-Plus/commit/8f8a0e7ca47a82e28351aa4c29556cd00177b254

After that commit, author name no longer translates.  For instance, in author.php, 
```
<?php
                    $archive_title = get_the_archive_title();
?>
```

This now returns `Author: slug-of-author-not-translated-name` in all languages.
Neither the author name (only the slug) nor the word 'Author: ' is translated.
Both should translate.

Backing out this commit fixes the issue.

rebeccahum:
PR #516 fixes this. Resolved now.

PR #516, opened 2018-05-25
Added i18n support and corrected author display name for author archive: addressing issue #515

blunce24:
I reproduced issue #515 locally by creating a few posts with the guest author Test Guest Author assigned to them. Then, I changed the language to Spanish and navigated to /?author_name=test-guest-author. Before the fix, the top of the page read: "Author: Test-Guest-Author" and after the fix, it reads "Autor: Test Guest Author"

The fix addresses both the translation issue and the author display name issue on the author archive page.


rebeccahum:
Nice job!

Issue #511, opened 2018-05-03
if current user doesn't have a a term, it returns PHP warnings on author archive page


rebeccahum:
When current user without a term gets passed in to `get_author_term` in `posts_where_filter`, it returns `false`:

```
$current_coauthor      = $this->get_coauthor_by( 'user_nicename', wp_get_current_user()->user_nicename );
					$current_coauthor_term = $this->get_author_term( $current_coauthor );
```

And thus throws PHP warnings later:

```
Notice: Trying to get property of non-object in /public_html/wp-content/plugins/Co-Authors-Plus-master/co-authors-plus.php on line 728

Notice: Trying to get property of non-object in /public_html/wp-content/plugins/Co-Authors-Plus-master/co-authors-plus.php on line 729
```

it stems from the lines:

```
					$current_user_query  = $wpdb->term_taxonomy . '.taxonomy = \''. $this->coauthor_taxonomy.'\' AND '. $wpdb->term_taxonomy .'.term_id = \''. $current_coauthor_term->term_id .'\'';
					$this->having_terms .= ' ' . $wpdb->term_taxonomy .'.term_id = \''. $current_coauthor_term->term_id .'\' OR ';
```

To reproduce, go to an author archive page as a user with no posts.

PR #517, opened 2018-05-26
check for coauthor terms before using them in query: addressing issue #511

blunce24:
This is addressing issue #511 to check if a user has terms before building a query with the terms.

To test, I went to an author archive page as a user (Subscriber) that didn't have any posts. I saw the following warnings:
Notice: Trying to get property of non-object in /srv/www/example/public_html/wp-content/plugins/co-authors-plus/co-authors-plus.php on line 728

Notice: Trying to get property of non-object in /srv/www/example/public_html/wp-content/plugins/co-authors-plus/co-authors-plus.php on line 729

With this fix, I don't see the warnings. I'm not certain if this is the right fix, or if the change in the where filter is needed for users that don't have terms. The fix looks to work with all my examples, but I may have missed an example to test.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -722,15 +722,15 @@ class CoAuthors_Plus {
 					is_author() &&
 					get_queried_object_id() != get_current_user_id()
 				) {
-				    $current_coauthor      = $this->get_coauthor_by( 'user_nicename', wp_get_current_user()->user_nicename );
-				    $current_coauthor_term = $this->get_author_term( $current_coauthor );
+					$current_coauthor      = $this->get_coauthor_by( 'user_nicename', wp_get_current_user()->user_nicename );
+					$current_coauthor_term = $this->get_author_term( $current_coauthor );
 
-				    if ( is_object( $current_coauthor_term ) && wpcom_vip_term_exists( $current_coauthor_term->term_id ) ) {
-					    $current_user_query = $wpdb->term_taxonomy . '.taxonomy = \'' . $this->coauthor_taxonomy . '\' AND ' . $wpdb->term_taxonomy . '.term_id = \'' . $current_coauthor_term->term_id . '\'';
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -725,10 +725,12 @@ function posts_where_filter( $where, $query ) {
 					$current_coauthor      = $this->get_coauthor_by( 'user_nicename', wp_get_current_user()->user_nicename );
 					$current_coauthor_term = $this->get_author_term( $current_coauthor );
 
-					$current_user_query  = $wpdb->term_taxonomy . '.taxonomy = \''. $this->coauthor_taxonomy.'\' AND '. $wpdb->term_taxonomy .'.term_id = \''. $current_coauthor_term->term_id .'\'';
-					$this->having_terms .= ' ' . $wpdb->term_taxonomy .'.term_id = \''. $current_coauthor_term->term_id .'\' OR ';
+					if ( $current_coauthor_term ) {
```
sboisvert: One thing that came to my mind while reviewing this (And I'm just mentioning it for completeness) was: 
Could this ever be a "truthy" result and still not be what we're expecting?
From what I can tell it seems get_author_term() runs get_term_by() which does some checks against is_wp_error() so the code shouldn't ever return a wp_error which is good. The function declaration of get_term_by() does mention that it could return an array but that seems to only be if $output is ARRAY_(A|N) so it shouldn't be a problem.
All that to say that sometime just checking is something is "true" can lead to unexpected results. PHP's type system can lead to unexpected results http://php.net/manual/en/types.comparisons.php

One small thing I would change is that it seems like the indentation is being changed here. Probably a case of spaces vs tabs, depending on your IDE you should be able to set it to ensure it follows what the file is currently using.

k-atul:
Hi by when will these changes get merged to WP non-vip plugin its been over 7 months?
https://wordpress.org/plugins/co-authors-plus/

PR #523, opened 2018-05-31
Added FAQ to readme on how to disable Guest Authors

### LINKED ISSUE #122: Include mention of how to disable guest authors in readme FAQ
It's becoming a common question.

TheCrowned:
Fix #122, where it was proposed to mention how to disable Guest Authors in a readme FAQ. 

I hope I have not been too short in the FAQ Answer, but I think that 

1. a lengthy explanation such as 
_there exists filter `name` that can handle a boolean value 
blabla
blabla_
was quite unneeded. I think a developer would be pissed off by that.

2. it is not the place to put a mini-tutorial on how to disable Guest Authors expecting virtually zero knowledge on the reader's part, such as
_put the following code in your theme functions.php -- 10 lines of code provided, with functions and `init` hooks._
If you don't know how to have a filter loaded on `init`, then a) you probably would not be reading a readme on github; b) you shouldn't be tweaking code in the first place.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -41,6 +41,11 @@ To create new guest author profiles, a WordPress will need the 'list_users' capa
 
 Yep! There's a template tag called `coauthors_wp_list_authors()` that accepts many of the same arguments as `wp_list_authors()`. Look in template-tags.php for more details.
 
+= Can I disable Guest Authors?
+
+Yep! Guest authors can be disabled entirely through an apt filter. Having the following line load on `init` will do the trick:
+`add_filter( 'coauthors_guest_authors_enabled', '__return_false' )`
+
 == Upgrade Notice ==
 
 = 3.1 =


sboisvert:
I really like that you put down the thought process here and the "Why"

I'm not sure my experience would match with:
>If you don't know how to have a filter loaded on init, then a) you probably would not be reading a readme on github; b) you shouldn't be tweaking code in the first place.

I think everyone starts somewhere and I don't know if I would of known how to add an action on init at all the points of my career. I don't think there is anything inherently wrong with learning by copy pasting code you find online :). All that to say that I think if it needed to be on init we should have it as part of the example. 
That being said, why do we need to only add the filter on the init hook?

TheCrowned:
> I think everyone starts somewhere and I don't know if I would of known how to add an action on init at all the points of my career. I don't think there is anything inherently wrong with learning by copy pasting code you find online :)

Oh of course - I may have sounded harsher than I meant. Let me elaborate.

I agree there's nothing wrong in copy pasting code, I have done it so many times! I thought that the readme shouldn't be the place because other answers were pretty coincise and mostly hinted at the solution, rather than walking through it step by step. 

Also, no specific reason to have it on init only - it just seemed a natural place in which to put it. There may be other hooks that work just as well, but again, I would leave that to the experience of the reader (while still giving an "answer" if you don't know what to hook to). Hope this helps clarifying! :)

Issue #525, opened 2018-05-31
Clicking on "No authors found" will add an empty author


TheCrowned:
**Steps to reproduce:**
1. Go to a post edit page
2. In the Authors box, enter a query that will yield no results
3. Click on the message that says that no results were found
4. A blank author is added

![cap](https://user-images.githubusercontent.com/7880569/40778570-e302f4e6-64d1-11e8-9dd4-5a3e8df687b6.gif)

When saving the post, the empty author doesn't create any issue and is just ignored, but it would be nice not to have anything happen when clicking on the error.

nielslange:
Closing this based on https://github.com/Automattic/Co-Authors-Plus/pull/527#event-2507133178.

rebeccahum:
@nielslange Is this issue still reproducible?

nielslange:
> @nielslange Is this issue still reproducible?

Nope, it's not. Empty authors can no longer be added.

PR #527, opened 2018-05-31
Fix no result creating a empty author

mariovalney:
Fixing issue #525, we should ignore selecting a author result without enough data (id, login, name and email).

In meanwhile, "Sorry, no matching authors found." string is now translatable.

_I guess this is the best approach without refactoring suggest.js to add a "No Result" option._


rebeccahum:
This fix didn't work for me.

Issue #518, opened 2018-05-29
Data Exporter Support


mjangda:
WordPress 4.9.6 includes the ability for users to request an export with details on personal data stored on the site. We should include matching Guest Author entries (and relevant details) as part of that export.

We should also make it possible for extensions like https://github.com/Automattic/Co-Authors-Plus-Social-Pack hook in and add additional details as well.

sboisvert:
@UP_Nguyen this is a test

mariovalney:
Thanks for your support @mjangda

PR #528, opened 2018-05-31
Allow guest authors have personal data exported

mariovalney:
This add the enhancement #518 to allow guest authors export their data as a regular author.

guest-author

A new filters was introduced: **coauthors_guest_authors_exported_extra_data**.


php/class-coauthors-guest-authors.php
```diff
@@ -1503,4 +1506,93 @@ public function filter_author_feed_link( $feed_link, $feed ) {
...
+
+		if ( ! $author ) {
+			return array(
+				'data' => array(),
+				'done' => true,
+			);
+		}
+
+		$author_data = array(
+			'ID'              => __( 'Guest Author ID', 'co-authors-plus' ),
```
mjangda: I think we can remove `Guest Author` from the labels here as it's a bit redundant.

php/class-coauthors-guest-authors.php
```diff
@@ -1503,4 +1506,93 @@ public function filter_author_feed_link( $feed_link, $feed ) {
...
+			'aim'             => __( 'Guest Author AIM', 'co-authors-plus' ),
+			'yahooim'         => __( 'Guest Author Yahoo IM', 'co-authors-plus' ),
+			'jabber'          => __( 'Guest Author Jabber / Google Talk', 'co-authors-plus' ),
+			'description'     => __( 'Guest Author Biographical Info', 'co-authors-plus' ),
+		);
+
+		$author_data_to_export = array();
+
+		foreach ( $author_data as $key => $name ) {
+			if ( empty( $author->$key ) ) continue;
```
mjangda: Minor: add braces for the `if` and move `continue` to new line.

php/class-coauthors-guest-authors.php
```diff
@@ -1503,4 +1506,93 @@ public function filter_author_feed_link( $feed_link, $feed ) {
...
+		/**
+		 * Filters extra data to allow plugins add data related to guest author
+		 *
+		 * @since 3.3.1
+		 *
+		 * @param array $extra_data A empty array to be populated with extra data
+		 * @param int $author->ID The guest author ID
+		 * @param string $email_address The guest author email address
+		 */
+		$extra_data = apply_filters( 'coauthors_guest_authors_exported_extra_data', [], $author->ID, $email_address );
```
mjangda: Instead of a separate array that we have to merge in, maybe we should just pass `$author_data_to_export` and other plugins can append to it directly?

mjangda: Also, for consistency, we should call this `coauthors_guest_author_personal_export_data`.

php/class-coauthors-guest-authors.php
```diff
@@ -1503,4 +1506,93 @@ public function filter_author_feed_link( $feed_link, $feed ) {
...
+	 * @since 3.3.1
+	 */
+    public function filter_personal_data_exporter( $exporters ) {
+        $exporters['cap-guest-author'] = array(
+            'exporter_friendly_name' => __( 'Guest Author', 'co-authors-plus' ),
+            'callback'               => array( $this, 'personal_data_exporter' ),
+        );
+
+        return $exporters;
+    }
```
mjangda: Minor: indentation looks off here.

php/class-coauthors-guest-authors.php
mariovalney: Done

php/class-coauthors-guest-authors.php
mariovalney: Filter changed but I think we should keep the merge. This way we don't allow other plugins to remove our data, which I think is mandatory to keep because GDPR

php/class-coauthors-guest-authors.php
mariovalney: Done

php/class-coauthors-guest-authors.php
mariovalney: I added it because user export has "User" label as well. But I agree it's redundant.

mariovalney:
Thank you @mjangda ! I think it's ok now.

php/class-coauthors-guest-authors.php
mjangda: That makes sense @mariovalney. I think because of we should rename the filter to `coauthors_guest_author_personal_export_extra_data`.

php/class-coauthors-guest-authors.php
```diff
@@ -1503,4 +1506,95 @@ public function filter_author_feed_link( $feed_link, $feed ) {
...
+			'jabber'          => __( 'Jabber / Google Talk', 'co-authors-plus' ),
+			'description'     => __( 'Biographical Info', 'co-authors-plus' ),
+		);
+
+		$author_data_to_export = array();
+
+		foreach ( $author_data as $key => $name ) {
+			if ( empty( $author->$key ) ) {
+                continue;
+            }
```
mjangda: Looks like the indentation is off here too.

mjangda:
Thank you again @mariovalney! I just noticed two small things. We can merge after those are fixed.

php/class-coauthors-guest-authors.php
mariovalney: Sublime is playing with me hahaha

I guess we should consider to add a `.editoconfig`

php/class-coauthors-guest-authors.php
mariovalney: So... back to "coauthors_guest_author_personal_export_extra_data" ?

mariovalney: Done

Issue #530, opened 2018-06-04
action_pre_user_query is useless and should be removed


TheCrowned:
`action_pre_user_query()` modifies `get_users()` behavior to search for `display_name` instead of `user_nicename`. It is added and removed as a filter when calling get_users() in `search_authors()`.

However, this was only needed before PR #519, when no argument `search_columns` was provided. In that case, the default WP behavior was to search in a bunch of fields that included `user_nicename` but not `display_name`. Now `user_nicename` is not among the search fields anymore, while `display_name` is, so that filter is useless.

PR #531, opened 2018-06-05
Removed function action_pre_user_query and its hooks

TheCrowned:
Fix for #530.

`action_pre_user_query()` had the purpose of tweaking `get_users()` behavior to search for `display_name` instead of `user_nicename`. It was added and removed as a filter when calling `get_users()` in `search_authors()` - and nowhere else.

However, this was only needed before PR #519, when no argument `search_columns` was provided to the `get_users()` call. In that case, the default WP behavior was to search in a bunch of fields that included `user_nicename` but not `display_name`. PR #519 provides the argument `search_columns` correctly and with that, `user_nicename` is not among the search fields anymore, while `display_name` is. The action has thus become obsolete and was just slowing every search query down.

**Note**: depends on #519 (already merged, at the time of writing).

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1174,9 +1174,7 @@ class CoAuthors_Plus {
 				),
 				'fields' => 'all_with_meta',
 			);
-		add_action( 'pre_user_query', array( $this, 'action_pre_user_query' ) );
 		$found_users = get_users( $args );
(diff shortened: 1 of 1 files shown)


sboisvert:
I really like the commit message here. Very descriptive of the "Why" this was done in the past and why that's no longer needed. Very well done 💯

PR #532, opened 2018-06-05
Fix possible discrepancy in author search with ignored authors

### LINKED ISSUE #273: Discrepancy between submitted coauthors and values returned through get_coauthors
Hello,

I believe there's a bug in the Coauthors initialization which stores the original author in the system with their login as their username.

I've written a save_post action on my LMS to add/remove coauthors on the course to it's associated Buddypress Group.  Part of it is doing a check to determine the added/removed coauthors which compares the submitted coauthors against the existing coauthors for the course. Sadly I've run into a case where the coauthors returned through get_coauthors have a different case for the user_login than those submitted. It appears that it's only for the primary coauthor who is most likely the original post author prior to Coauthors Plus being installed.

So the following code;

```
$new_coauthors = (array) $_POST['coauthors'];
$new_coauthors = array_map( 'sanitize_text_field', $new_coauthors );
write_log("New Coauthors = ".print_r($new_coauthors,1));
$existing_coauthors = wp_list_pluck( get_coauthors( $post_id ), 'user_login' );
write_log("Existing Coauthors = ".print_r($existing_coauthors,1));
```

Provides the results;

```
[23-Jul-2015 18:40:07 UTC] New Coauthors = Array
(
    [0] => lj-lee
    [1] => andrea_thomson
    [2] => bret_swigard
    [3] => leigh_negin
    [4] => lindsay_geddes
    [5] => julie_block
)

[23-Jul-2015 18:40:07 UTC] Existing Coauthors = Array
(
    [0] => LJ-Lee
    [1] => andrea_thomson
    [2] => bret_swigard
    [3] => leigh_negin
    [4] => lindsay_geddes
)
```

And causes this comparison code;

```
$added_coauthors = array_diff( $new_coauthors, $existing_coauthors );
write_log("Added Coauthors = ".print_r($added_coauthors,1));
$removed_coauthors = array_diff( $existing_coauthors, $new_coauthors );
write_log("Removed Coauthors = ".print_r($removed_coauthors,1));
```

To show the one user as both being added and removed since their case on the user_login is different.

```
[23-Jul-2015 18:40:07 UTC] Added Coauthors = Array
(
    [0] => lj-lee
    [5] => julie_block
)

[23-Jul-2015 18:40:07 UTC] Removed Coauthors = Array
(
    [0] => LJ-Lee
)
```

Removing and re-adding this lj-lee user fixes her case moving forward. And I'm going through doing this for all my courses to avoid an issue. As well I'm adding a safety check prior to the array_diff to lowercase both arrays. But I wanted to raise this as a potential issue and see if there's an easy way to mass update all instances of authors to the right lowercase version of user_login.

Let me know if I can assist in any troubleshooting.

Thanks

TheCrowned:
This fixes #273.

The issue was that, in some cases, it was possible to add the same author more than once. Even though it was already a co-author of the post, it would still show up in search and be available for another addition - here, the last suggestion is the same user who is already an author of the post:

before

**_tl;dr - the issue was due to JS referencing existing authors by `user_nicename` while PHP by `user_login`._** 

Getting a peak at what happened when a post is saved, it was clear that there was some mismatch between the newly-added coauthors and the already present ones. It looked like some case issue:

screenshot from 2018-06-05 09-20-20

I expected this to be an issue with `search_authors()` - in particular, something to do with `$ignored_authors`. Indeed, diving into the JS and PHP, I found out it was an issue with how ignored authors are handled by the two sides. 

JS takes care of making the AJAX request to `ajax_suggest()`, which in turn calls `search_authors()`. It sends `existing_authors` as parameter, populating it with current authors `user_nicename`s.

```js
function coauthors_create_author_hidden_input ( author ) {
	var input = jQuery( '<input />' )
		.attr({
			'type': 'hidden',
			'id': 'coauthors_hidden_input',
			'name': 'coauthors[]',
			'value': decodeURIComponent( author.nicename )
		});
	return input;
}

jQuery( document ).ajaxSend(function( e, xhr, settings ) {
	var existing_authors = jQuery( 'input[name="coauthors[]"]' ).map(function(){return jQuery( this ).val();}).get();
```

But on the other side, PHP was deciding whether an author should be excluded from search comparing `user_login`s with the AJAX-supplied list of `user_nicename`s. This would work fine in several situations - basically, all the times `user_login` did not have any special chars or uppercase letters. But when `user_nicename` was different from `user_login`, this was clearly an issue.

```php
$found_users[ $found_user->user_login ] = $found_user;

foreach ( $found_users as $key => $found_user ) {
	if ( in_array( $found_user->user_login, $ignored_authors ) ) { 
		unset( $found_users[ $key ] );
	}
}
```

I thus changed `user_login` with `user_nicename` in the PHP bit. After this edit, the issue has gone! :)

after

This edit does not affect anything else, as far as I am been able to check. All PHPUnit tests still succeed as they did before. Also, `search_authors()` is only called by `ajax_suggest()`, so it should really only affect its behavior (and that's what my tests have confirmed) :)

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1212,7 +1212,7 @@ class CoAuthors_Plus {
 		$ignored_authors = apply_filters( 'coauthors_edit_ignored_authors', $ignored_authors );
 		foreach ( $found_users as $key => $found_user ) {
 			// Make sure the user is contributor and above (or a custom cap)
-			if ( in_array( $found_user->user_login, $ignored_authors ) ) {
+			if ( in_array( $found_user->user_nicename, $ignored_authors ) ) { //AJAX sends a list of already present *users_nicenames*
 				unset( $found_users[ $key ] );
 			} else if ( 'wpuser' === $found_user->type && false === $found_user->has_cap( apply_filters( 'coauthors_edit_author_cap', 'edit_posts' ) ) ) {
 				unset( $found_users[ $key ] );
--- a/tests/test-coauthors-plus.php
+++ b/tests/test-coauthors-plus.php
@@ -367,7 +367,7 @@ class Test_CoAuthors_Plus extends CoAuthorsPlus_TestCase {
 		$authors = $coauthors_plus->search_authors();
 
 		$this->assertNotEmpty( $authors );
-		$this->assertNotContains( $subscriber1->user_login, $authors );
+		$this->assertArrayNotHasKey( $subscriber1->user_login, $authors );
 
 		// Checks when search term is empty and any contributor exists.
 		$contributor1 = $this->factory->user->create_and_get( array(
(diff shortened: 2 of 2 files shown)


mdbitz:
Pending @philipjohn for Merge or feedback.

garrett-eclipse:
Thanks for taking the time to address my issue guys. I appreciate the thorough recon and explanation provided by @TheCrowned

PR #534, opened 2018-06-05
fix "Trying to get property of non-object" warning

### LINKED ISSUE #533: "Trying to get property of non-object" warning in class-coauthors-wp-list-table.php
On /wp-admin/users.php?page=view-guest-authors&orderby=last_name&order=asc, I'm seeing this error.

```
Notice: Trying to get property of non-object in /srv/www/fivethirtyeight/public_html/wp-content/plugins/vip/co-authors-plus/php/class-coauthors-wp-list-table.php on line 256
```

paulschreiber:
Fixes #533.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
     include:
         # aliased to a recent 5.6.x version
         - php: "5.6"
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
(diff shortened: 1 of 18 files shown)


GaryJones:
The `column_posts()` method in the master branch no longer a reference to `->linked_account` in it, so I think this may have been handled already.

PR #535, opened 2018-06-05
🐛Only filter author template for title

jacobarriola:
Repro steps:
- in `archive.php`, use `the_archive_title()` method
- load a category archive page

Expected results:
"Category: category_name"

Actual results:
null

Notes:
The `filter_author_archive_title()` function is hooked to the `get_the_archive_title` filter, which can run in other archive templates such as `category` or `tag`. As a result, the function only returns a value when `is_author()` returns `true`, thereby resulting in `null` instead of `$title` for all other archive templates.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1610,13 +1610,22 @@ class CoAuthors_Plus {
 
 	/**
 	 * Filter of the header of author archive pages to correctly display author.
+	 *
+	 * @param $title
+	 *
+	 * @return string
 	 */
-	public function filter_author_archive_title() {
-		if ( is_author() ) {
-			$author_slug = sanitize_user( get_query_var( 'author_name' ) );
-			$author = $this->get_coauthor_by( 'user_nicename', $author_slug );
-			return sprintf( __( 'Author: %s' ), $author->display_name );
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -1610,13 +1610,22 @@ public function clear_cache_on_terms_set( $object_id, $terms, $tt_ids, $taxonomy
 
 	/**
 	 * Filter of the header of author archive pages to correctly display author.
+	 *
+	 * @param $title
+	 *
+	 * @return string
```
TheCrowned: This is just a picky thing, but if you get the time, would you mind completing the inline doc following the standard `$var_name var_type short_var_description` format? Since you have added them, which is great, we may also complete them for future reference :)

TheCrowned:
Thanks for your contribution! Other than the small note on the online function comment, this fix is great! I can confirm the issue and that that this PR fixes it :)

jacobarriola:
@TheCrowned glad to help. I resolved the merge conflicts and added the comments to the function 👍

msassmann:
The issue still exists in current version?

rebeccahum:
@msassmann Have you pulled from master?

PR #421, opened 2017-05-30
Replaced with cap_notify_coauthors

rebeccahum:
As per issue #403, new function cap_notify_coauthors hooks onto comment_notification_recipients filter and replaces old custom function wp_notify_postauthor.  

Note: This would render PR #412 (issue #411) unnecessary.

PR #539, opened 2018-06-08
Splitting out functions from main file

### LINKED ISSUE #396: Separate out some functions
`co-authors-plus.php` is pretty big and unnecessarily so. It should be split up into more manageable chunks.

For example, `filter_ef_calendar_item_information_fields()` and `filter_ef_story_budget_term_column_value()` could move into an Edit Flow compatibility file instead.

Admin UI functionality could be split into a separate file/class too.

TheCrowned:
This fixes #396. The main CAP file was unnecessarily big, and some functions could be taken out of it to reduce its size.

I have done that basically following the [plan I had outlined](https://github.com/Automattic/Co-Authors-Plus/issues/396#issuecomment-393106849) in #396. All the functions that I moved are now in standalone files (i.e. not wrapped in classes) and prefixed with **cap\_**. I have updated all relevant hooks as well, of course!

In some cases, I have taken the chance to update the functions inline documentation and also make them adherent to our coding standards. 

New files are:
- php/coauthors-edit-screen.php - _Functions that change the default behavior of the edit.php page._
- php/coauthors-users-screen.php - _Functions that change the default behavior of the users.php page._
- php/coauthors-meta-box.php - _Functions related to metabox handling._
- php/integrations/edit-flow.php - _Functions related to Edit Flow integration._
- php/integrations/jetpack.php - _Functions related to Jetpack integration._

All functions have been moved out (i.e. **deleted**) from the main file, **except** the metabox ones, for which I have left helper functions where they were before, so that any external code reference to those would not break. 

I skipped the planned _php/email-notifications.php_ one since those functions have actually been (or are going to be?) removed as per PR #421. Those may me moved, although they would be a small change, when the relevant PR is merged.

Before splitting, I created a bunch of **new unit tests** for the functions I was going to move. Since my new tests were covering functions I had moved out of the main file, I moved the tests out of the main test file as well, resulting in the creation of:
- tests/test-edit-screen.php
- tests/test-integration-edit-flow.php
- tests/test-users-screen.php

All unit tests worked before the split, and they still work afterwards. I have tested CAP functionalities (at least, the _normal/basic_ ones, if that makes sense) and I found nothing broke as a result of my changes. 

As a final note, although this look like a huge change set, most of the changes are just copy-pasting code around, so there was little room for breaking.


co-authors-plus.php
```diff
@@ -314,221 +319,30 @@ public function is_post_type_enabled( $post_type = null ) {
 	}
 
 	/**
+	 * DEPRECATED - Use cap_remove_authors_box() instead!
```
emrikol: For these deprecated functions, can we call `_deprecated_function()` to throw a warning and add a `@deprecated` and/or `@see` tag (Kinda like https://github.com/WordPress/WordPress/blob/505c2ddb61b1509c49bd0d9e8328c6a1856f6827/wp-admin/includes/deprecated.php#L97)

TheCrowned: Sure - will update that! What version should I use in the deprecated notice? 

`_deprecated_function( __FUNCTION__, '??????', 'cap_remove_authors_box()' );`

mdbitz: Lets use 3.3.0 as the version as I don't think we've nailed down where this will be integrated to the code can be updated again when it is merged if needed to align to the latest version number.

mdbitz:
@philipjohn I'm adding you as a review for this. Thanks in advance!

GaryJones:
While I'm greatly in favour of organising the codebase, and someone could use the commits here as ideas and inspiration for how to do it, it's clear that this PR is so far behind and conflicted that it would be significantly easier to create fresh PRs with much smaller iterative changes. As such, I'm going to close this one out, but not because of the tremendous amount of effort that had gone into it!

PR #542, opened 2018-06-08
PHPDoc improvements

paulschreiber:
Adds missing PHPDoc comments. Fixes warnings.

Assumes presence/merging of `fix/phpcs`.


GaryJones:
This can't be merged as is, but I'm adding this to a future milestone so that it can be used as inspiration for fresh PRs.

GaryJones:
Thanks again for opening this back in 2018, @paulschreiber — and apologies for the long wait. I've taken another look as part of clearing out older PRs.

Two things have changed since I last revisited this. The first is that the codebase has moved on a long way: a lot of the cosmetic work here (PHPCS-style alignment, comment punctuation, array reformatting) and the structural pieces (composer/CI scaffolding, the test files, the GDPR exporter docblocks) have all been done in some form on `develop`. The second is that, after combing through the diff, the genuinely "new docblock" content that isn't already in the codebase is quite small — and the few public functions that still have no docblock at all (`get_coauthors()`, `coauthors__echo()`, `the_coauthor_meta()`, and `coauthors_plus_upgrade()`) aren't actually covered by this PR.

So rather than try to rebase 8 years of drift, I've opened #1265, which is a tiny, scoped PR that adds docblocks to exactly those four remaining functions. That seems like a more honest way to land the PHPDoc improvement this PR was reaching for.

Closing this one in favour of #1265. Really appreciate the original push to get the documentation in better shape — even where the diff itself didn't survive, the prompt was a useful one.

Issue #235, opened 2014-11-20
Ajax suggest metabox shows 0 results frequent


dpacmittal:
In search_authors() function:

```
$args = array(
    'search' => $search,
    'get' => 'all',
    'number' => 10,
);
$args = apply_filters( 'coauthors_search_authors_get_terms_args', $args );
add_filter( 'terms_clauses', array( $this, 'filter_terms_clauses' ) );
$found_terms = get_terms( $this->coauthor_taxonomy, $args );
remove_filter( 'terms_clauses', array( $this, 'filter_terms_clauses' ) );

if ( empty( $found_terms ) )
    return array();
// Get the co-author objects
$found_users = array();
foreach( $found_terms as $found_term ) {
    $found_user = $this->get_coauthor_by( 'user_nicename', $found_term->slug );
    if ( !empty( $found_user ) )
        $found_users[$found_user->user_login] = $found_user;
}

```

This searches for only 10 terms. The array is then filtered using get_coauthor_by function. Sometimes this leaves the resultset to be empty, when it shouldn't be (because we only looked for 10 terms in the first place). 

We should increase the 'number' to more sensible value, or find a better way (using SQL?) to get coauthors from the terms. Another solution would be to loop it until we have atleast X number of results.

rohitkeshwani07:
any update on this?

TheCrowned:
So, as of now a coauthor selected by `get_terms()` can be dropped if
- it is in the ignored set (for ex. it already belongs to the post we are assigning coauthors to)
- it does not have sufficient permissions.

The first case can be filtered out in `get_terms()` itself with the `exclude` argument, so that' good. The second case can't really be filtered out _a priori_, as far as I know. 

One solution would be to select maybe 20 coauthors already excluding the `$ignored` ones, and then filtering for permissions. At this point I expect the resulting set to contain something, at least, but possibly more than 10 coauthors. We may then return at most 10 of them. What do you guys think?

philipjohn:
Even with a limit of 20 we're likely to hit the same issue on a site with a large number of contributors.

What's stopping us from bumping the limit up to something like 100? What effect would that have on the query - would it make it too expensive? If so, could we be clever with caching to limit the impact?

TheCrowned:
[...]

PR #544, opened 2018-06-13
Stop handling wp_terms in search_authors, do it in install and profile update

### LINKED ISSUE #486: Adding 'extra' authors to the wp_term_taxononmy
I've been testing this plug-in in a development environment to see if we want to use it and I have a question/concern about something I'm seeing in the database. It seems to be adding entries to the `wp_term_taxononmy` table even if all I am doing is searching for an author in the admin interface. Is that "normal" behavior?

TheCrowned:
This fixes #486. The issue was that the plugin created `wp_terms` for all users upon each call to `search_authors()`, regardless of whether they could be added as coauthors. This ended up flooding the database with tons of unneeded entries, in particular on sites with many subscribers.

_**tl;dr the fix is: stop creating/updating wp_terms on search; instead, do it in bulk on plugin install and in profile updating.**_

What I have done is:

1. **Set up a cron job on plugin activation** 

It selects WP users sorting by _registration time ASC_ and creates the needed author term for each of them, provided they have the capability of being added as coauthors. This is also a 80% fix for #235, because users that should never show up in search do not get a `wp_term`. 

At the end, the cron job just re-spawns itself as a new one, keeping track of how many users have already been processed. The process automatically stops (leaving no cron jobs behind) when the existing users import is complete. It updates 100 users every time. It may be _just a little bit_ heavy, but keep in mind that **until this runs, no authors are found in the Co-Authors search box!**

(Side note: sorting by registration time ASC ensures that, even on sites that get new users every few seconds, the import works fine. Indeed, the import deals first with non-changing data and tackles recent data at the end. It is guaranteed to cover all users. In truth, this is just an extra-safety measure, because all users created after CAP installation would have their term created as per 2. ).

2. **Create/Update author terms when a user profile is updated**

This is what makes sure terms are kept updated after plugin installation.

3. **Lighten the `search_authors()` function** 

Before, it used to the following:
- query for WP users and update author term for selected users
- query for terms. These would be where the _real_ search results would come from
- foreach the selected users and exclude the ones that did not have permission to be added as coauthors.

Now, it
- queries for terms (search results)
- foreaches the selected users and exclude the ones that do not have permission to be added as coauthors. However, when a user is found not to have enough permissions BUT has the CAP wp_term, the wp_term is deleted so that the user will not be selected in future searches. **This will gradually de-pollute the database!** (We could not just strip this check because current CAP installations have subscribers with CAP term.)

4. **Updated PHPUnit tests to work**

When testing `search_authors()`, the tests created some users and then immediately called `search_authors()`, relying on the fact that it would create the missing wp_term for the new users. This is no longer the case, so we need to call `cap_create_author_terms()` after creating the users. It defaults to updating 100 users, so if ever we will test creating more than 100 users, will need to call it like `cap_create_author_terms(0, 1000)`.

All tests succeed now after the last commit.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -37,6 +37,7 @@ define( 'COAUTHORS_PLUS_VERSION', '3.3.0' );
 require_once( dirname( __FILE__ ) . '/template-tags.php' );
 require_once( dirname( __FILE__ ) . '/deprecated.php' );
 
+require_once( dirname( __FILE__ ) . '/php/coauthors-install.php' );
 require_once( dirname( __FILE__ ) . '/php/class-coauthors-template-filters.php' );
 require_once( dirname( __FILE__ ) . '/php/integrations/amp.php' );
 
@@ -68,6 +69,8 @@ class CoAuthors_Plus {
 	 * __construct()
 	 */
 	function __construct() {
+		//Set up on activation
+		register_activation_hook( __FILE__, 'cap_install_setup' );
 
 		// Register our models
 		add_action( 'init', array( $this, 'action_init' ) );
@@ -131,6 +134,10 @@ class CoAuthors_Plus {
 
 		// Filter to correct author on author archive page
 		add_filter( 'get_the_archive_title', array( $this, 'filter_author_archive_title'), 10, 2 );
+
+		// Update author term on user update
+		add_action( 'user_register', array( $this, 'action_user_profile_update' ) );
+		add_action( 'profile_update', array( $this, 'action_user_profile_update' ) );
 	}
 
 	/**
@@ -1221,6 +1228,7 @@ class CoAuthors_Plus {
 		return (array) $found_users;
 	}
 
+
 	/**
 	 * Modify get_terms() to LIKE against the term description instead of the term name
 	 *
@@ -1444,6 +1452,26 @@ class CoAuthors_Plus {
 		}
 		wp_cache_delete( 'author-term-' . $coauthor->user_nicename, 'co-authors-plus' );
 		return $this->get_author_term( $coauthor );
+	}
+
+	/**
+	 * Update author term when user profile is updated.
+	 *
+	 * @param int $userid
+	 */
+	function action_user_profile_update( $userid ) {
+		global $coauthors_plus;
+
+		if( ! current_user_can( 'manage_options' ) ) {
+			return;
+		}
+
+		$user = get_userdata( $userid );
+
+		// Update author term only if user can be added as coauthor
+		if( $user->has_cap( apply_filters( 'coauthors_edit_author_cap', 'edit_posts' ) ) ) {
+			$coauthors_plus->update_author_term( $user );
+		}
 	}
 
 	/**
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1172,26 +1172,6 @@ class CoAuthors_Plus {
 	 */
 	public function search_authors( $search = '', $ignored_authors = array() ) {
 		$args = array(
-				'count_total' => false,
-				'search' => sprintf( '*%s*', $search ),
-				'search_columns' => array(
-					'ID',
-					'display_name',
-					'user_email',
-					'user_login',
-				),
-				'fields' => 'all_with_meta',
-			);
-		$found_users = get_users( $args );
-
-		foreach ( $found_users as $found_user ) {
-			$term = $this->get_author_term( $found_user );
-			if ( empty( $term ) || empty( $term->description ) ) {
-				$this->update_author_term( $found_user );
-			}
-		}
-
-		$args = array(
 			'search' => $search,
 			'get' => 'all',
 			'number' => 10,
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1442,10 +1442,6 @@ class CoAuthors_Plus {
 	function action_user_profile_update( $userid ) {
 		global $coauthors_plus;
 
-		if( ! current_user_can( 'manage_options' ) ) {
-			return;
-		}
-
 		$user = get_userdata( $userid );
 
 		// Update author term only if user can be added as coauthor
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1466,7 +1466,7 @@ class CoAuthors_Plus {
 	}
 
 	/**
-	 * Update author term when user profile is updated.
+	 * Create guest author profile when WP user profile is updated.
 	 *
 	 * @param int $userid
 	 */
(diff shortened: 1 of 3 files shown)


co-authors-plus.php
```diff
@@ -1211,11 +1197,13 @@ public function search_authors( $search = '', $ignored_authors = array() ) {
 		// Allow users to always filter out certain users if needed (e.g. administrators)
 		$ignored_authors = apply_filters( 'coauthors_edit_ignored_authors', $ignored_authors );
 		foreach ( $found_users as $key => $found_user ) {
-			// Make sure the user is contributor and above (or a custom cap)
 			if ( in_array( $found_user->user_login, $ignored_authors ) ) {
 				unset( $found_users[ $key ] );
+
+			//If user does not have permission, they should not have a term in the first place
```
sboisvert: I think this might be mistaken. For example we have an editor and they leave the site so we downgrade them to contributor (or even subscriber). We'd still want them to be co-authors on pieces and not remove them completely from the site and from all the bylines if they can't edit posts anymore.

I'm also not sure a change like this should be bundled with this PR. It feels a bit like there was another potential issue that was found while working on the main issue and the fix gets tagged along. Creating multiple issues might be better in the long run.

co-authors-plus.php
```diff
@@ -1446,6 +1434,26 @@ public function update_author_term( $coauthor ) {
...
 
+	/**
+	 * Update author term when user profile is updated.
+	 *
+	 * @param int $userid
+	 */
+	function action_user_profile_update( $userid ) {
+		global $coauthors_plus;
+
+		if( ! current_user_can( 'manage_options' ) ) {
```
sboisvert: Is manage_options the permission used for managing authors? It seems this would mean that when a profile page is changed this might not run. Is this intentional?

php/coauthors-install.php
```diff
@@ -0,0 +1,57 @@
...
+	global $coauthors_plus;
+
+	$args = array(
+		'orderby' => 'registered',
+		'order' => 'ASC',
+		'offset' => $imported_count,
+		'number' => $number_to_update,
+		'fields' => 'all_with_meta',
+	);
+	$found_users = get_users( $args );
```
sboisvert: I suspect that we might want to test this query. I have a feeling that if this query were ever to be run on a very large site it could break things since there is no index on registered. This means MySQL would have to try to do a table sort on all the xyz number of users. I'm also unsure of how this is handled in a multi-site, clearly the UI can scale but those are ordered by user_nicename which is indexed. I think we should have a look at the types of queries this generates  (and potentially even benchmarking it) before merging.

php/coauthors-install.php
```diff
@@ -0,0 +1,57 @@
...
+		if( ! $found_user->has_cap( apply_filters( 'coauthors_edit_author_cap', 'edit_posts' ) ) ) {
+			continue;
+		}
+		
+		$coauthors_plus->update_author_term( $found_user );
+	}
+
+	$imported_count += $number_to_update;
+
+	wp_schedule_single_event( time(), 'cap_import_existing_users', array( $imported_count ) );
```
sboisvert: Should we potentially give the site a bit of time to breath in between each batch since these would most likely be up to 100 write operations in a short time span.

co-authors-plus.php
TheCrowned: Spot on! You are right. I intended that to be a security measure, so that I would be _doing_it_right, but ended up backfiring. There is the `edit_users` cap for managing users, but it would still not run when one updates their own profile. I'll just remove that check.

co-authors-plus.php
TheCrowned: > we have an editor and they leave the site so we downgrade them to contributor

Thanks for bringing this up, I hadn't thought about it! I still think this change would be a good one to remove a lot of garbage from the database, especially for large sites with lots of subscribers. Would also checking whether the user is coauthor of any pieces fix it? 

> It feels a bit like there was another potential issue that was found while working on the main issue and the fix gets tagged along

At first, I had actually completely removed that check, thinking that after my changes all users who had a term were good enough to be displayed, since I was creating terms only for relevant users. Of course, I hadn't thought about existing terms created in previous plugin version, so I restored that back. And then I also thought: "my new feature is going to leave databases quite clear, but only on new installations. Can I do something for old ones as well?". So that's its story, just to let you know what brought me to edit that piece - but I can certainly split it in a different PR if you believe would be more fit :)

TheCrowned:
> it does seem to duplicate somewhat the functionality in here: https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-wp-cli.php#L20

We do similar things, but it looks like we are doing them very differently (which makes me wonder whether I am doing them all right?). 

The CLI function calls `create_guest_author_from_user_id()` and creates guest author entries for each WP user. As far as I understand, that means that it creates a CAP profile and links it to the WP user. This shouldn't be required, as CAP can add as coauthors even simple WP users - so what I am doing is just creating the relevant `wp_term` for them with `update_author_term()`. Is that right?

php/coauthors-install.php
TheCrowned: > (Side note: sorting by registration time ASC ... is just an extra-safety measure, because all users created after CAP installation would have their term created as per 2. ).

**tl;dr I can sort by `user_nicename` and nothing bad will happen.**

Just to give you a bit of context on the _why_, I implemented this function _before_ the one that updated author term on profile update. So it seemed neat to make it process in a way that would cover all of them, even if some new ones popped up after the first cron had run. Then I implemented `action_user_profile_update()` and the kind of sorting here became irrelevant. But I decided to keep it because I thought it was neat, except it wasn't because I didn't know about the index thing you mentioned.

_Moral: I fell into the premature-optimization/smarter-but-stupid trap :D_

sboisvert:
Given your last comment I think there might of been a misunderstanding.

>This shouldn't be required, as CAP can add as coauthors even simple WP users - so what I am doing is just creating the relevant wp_term for them with update_author_term(). Is that right?

That's what causes the performance issues and what `add_filter( 'coauthors_plus_should_query_post_author', '__return_false' );` fixes.

If your code is run, can we use the add_filter( 'coauthors_plus_should_query_post_author', '__return_false' ); and ensure that folks who were co-authors (or just regular authors) actually get pulled up in those queries?

If not, then I suspect that this method would only partly fix the problem. Or I guess it fixes the search problem but doesn't fix the larger performance problems.

TheCrowned:
I updated my implementation to use `CoAuthors_Guest_Authors->create_guest_author_from_user_id()` instead of `CoAuthors_Plus->update_author_term()`. This makes sure the result of the import on install is the same as running `wp co-authors-plus create-guest-authors`.

TheCrowned:
(Hold this back until tomorrow as the same changes done on install need to be done on user update)

php/coauthors-install.php
sboisvert: >replacing time() with time() + 90

I think that would be the best bet ya, even just 15 seconds should be fine. I'd venture a guess that most installs are running innodb at this point so the write locks aren't as big a deal as it would have been before. Just giving it a bit of breathing room so it's not trying to do a million writes in a row if there are 1 million users on the site.

TheCrowned: Implemented in #587.

philipjohn:
@TheCrowned it looks like this PR before the tests - could you take a look at the commits and recreate this PR with the necessary changes please? Thanks!

philipjohn:
Umm or not....

PR #547, opened 2018-06-18
Add post type parameter to the Mine link, so it works for Pages

pdemarte:
@rebeccahum ... to replace GH-493.


rebeccahum:
Thanks for your contribution, @pdemarte!

PR #548, opened 2018-06-19
Add missing post count in  Mine links. 

pdemarte:
@rebeccahum Here it is!

If the post count is zero, the "Mine" link  is not displayed, similar to the behavior of Pending and Drafts links.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -89,7 +89,7 @@ class CoAuthors_Plus {
 		// Action to reassign posts when a guest author is deleted
 		add_action( 'delete_user',  array( $this, 'delete_user_action' ) );
 
-		add_filter( 'get_usernumposts', array( $this, 'filter_count_user_posts' ), 10, 2 );
+		add_filter( 'get_usernumposts', array( $this, 'filter_count_user_posts' ), 10, 3 );
 
 		// Action to set up co-author auto-suggest
 		add_action( 'wp_ajax_coauthors_ajax_suggest', array( $this, 'ajax_suggest' ) );
@@ -501,7 +501,7 @@ class CoAuthors_Plus {
 			return $value;
 		}
 		// We filter count_user_posts() so it provides an accurate number
-		$numposts = count_user_posts( $user_id );
+		$numposts = count_user_posts( $user_id, get_current_screen()->post_type );
 		$user = get_user_by( 'id', $user_id );
 		if ( $numposts > 0 ) {
 			$value .= "<a href='edit.php?author_name=$user->user_nicename' title='" . esc_attr__( 'View posts by this author', 'co-authors-plus' ) . "' class='edit'>";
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -994,24 +994,40 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
 	/**
 	 * Filter the count_users_posts() core function to include our correct count
 	 */
-	function filter_count_user_posts( $count, $user_id ) {
+	function filter_count_user_posts( $count = NULL, $user_id, $post_type = NULL ) {
```
TheCrowned: Is there a reason for marking the first and third arguments as optional? They come from WP Core, so shouldn't they be always set? :)

co-authors-plus.php
```diff
@@ -994,24 +994,40 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
 		$term = $this->get_author_term( $user );
 		$guest_term = get_term_by( 'slug', 'cap-' . $user->user_nicename, $this->coauthor_taxonomy );
-		// Only modify the count if it has a linked account with posts or the author exists as a term
-		if ( $user->linked_account && $guest_term->count ) {
-			if ( $term && ! is_wp_error( $term )) {
-				$count = $guest_term->count + $term->count;
-			} else {
-				$count = $guest_term->count;
+
+		if ($post_type == 'post' || empty($post_type)) {
```
TheCrowned: I am not sure I fully understand this check, also because I don't see `$post_type` being used anywhere in this block? Could you explain what your thoughts are, and why the custom query in the `else` block is needed? :)

co-authors-plus.php
```diff
@@ -994,24 +994,40 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
+					$count = $guest_term->count;
+				}
+			} elseif ( $term && ! is_wp_error( $term ) ) {
+				$count = $term->count;
 			}
-		} elseif ( $term && ! is_wp_error( $term ) ) {
-			$count = $term->count;
 		}
-
+		else {
```
TheCrowned: Trivial thing here, but could you please change it to `} else {` so that it matches the style of the whole file? :)

TheCrowned:
Hi @pdemarte, thanks for your contribution, really appreciated!

I have tested this code in my local install, and it looks like it's not working as expected. As you can see from the screenshot below, the `admin` account has lots of post 9000+, but the `Mine` count only shows 2. Without this PR code, the count is (I don't know if correct) is a number above 9000.

screenshot from 2018-06-20 22-25-13

I believe it has something to do with the fact that you remove the existing view, but you never keep track of the already counted posts. We should probably incorporate that in `filter_count_user_posts()`, which is only considering guest posts/posts whose author is a CAP user assigned to a WP user.

I'd encourage you to test it out once again and make sure it all works as expected, even in different conditions maybe. Please also have a look at the inline comments I have left to the code :)

pdemarte:
Hi! Crazy week and I haven’t had much time to review these yet. Also my client’s circumstances have changed, so I’m not sure I’ll be able to continue developing this, but I’ll try to look and respond this weekend.

TheCrowned:
No worries! :)

pdemarte:
Closing this but tried again one last time in #557

PR #557, opened 2018-07-08
Make Mine count accurately reflect the number of posts displayed in t…

pdemarte:
Hi @rebeccahum and @TheCrowned,
Closing down #548 and created this new branch to test out what I hope is a simpler solution to the post count, incorporating some of the feedback from that PR.

I tested this with both a fresh install and an existing install where they were using the plugin for quite a while (which had around 900 posts with co-authors). In my tests, it worked as intended.

I think this has to be my last contribution, because of our circumstances with this client are changing soon, but see if this might work. I won't be able to revisit, but if it at least gets somebody closer, feel free to use as a starting point for modifications. 

All the best!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
     include:
         # aliased to a recent 5.6.x version
         - php: "5.6"
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '5.6'
           env:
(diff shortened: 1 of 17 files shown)


TheCrowned:
Hi @pdemarte,
thanks again for your re-iterated effort on this! :)

I see changes for two issues here: adding the count to the Mine view, and making the output of `count_user_posts()` more accurate. Since these are two separate issues, I created a new PR with your code for adding the Mine view. I will have a more in-depth look to the other issue, as we would ideally use as few `join`s as possible, since they can slow things down significantly.

GaryJones:
As this PR contains is conflicted and old, I'm going to close it out. Feel free to open a new targeted PR if the change is still desired.

PR #559, opened 2018-07-09
Added posts count to Mine view in edit page

### LINKED ISSUE #536: Missing post count in "Mine" view for coauthor
If an author is (only) coauthor of some posts, CAP makes sure to add the `Mine` view filter link at the top of the edit.php page. It works, but lacks the posts count that all other views have.

cap missing count on mine

Head to `filter_views()` to fix.

TheCrowned:
This fixes #536, unpacking the code of #557 and making it obsolete.

I have added the posts count to the `Mine` view in the edit page. #557 had some more code to make the count more accurate, but since it acted on `filter_count_user_posts()`, I left it out from this PR.


co-authors-plus.php
```diff
@@ -1278,8 +1278,10 @@ function load_edit() {
 	 * @since 3.0
 	 */
 	function filter_views( $views ) {
+		$post_type = get_current_screen()->post_type;
```
rebeccahum: Since we're getting the post type already here, I think we can use the var `$mine_args['post_type']` from line 1294 and then, get rid of that entire conditional block in line 1293...so that we don't have to grab the post type twice.

co-authors-plus.php
```diff
@@ -1278,8 +1278,10 @@ function load_edit() {
...
+		$numposts = count_user_posts( wp_get_current_user()->ID, $post_type);
```
rebeccahum: We'll have to update `$post_type` to `$mine_args['post_type']`.  

I'm also seeing the count off, as it does not include the user's posts before a guest author account is linked.  Only the count of the posts assigned after a guest author is linked to the user is accounted for.

co-authors-plus.php
```diff
@@ -1289,14 +1291,16 @@ function filter_views( $views ) {
 				'author_name'           => wp_get_current_user()->user_nicename,
 			);
 		if ( 'post' != get_post_type() ) {
-			$mine_args['post_type'] = get_current_screen()->post_type;
+			$mine_args['post_type'] = get_post_type();
```
rebeccahum: We can probably get rid of this entire conditional block if we are getting the post type up as per ^^.

co-authors-plus.php
```diff
@@ -1289,14 +1291,16 @@ function filter_views( $views ) {
...
 		}
 		if ( ! empty( $_REQUEST['author_name'] ) && wp_get_current_user()->user_nicename == $_REQUEST['author_name'] ) {
 			$class = ' class="current"';
 		} else {
 			$class = '';
 		}
-		$views['mine'] = $view_mine = '<a' . $class . ' href="' . esc_url( add_query_arg( array_map( 'rawurlencode', $mine_args ), admin_url( 'edit.php' ) ) ) . '">' . __( 'Mine', 'co-authors-plus' ) . '</a>';
+
+		$count = sprintf(' <span class="count">(%s)</span>', number_format_i18n( $numposts ) );
+		$views['mine'] = $view_mine = '<a' . $class . ' href="' . esc_url( add_query_arg( array_map( 'rawurlencode', $mine_args ), admin_url( 'edit.php' ) ) ) . '">' . __( 'Mine' ) . $count . '</a>';
```
rebeccahum: Is there a reason why the `co-authors-plus` string was removed from the domain parameter of `__( 'Mine', 'co-authors-plus' )`?

rebeccahum (changes requested):

co-authors-plus.php
TheCrowned: Yes - it seemed unnecessary to me. `Mine` is also a string from WP Core, and it made sense to me to use its translation. This would avoid the (possible) issue of someone translating CAP's `Mine` in a different way from Core, and end up with the string being displayed sometimes in one way, sometimes another :)

TheCrowned:
Thanks for the review Rebecca :) Went through your comments - for the ones related to code improvements, please see my other PR #560. I thought it was better to separate the two things :)

co-authors-plus.php
rebeccahum: Fair enough, carry on!

co-authors-plus.php
rebeccahum: The count is being pulled from `$numposts` and printed in line 1303, so it appears to be related.

rebeccahum (changes requested):

GaryJones:
The conflicted nature and need of a rebase makes the relevant changes here hard to spot, though I have noticed that the count seems to be missing. I will close it out, but keep the related issue open.

GaryJones:
[Last three commits](https://github.com/Automattic/Co-Authors-Plus/pull/559/files/604b3f028406171002de39d2bcca6ff4a2c12249..eca6c642eb2de236cc27ce35a1459445971bfa5b) are relevant.

PR #560, opened 2018-07-09
Improved filter_views() function code

TheCrowned:
This does not affect functionality, it only improves the code of the `filter_views()` function. Please review #559 first, as this incorporates those changes already.

Two kind of changes:

1. Coding Standards: whitespace changes, using `===` instead of `==`.
2. Improving performance and _flow_: 

- avoid a `str_replace` when not needed, 
- re-create the `Mine` view even if it exists already, since we have to append our custom count. It's easier to do so than do parse the already-present link and put the count inside it.
- the only more difficult to understand change could be this:

```php
if ( 'post' != get_post_type() ) {
	$mine_args['post_type'] = get_current_screen()->post_type;
}
```

This arg is later used to build the `Mine` view link. Here, the error is _philosophical_ - although it would have never generated issues, I believe it was wrong. `get_post_type()` and `get_current_screen()->post_type` are not warranted to always return the same value, since they rely on different globals. In fact, `get_post_type()` relies on the loop to get the post type of the current post, while `get_current_screen()->post_type` takes the post type of the current edit screen. These would ideally always match, but there is no reason to mix them and it can be confusing. There were also several calls to query the same information multiple times, so one further reason to change this.

That check was to avoid appending `?post_type` to the `edit.php` querystring when the value would be `post`, relying on the default fallback behavior, but since Core appends it even in that case, no reason to do it differently :)


GaryJones:
Adding to a future milestone to review the value of the last two commits.

PR #563, opened 2018-07-16
Changed posts_selection from filter to action

### LINKED ISSUE #562: `posts_selection` is an action, not a filter
The source code has the following filter registration:
```php
add_filter( 'posts_selection', array( $this, 'fix_author_page' ) ); // use posts_selection since it's after WP_Query has built the request and before it's queried any posts
```

However, `posts_selection` [is an action, not a filter](https://developer.wordpress.org/reference/hooks/posts_selection/), so this line presently does nothing.

TheCrowned:
This fixes #562.

The issue was that `posts_selection` was being used as a WP filter rather than an action, as it should have been. This was wrong, but did not impact the plugin behavior because `add_action()` is actually [just a wrapper](https://developer.wordpress.org/reference/functions/add_action/#source) of `add_filter()`. The other way around, it would have not worked, but this way... it works, even if it is wrong.

I have also taken the chance to add the apt argument to the callback (since the hook is passing it) and to update the inline doc.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -110,7 +110,7 @@ class CoAuthors_Plus {
 		add_filter( 'wp_get_object_terms', array( $this, 'filter_wp_get_object_terms' ), 10, 4 );
 
 		// Make sure we've correctly set data on guest author pages
-		add_filter( 'posts_selection', array( $this, 'fix_author_page' ) ); // use posts_selection since it's after WP_Query has built the request and before it's queried any posts
+		add_action( 'posts_selection', array( $this, 'fix_author_page' ) ); // use posts_selection since it's after WP_Query has built the request and before it's queried any posts
 		add_action( 'the_post', array( $this, 'fix_author_page' ) );
 
 		// Support for Edit Flow's calendar and story budget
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -1071,8 +1071,11 @@ function current_user_can_set_authors( $post = null ) {
 	 * the query_var is changed.
 	 *
 	 * Also, we have to do some hacky WP_Query modification for guest authors
+	 *
+	 * @param string $selection The assembled selection query
```
sboisvert: It might be useful to add a @uses http://docs.phpdoc.org/references/phpdoc/tags/uses.html to point to how this modifies the global WP_Query object.

sboisvert:
Why are the checks for 1) Test_CoAuthors_Plus::test_search_authors_when_ignored_authors_provided failing?

TheCrowned:
I don't know, but they fail in master as well, so it's not related to this PR, probably to some of the bugs about search being fixed. I'll have a look in another PR :)

co-authors-plus.php
TheCrowned: You mean something like
`@uses Changes $wp_query->queried_object, $wp_query->queried_object_id, $wp_query->is_author, $wp_query->is_archive, $wp_query->is_404.`
?

PR #565, opened 2018-07-16
Avoid partial author ID replacement in `posts_where_filter`

jeremyfelt:
If a user with the ID of 1 visited the author archive page for a user with the ID of 17, the regex would filter out the `1` of `17` and replace with the corresponding guest author ID resulting in a SQL string like `'2039'7`. This results in invalid SQL.

This PR adds a trailing space to the end of the current user query and requires that space in the `preg_replace()`. There's probably another way to do this by forcing the match to occur at the end of the string, but this is what I've got. :)

Query text generated before (bad):

```
AND ((wp_44_term_taxonomy.taxonomy = 'author' AND wp_44_term_taxonomy.term_id = '2039'7 OR (wp_44_term_taxonomy.taxonomy = 'author' AND wp_44_term_taxonomy.term_id = '1916'))) AND wp_44_posts.post_type = 'post' AND (wp_44_posts.post_status = 'publish' OR wp_44_posts.post_status = 'private')"
```

Query text generated after (good):

```
AND ((wp_44_posts.post_author = 17 OR (wp_44_term_taxonomy.taxonomy = 'author' AND wp_44_term_taxonomy.term_id = '1916'))) AND wp_44_posts.post_type = 'post' AND (wp_44_posts.post_status = 'publish' OR wp_44_posts.post_status = 'private')
```

I think this was introduced in 08c0315.


jeremyfelt:
Closed this because I deleted the fork. Looks like the commit is still visible if needed.

Issue #566, opened 2018-07-17
Migrate author terms from `$username` to `cap-$username`


TheCrowned:
Since the non-prefixed `$username` term is outdated (see [`update_author_term`](https://github.com/TheCrowned/Co-Authors-Plus/blob/master/co-authors-plus.php#L1452), which, when inserting, uses the prefixed `cap-`version) I think we should migrate terms to all having the prefix. We could run a `wp_update_term()` on `update_author_term()` if needed and that should settle it.

TheCrowned:
Quick note to say that there is already a CLI command `migrate_author_terms` to do so. Still nice to have it done on old installs on the fly as outlined above.

GaryJones:
I think the number of old installs would now be insignificant.

Issue #460, opened 2017-10-30
Inaccurate Linked Guest Author post count if Guest Author made post prior to Linking


rebeccahum:
1. Create user account named "Editor" with user role of "Editor"
2. Create guest author account named "Guests"
3. Create post by "Guests".
4. Link user account "Editor" to "Guests".
5. Go to Users > Guest Accounts and you will see that there is an inaccurate post count.

It should count both the posts from the guest author and linked account for accuracy.

sboisvert:
Closed in #461

rebeccahum:
Re-opening as this has been merged over in https://github.com/Automattic/Co-Authors-Plus/commit/ecf869994b79e2e90976aa293bc3f7a861509358 so the issue still remains.

nielslange:
**Detailed steps to reproduce:**

1. Create an editor account
2. Create two posts using the editor account
3. Create a guest author account
4. Create one post using the guest author account
5. Link the guest author account to the editor account

**Expected behaviour**

- Editor post count: 3 → ✅
- Guest author post count: 3 → ✅

**Current behaviour**

- Editor post count: 3 → ✅
- Guest author post count: 5 → ❌

htdat:
I am not sure if this makes sense but here what I could find out so far. 

Start at these lines in getting post counts for co-authors:

https://github.com/Automattic/Co-Authors-Plus/blob/c9ebbd1e02741743e109e8c3f45cc38d62d25e4d/php/class-coauthors-wp-list-table.php#L271-L284

As I understand, the `if` logic here means if this is a linked account, it will make the sum of two parts:

*   number of posts from the co-author. (1)
*   number of posts from the linked author (real user in WP system). (2)

But part (1) actually has already covered part (2) in this case as we can see in the method `CoAuthors_Plus::get_guest_author_post_count` [(code ref)](https://github.com/Automattic/Co-Authors-Plus/blob/c9ebbd1e02741743e109e8c3f45cc38d62d25e4d/co-authors-plus.php#L1729-L1729
).

*   It checks if this user is a linked account, then it will return `count_user_posts( $user->ID )`;
*   This function `count_user_posts` actually applies this filter `get_usernumposts`. ([code ref](https://href.li/?https://github.com/WordPress/WordPress/blob/133c4e2e509845e29ab7b25671bea2cc7cbf7f52/wp-includes/user.php#L569))
[...]

PR #567, opened 2018-07-17
Fixing double posts counts for users with linked accounts

### LINKED ISSUE #561: Wrong post count for guest authors with terms
Calling `count_user_posts()` with CAP enabled leads to wrong counts for users that have a guest author profile linked to them. 

This is an example with CAP enabled - clearly off since the `Mine` posts is more than the total:
w cap

With CAP disabled:
wo cap

The issue lies in `filter_count_user_posts()`, in particular these two lines:
```php
$guest_term = get_term_by( 'slug', 'cap-' . $user->user_nicename, $this->coauthor_taxonomy );
```
the issue being that `get_author_term()` will already try to retrieve the same term `$guest_term` holds, if available. In this situation, `$term` and `$guest_term` contain the same value and thus the post count gets doubled.

This would fix (at least partly, not sure if wholly) fix the issue addressed in PR #557, without using any `JOIN`s.

TheCrowned:
This fixes #561.

On a correct, up-to-date CAP installation, posts counts for authors with guest authors profile are absolutely off - basically, they are double what they should.

This is an example with CAP enabled - clearly off since the `Mine` posts is more than the total:
w cap

With CAP disabled:
wo cap

Prior to this PR, the code which filtered the WP posts count was this:
```php
$guest_term = get_term_by( 'slug', 'cap-' . $user->user_nicename, $this->coauthor_taxonomy );

// Only modify the count if it has a linked account with posts or the author exists as a term
if ( $user->linked_account && $guest_term->count ) {
	if ( $term && ! is_wp_error( $term )) {
		$count = $guest_term->count + $term->count;
	} else {
		$count = $guest_term->count;
	}
} elseif ( $term && ! is_wp_error( $term ) ) {
	$count = $term->count;
}
```
Both `$term` and `$guest_term` are considered, guessing that `$term` would hold the non-prefixed version of the CAP author term, while `$guest_term` the prefixed `cap-` one. The counts are later summed under the assumption that they are complementary. The flaw is that [`$this->get_author_term()`](https://github.com/TheCrowned/Co-Authors-Plus/blob/master/co-authors-plus.php#L1424) already fetches the one term that exists, giving priority to the prefixed `cap-` version of it. So what happens is that, **if `cap-$username` exists, both `$term` and `$guest_term` will contain its content, and `$count` will then be doubled.**

The fix is thus to get rid of the `$guest_term` var and only retain the final conditional:
```php
if ( $term && ! is_wp_error( $term ) ) {
	$count = $term->count;
}
```
But aren't we losing some counts on the way? Aren't we discarding the non-prefixed version of the term? Is it just useless? No, we don't lose counts; No, we are discarding it, and no, it is not useless. The pivot is [`$this->update_author_term()`](https://github.com/TheCrowned/Co-Authors-Plus/blob/master/co-authors-plus.php#L1452): it will update any of the terms that exist, also suggesting that it should never happen that they co-exist: either there is the prefixed `cap-` term, _or_ the non-prefixed one. In any case, `get_author_term()` will fetch the one that exists for the author, which will already contain the total count, and we don't need to sum anything to it. 

(I think we should also think about migrating terms to all having the `cap-` prefix, but this is another issue #566)

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -994,23 +994,19 @@ class CoAuthors_Plus {
 	}
 
 	/**
-	 * Filter the count_users_posts() core function to include our correct count
+	 * Filter the count_users_posts() core function to include our correct count.
+	 *
+	 * @param int $count Post count
+	 * @param int $user_id WP user ID
+	 * @return int Post count
 	 */
 	function filter_count_user_posts( $count, $user_id ) {
 		$user = get_userdata( $user_id );
-
 		$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
 
 		$term = $this->get_author_term( $user );
-		$guest_term = get_term_by( 'slug', 'cap-' . $user->user_nicename, $this->coauthor_taxonomy );
-		// Only modify the count if it has a linked account with posts or the author exists as a term
-		if ( $user->linked_account && $guest_term->count ) {
-			if ( $term && ! is_wp_error( $term )) {
-				$count = $guest_term->count + $term->count;
-			} else {
-				$count = $guest_term->count;
-			}
-		} elseif ( $term && ! is_wp_error( $term ) ) {
+		
+		if ( $term && ! is_wp_error( $term ) ) {
 			$count = $term->count;
 		}
 


rebeccahum:
How does this affect the count for guest authors with linked accounts?

sboisvert:
This is an awesome Fix and an awesome explanation of the fix!

rebeccahum:
Just a follow-up on this, since it merges over the fixes in PR #461, the edge case of what is described in #460 is present again.

Issue #553, opened 2018-07-04
WP_CLI create-guest-authors creates profiles for ALL users


TheCrowned:
`wp co-authors-plus create-guest-authors` will create guest authors profiles for all WP users, including subscribers, which we wouldn't want to ever add as coauthors, and uselessly pollutes the database.

sboisvert:
@smistephen Do you want to take a look?

smistephen:
The offending line of code seems to be https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-wp-cli.php#L28, which just plucks out `get_users()` without any qualifiers.

My proposed solution is to simply whitelist the roles that get profiles created for them.

We could do this via changing the args to `get_users()` to `get_users( array( 'role__in" => array( 'administrator', 'editor', 'author' ) ) )`, but that might get kind of slow with a bunch of LIKE queries on meta values:

```
SELECT wp_users.* FROM wp_users INNER JOIN wp_usermeta ON ( wp_users.ID = wp_usermeta.user_id ) WHERE 1=1 AND ( 
  ( 
    ( wp_usermeta.meta_key = 'wp_capabilities' AND wp_usermeta.meta_value LIKE '%\"administrator\"%' ) 
    OR 
    ( wp_usermeta.meta_key = 'wp_capabilities' AND wp_usermeta.meta_value LIKE '%\"editor\"%' ) 
    OR 
    ( wp_usermeta.meta_key = 'wp_capabilities' AND wp_usermeta.meta_value LIKE '%\"author\"%' )
  )
) ORDER BY user_login ASC
```

We could also do the processing on the back-end, like this:

```
$users_unfiltered = get_users();
$users            = array();

// Roles to generate guest authors for.
$role_whitelist = array( 'administrator', 'editor', 'author' );

foreach ( $users_unfiltered as $user ) {
	foreach ( $user->roles as $role ) {
		if ( in_array( $role, $role_whitelist, true ) ) {
			// Using the user ID ensures that users that would otherwise
			// be in the list twice overwrite each other.
			$users[ $user->ID ] = $user;
		}
	}
}

// Discard user ID indices, rebase to 0.
$users = array_values( $users );
```

Tomorrow, I'll generate a bunch of users and pit the two against each other.

I'll probably just write a bash script to call `wp user create` 1000 times, if anyone has a better idea I'm open to it.

smistephen:
[...]

PR #575, opened 2018-07-27
Added role whitelist to internals of wp-cli guest author generation command.

smistephen:
Previously, as noted in https://github.com/Automattic/Co-Authors-Plus/issues/553, `wp co-authors-plus create-guest-authors` created a guest author for every user in the database.

This was a massive performance suck, as well as being largely unnecessary. Adding a subscriber as a guest author is not a common use case.

Now, only users having the role of administrator, editor, and author have guest authors generated for them.

The query to filter users by role is a touch heavy (it performs LIKEs on meta values), but testing it on a site with 1000 users showed no major issues.

It was also compared to simply querying all the users, and doing the filtering in PHP. Head to head, there was no significant difference in performance between the two approaches, and using the WordPress API is always a better bet than writing your own one-off custom filter algorithm.


php/class-wp-cli.php
```diff
@@ -25,7 +25,7 @@ public function create_guest_authors( $args, $assoc_args ) {
 			);
 		$this->args = wp_parse_args( $assoc_args, $defaults );
 
-		$users = get_users();
+		$users = get_users( array( 'role__in' => array( 'administrator', 'editor', 'author' ) ) );
```
TheCrowned: I think we may want to allow filtering of this in some way. Some sites may have other roles they want to be allowed to be added as coauthors. Since it's a CLI command we may want to allow an optional argument to overwrite the default list of roles :)

Also, I think the `contributor` role should be included by default, as it has the `edit_posts` cap and thus permission to be added as coauthor :)

smistephen: Works for me. After lunch I'll put together some thoughts on a --roles flag :)

php/class-wp-cli.php
```diff
@@ -21,11 +48,77 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+		foreach ( $role_whitelist as $role ) {
+			// Does this role exist in this instance?
+			if ( ! $GLOBALS['wp_roles']->is_role( $role ) ) {
+				WP_CLI::error( __( 'Role ', 'co-authors-plus' ) . $role . __( ' does not exist.', 'co-authors-plus' ) );
+			}
+
+			// Subscribers cannot be made guest authors en masse
+			// due to potential performance issues (sites can have millions of subscribers).
+			if ( 'subscriber' === $role ) {
+				WP_CLI::error( __( 'For performance reasons, subscribers cannot be made guest authors via wp-cli. Please use the admin interface.', 'co-authors-plus' ) );
```
sboisvert: Maybe we should have a "force" option?

php/class-wp-cli.php
```diff
@@ -21,11 +48,77 @@ public function create_guest_authors( $args, $assoc_args ) {
 		global $coauthors_plus;
 
 		$defaults = array(
-				// There are no arguments at this time
-			);
+			'roles' => array( 'administrator', 'editor', 'author', 'contributor' ),
+			'batch-size' => 1000,
```
sboisvert: Why would folks need to specify this? What would be the difference if we were to limit it to batches of 100 at a time?

php/class-wp-cli.php
```diff
@@ -21,11 +48,77 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+		if ( $this->args['batch-size'] > 0 && is_numeric( $this->args['batch-size'] ) ) {
+			$batch_size = (int) $this->args['batch-size'];
+		} else {
+			WP_CLI::error( __( 'Batch size must be an integer greater than zero.', 'co-authors-plus' ) );
+		}
+
+		foreach ( $role_whitelist as $role ) {
+			// Does this role exist in this instance?
+			if ( ! $GLOBALS['wp_roles']->is_role( $role ) ) {
+				WP_CLI::error( __( 'Role ', 'co-authors-plus' ) . $role . __( ' does not exist.', 'co-authors-plus' ) );
```
sboisvert: Should we instead inject $role as a variable inside a singular translation? That would seem to be best for many languages where the sentence structure may be different.

php/class-wp-cli.php
```diff
@@ -21,11 +48,77 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+			}
+
+			// Subscribers cannot be made guest authors en masse
+			// due to potential performance issues (sites can have millions of subscribers).
+			if ( 'subscriber' === $role ) {
+				WP_CLI::error( __( 'For performance reasons, subscribers cannot be made guest authors via wp-cli. Please use the admin interface.', 'co-authors-plus' ) );
+			}
+		}
+
+		$users_unfiltered_query = new WP_User_Query( array( 'number' => $batch_size ) );
```
sboisvert: Do we get any benefits if we order by primary key?

sboisvert (changes requested):
I felt a few comments inline.
There doesn't seem to be a dry run option and there doesn't seem to be a way to capture the log of what was changed. Could we add those?

php/class-wp-cli.php
smistephen: Not really, we're going to iterate over the entire array no matter what, so the order isn't important :)

php/class-wp-cli.php
smistephen: Sure, I can add that

php/class-wp-cli.php
smistephen: :+1:

php/class-wp-cli.php
smistephen: It's to stop the script from fetching **ALL** of the users at once and crashing servers with low memory limits, while also allowing servers with greater resources to open up the throttle.

smistephen: If we were to limit it to batches of 100 at a time, it would take longer but use less memory (a tradeoff I've had to make many times in the past running on low-end equipment, haha)

php/class-wp-cli.php
sboisvert: If we order by primary key, would it make the lookups and offsetting be faster?

php/class-wp-cli.php
```diff
@@ -21,28 +67,149 @@ public function create_guest_authors( $args, $assoc_args ) {
...
-		$users = get_users();
-		$created = 0;
-		$skipped = 0;
-		$progress = \WP_CLI\Utils\make_progress_bar( 'Processing guest authors...', count ( $users ) );
-		foreach ( $users as $user ) {
+		if ( is_array( $this->args['roles'] ) ) {
+			$role_whitelist = $this->args['roles'];
+		} else {
+			$role_whitelist = explode( ',', $this->args['roles'] );
+		}
```
TheCrowned: This is really just minor and can be ignored, but I still believe worth pointing out. Cleaner:
```
$role_whitelist = $this->args['roles'];
if ( false === is_array( $role_whitelist ) ) {
	$role_whitelist = explode( ',', $role_whitelist );
}
```

TheCrowned:
Hey! We may want to improve the inline documentation of the function to comply with WP_CLI PHPDoc standards. Resources that have been very helpful for me when I dealt with similar issue have been this [Commands Cookbook - Annotating with PHPDoc](https://make.wordpress.org/cli/handbook/commands-cookbook/#annotating-with-phpdoc) and [How to write custom WP_CLI commands for WordPress automation](https://firxworx.com/blog/wordpress/how-to-write-custom-wp-cli-commands-for-wordpress-automation/) (scroll to _PHPDoc comments and controls_ section). 

For example, we'd want to split the description in short and long; all optional args should be wrapped in `<` `>`; add a `synopsis`; etc.

Also, boolean options shouldn't be treated as all others. For example, specifying `[--<dry-run>]` will automatically force it to be a boolean arg, and will allow to disable with `--no-dry-run`. This way, you should also manage to get rid of all the `is_bool()` checks, and any boolean option can be turned off with a `no-` in front of it :) This way everything is simpler and we avoid having a `not-a-dry-run` => false` by default, which is a bit confusing.

On a different point, I don't think that the dry run mode should be the default one - let developers toggle it if they want :) (that is also as other CLI commands work)

Finally, if I understand @sboisvert intentions correctly, he meant we _should_ log the routine progress, but not necessarily give an option to do so. The logging should happen by default. Then, if someone doesn't want it, they can redirect it to `/dev/null`. Adding an option for that would really just be adding another way to write `> /dev/null`, and I don't think it's necesary :)

(I also left inline comments :))

php/class-wp-cli.php
```diff
@@ -21,28 +67,149 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+		} else {
+			$created = 0;
+			$skipped = 0;
+			$progress = \WP_CLI\Utils\make_progress_bar( 'Processing guest authors...', count ( $users ) );
+			foreach ( $users as $user ) {
+				$result = $coauthors_plus->guest_authors->create_guest_author_from_user_id( $user->ID );
+				if ( is_wp_error( $result ) ) {
+					$skipped++;
+					if ( true === $log_output ) {
+						foreach ( $result->get_error_messages() as $error_message ) {
```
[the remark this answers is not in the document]
smistephen: This is just me programming defensively.

`WP_Error->get_error_messages()` (https://developer.wordpress.org/reference/classes/wp_error/) returns an array of error messages even if there's just one.

I could use `WP_Error->get_error_message()`, which returns an int or string, but that makes me incredibly nervous, as it silently suppresses all other error messages even if they exist.

So yes, currently `create_guest_author_from_user_id()` only ever returns one error message. But I can't predict that it always will, so I coded it this way :)

php/class-wp-cli.php
smistephen: Turns out yes, yes it would: https://github.com/smistephen/Co-Authors-Plus/commit/7231625d3d42c3b3680c3ba036baa430944a4f64

php/class-wp-cli.php
```diff
@@ -21,28 +67,149 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+					/* translators: %s: User role */
+					__( 'Role %s does not exist.', 'co-authors-plus' ),
+					$role
+				) );
+			}
+
+			// Subscribers cannot be made guest authors en masse
+			// due to potential performance issues (sites can have millions of subscribers).
+			if ( 'subscriber' === $role && false === $force_subscribers ) {
+				WP_CLI::error( __( 'For performance reasons, subscribers cannot be made guest authors via wp-cli. Please use the admin interface.', 'co-authors-plus' ) );
```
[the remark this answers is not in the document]
smistephen: Done :+1:

php/class-wp-cli.php
```diff
@@ -21,28 +67,149 @@ public function create_guest_authors( $args, $assoc_args ) {
...
-			$result = $coauthors_plus->guest_authors->create_guest_author_from_user_id( $user->ID );
-			if ( is_wp_error( $result ) ) {
-				$skipped++;
-			} else {
-				$created++;
+		if ( $this->args['batch-size'] > 0 && is_numeric( $this->args['batch-size'] ) ) {
+			$batch_size = (int) $this->args['batch-size'];
+		} else {
+			WP_CLI::error( __( 'Batch size must be an integer greater than zero.', 'co-authors-plus' ) );
+		}
```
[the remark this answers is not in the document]
smistephen: Done :+1:

php/class-wp-cli.php
smistephen: Done :+1:

php/class-wp-cli.php
```diff
@@ -11,38 +11,225 @@
...
+		$batch_size = (int) $this->args['batch-size'];
+		if ( $batch_size <= 0 ) {
+			WP_CLI::error( __( 'Batch size must be an integer greater than zero.', 'co-authors-plus' ) );
+		}
+
+		if ( true === $this->args['force-subscribers'] ) {
+			$force_subscribers = true;
+		} else {
+			$force_subscribers = false;
+		}
```
TheCrowned: At this point we can just have a `$force_subscribers = $this->args['force_subscribers'];` , and the same applies to the other bool options :)

php/class-wp-cli.php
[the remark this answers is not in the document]
TheCrowned: Yes, it is possible. I was skeptical just because I have never seen that in any plugin/core piece I ever looked at, but yes :)

TheCrowned (commented):
One final thing: I'm not sure we want to include the command expected output in the example lines? I think it's enough to have a (brief) description of what it does, and then the command, without the related output :)

smistephen:
The output is in keeping with documentation specification.

GaryJones:
The key changes are going to need to be manually picked out of this, but there's value to be had. Adding to a future milestone.

GaryJones:
Thanks for this, @smistephen, and apologies for the long delay.

We're closing this out rather than merging. As with the companion delete-guest-authors PR, after eight years of `develop` drift the branch is conflicting across 19 files — as noted on this PR previously, the key changes would need to be manually picked out, so this is effectively a fresh implementation rather than a rebase.

The underlying need is still valid, so the tracking issue (#553) stays open as the home for it. Your role-whitelist logic here remains a useful reference for whoever picks it up — this PR is linked from #553 so the code stays one click away.

Closing in favour of #553.

Issue #537, opened 2018-06-07
Guest Authors mapping to WordPress account not working


mikeavena:
I have multiple guest authors that also have WordPress accounts on my site. Guest authors that do have WordPress accounts are linked via the co-authors plus user mapping tool. The issue is the posts attributed to guest authors do not appear in listings for the WordPress account. Clicking on the link for the author in the WordPress admin panel says there are no posts found for the author. It seems as if the mapping is completely ignored.

TheCrowned:
Has to do with #555.

psaikali:
FYI, https://github.com/Automattic/Co-Authors-Plus/pull/576 will fix this.

https://github.com/Automattic/Co-Authors-Plus/issues/508 for more technical details.

Issue #508, opened 2018-04-23
Plugin Preventing Correct Listing of Pages When Selecting "Mine" Link


catchmyfame:
The issue is that when you visit the Pages section of WP and click the "Mine" link, the full list of pages belonging to the logged in user isn't shown. From what I'm seeing, only pages marked as Private are being listed. However, if you instead select the "All" link, and the pick the same user from the "All users" dropdown menu and then click the Filter button, the proper listing is displayed.

The "Mine" link is something like 

`example.com/wp-admin/edit.php?post_type=page&author=25` or `example.com/wp-admin/edit.php?post_type=page`

And is the same regardless of whether the CAP is enabled or disabled, so something in the plugin is preventing the proper listing. 

When I disable the CAP plugin this issue goes away, and when I enable it it returns. I have disabled all other plugins and confirmed that it's CAP that is causing the issue.

catchmyfame:
So far I've tracked the problem down to being in co-authors-plus.php

catchmyfame:
Upon further investigation it appears as if the change in SQL in the `posts_join_filter` function in  co-authors-plus.php file is causing the issue. The difference in the rendered code:

**NEW** (buggy):

```
LEFT JOIN wordpress_cms_term_relationships 
AS tr1 
ON (wordpress_cms_posts.ID = tr1.object_id) 
LEFT JOIN wordpress_cms_term_taxonomy 
ON ( tr1.term_taxonomy_id = wordpress_cms_term_taxonomy.term_taxonomy_id )
```

**OLD** (working):

```
LEFT JOIN wordpress_cms_term_relationships 
ON (wordpress_cms_posts.ID = wordpress_cms_term_relationships.object_id) 
LEFT JOIN wordpress_cms_term_relationships 
AS tr1 
ON (wordpress_cms_posts.ID = tr1.object_id) 
LEFT JOIN wordpress_cms_term_taxonomy 
ON ( tr1.term_taxonomy_id = wordpress_cms_term_taxonomy.term_taxonomy_id )
```

For testing purposes, if I replace the changed new SQL with the 3.2.2 version the problem goes away.

combs:
We are seeing this at washingtonian.com too -- would love to see a fix.

dartiss:
Heavy.com are also seeing the same issue.

They have the following filter in place for database performance issues, and it's been found that removing this temporarily fixes the user posts issue...

[...]

PR #576, opened 2018-07-27
Modified query filter regex.

smistephen:
This code is meant to append a conditional statement to the WHERE clause of a query.

But because the regex mistakenly believed that the statement should be appended after a user ID, and not after 'private', we ended up with queries like this:

```
OR wp_posts.post_author = 3
OR (
    wp_term_taxonomy.taxonomy = 'author'
    AND wp_term_taxonomy.term_id = '8'
)
AND wp_posts.post_status = 'private'
```

instead of this:

```
OR wp_posts.post_author = 3 AND wp_posts.post_status = 'private'
OR (
    wp_term_taxonomy.taxonomy = 'author'
    AND wp_term_taxonomy.term_id = '8'
)
```

Those are two very different queries: One returns posts including those private posts that belong to user 3, the other excludes all posts that are not private posts belonging to user 3.

The modified regex now allows the query to be constructed appropriately.

I've run unit tests on the code and no regressions were reported.

You can also review my process of solving the problem here: https://github.com/Automattic/Co-Authors-Plus/issues/508

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -712,7 +712,7 @@ class CoAuthors_Plus {
 
 					$maybe_both_query = $maybe_both ? '$0 OR' : '';
 
-					$where = preg_replace( '/\s\b(?:' . $wpdb->posts . '\.)?post_author\s*IN\s*(.*private\'.)/', ' (' . $maybe_both_query . ' ' . $terms_implode . ')', $where, -1 ); #' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
+					$where = preg_replace( '/\s\b(?:' . $wpdb->posts . '\.)?post_author\s*IN\s*(.*' . $id . '(?:.*private\')?.)/', ' (' . $maybe_both_query . ' ' . $terms_implode . ')', $where, -1 ); #' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
 
 				} else {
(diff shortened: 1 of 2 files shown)


smistephen:
I checked the Travis results, and the failure is due to:

```
There was 1 failure:
1) Test_Author_Queries::test__author_arg__user_is_coauthor__author_arg
Failed asserting that 0 matches expected 1.
/home/travis/build/Automattic/Co-Authors-Plus/tests/test-author-queries.php:79
```
which is currently being examined.

smistephen:
Yep, you're right, the conditionals being grouped in the query definitely helped tip me off to the fact that something was wrong.

I didn't put the full query here in the PR, because it's pretty massive, but it is available here, groups and all: https://github.com/Automattic/Co-Authors-Plus/issues/508#issuecomment-407827783

As to unit tests, `Test_Author_Queries::test__author_arg__user_is_post_author` should have caught this. The problem is that the query was run as an admin _(where the post is visible and the test passes)_ and not as the post author _(where it isn't and it doesn't)_.

I've remedied this by adding

`wp_set_current_user( $author_id );`

to the test, so the query is performed in the context of the user that created the post.

smistephen:
I'll never say no to more tests, haha

psaikali:
This fixes https://github.com/Automattic/Co-Authors-Plus/issues/537 too 👍

smistephen:
@psaikali Ooh, bonus.

jtbayly:
Can we get this merged into master?

Issue #573, opened 2018-07-20
Some author pages empty for some logged in users


jtbayly:
I first posted about this over on Wordpress.org: https://wordpress.org/support/topic/some-author-pages-empty-for-some-logged-in-users/

Now there is another user commenting with the same problem, so I thought I'd post the issue here. 

Some authors display as having no posts, but only to logged in users.

For example, while logged in, this page will display “No posts found.” to me:
https://example.com/wp-admin/edit.php?author_name=randomauthor1
And if that one doesn’t work for me, then neither does this one:
https://example.com/author/randomauthor1/

But randomauthor2 will work just fine.

However, oddly enough, logging in as a different user will change which authors don’t work and which ones do. In fact, we’ve got more than 20 authors, and all but one of the author pages work for me. But another user on the site reports that 5 of them fail to load for him.

And if you’re not logged in, the public author pages seem to work just fine (although that’s hard for me to prove with 100% certainty.)

ammist:
I'm having more site-wide issues w/ the new version of Co-authors plus.  Most author pages on the live site only show one post (even though some guest authors have hundreds).  When I'm logged in, I see zero posts, and some of the time I get taken back to the site homepage.  I've traced the issue to the SQL query that gets generated when the author page is displayed - it returns zero or one result.

Switching back to version 3.2.2 restores the author pages to their former glory.

It looks like the query filters are doings something funky (even when I've turned off all the other plugins that have a function on posts_where and posts_join filters).  I did not try to apply @codemacherUG's fixes yet.

antonlukin:
The same here

rebeccahum:
> However, oddly enough, logging in as a different user will change which authors don’t work and which ones do. In fact, we’ve got more than 20 authors, and all but one of the author pages work for me. But another user on the site reports that 5 of them fail to load for him.

Do you know what it is about the logged in user that is causing this query to fail for them?  E.g. any reproduction steps?

jtbayly:
The commit on #580 claims to fix this. I haven't tested it, but from what I've read of all the connected issues, it seems that the logged in user's ID must match the start of the other user's id. If so, a simple way to reproduce would be to create a site where you are ID '1' and then create users until you've got users with id's in the teens (eg user id '13'). From there, assign them content, and then while logged in as user 1 view the author page for user 13.

rebeccahum:
@jtbayly Thanks for the information, I'll look into this.

jtbayly:
[...]

PR #580, opened 2018-07-31
BUGFIX regular expression in posts_where_filter()

ghost:
According to issue #573 the error is, when the user_id of the currently logged in user is a prefix of the user_id queried for archive pages. This effects archive pages in frontend, when logged in and the posts archives in backend.

For example: the logged in user_id is 1 and the user_id for displaying the archive is 16

AND ((posts.post_author = 16

will be substituted to

AND ((term_taxonomy.taxonomy = 'author' AND term_taxonomy.term_id = '243' 6

This produces a sql error and the query is empty.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -731,7 +731,7 @@ class CoAuthors_Plus {
 						$current_user_query = $wpdb->term_taxonomy . '.taxonomy = \'' . $this->coauthor_taxonomy . '\' AND ' . $wpdb->term_taxonomy . '.term_id = \'' . $current_coauthor_term->term_id . '\'';
 						$this->having_terms .= ' ' . $wpdb->term_taxonomy . '.term_id = \'' . $current_coauthor_term->term_id . '\' OR ';
 
-						$where = preg_replace( '/(\b(?:' . $wpdb->posts . '\.)?post_author\s*=\s*(' . get_current_user_id() . '))/', $current_user_query, $where, - 1 ); #' . $wpdb->postmeta . '.meta_id IS NOT NULL AND}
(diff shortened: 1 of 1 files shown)


jtbayly:
Awesome! Hope to see this in the next update.

TheCrowned:
Thanks a lot for your contribution! Just wanted to leave a note to say I tested this, and works for me :)

jtbayly:
Just tested this and it works as described. Fixes issue #573 also.

Really needs to be merged, in my opinion.

rebeccahum:
I was able to reproduce, however, this didn't fix it for me.  What worked was replacing this line:

https://github.com/Automattic/Co-Authors-Plus/blob/07cbfe0eb57c00c306e3a5d441bd02a25a26dad7/co-authors-plus.php#L718

With:

`$where = preg_replace( '/(\b(?:' . $wpdb->posts . '\.)?post_author\s*=\s*(' . $id . ') )/', '(' . $maybe_both_query . ' ' . $terms_implode . ')', $where, -1 ); #' . $wpdb->postmeta . '.meta_id IS NOT NULL AND`

Perhaps I am not reproducing the issue to a 'T'?  This is on a fresh WP install with CAP however.  Is the author archive we are viewing linked to a guest author and same with the logged in user attempting to view the author archive?

jtbayly:
I'm talking about real users, not guest authors. Perhaps there are two issues? One problem with real users and one for guest authors?

And concerning real users, I'm seeing the problem while logged in as a site admin. I'm not sure whether it matters what role the others have. Just haven't tested at that level of detail.

PR #585, opened 2018-08-08
Make sure post is reassigned to linked guest author if possible, when original author is removed

### LINKED ISSUE #582: Change post_author to Guest Author linked account, if any
Right now, CAP updates the WP_Post->post_author field only if one of the added users is only a WP user. Consider the following scenario:

1. Create two guest authors, link them to two real WP users other than `admin`
2. Logged in as `admin`, create a new post and publish it
3. Add the two guest authors above as guest authors and remove yourself; save post

At this point, the post `post_author` would still hold `admin` as value, even though the post could have been assigned to any other of the two new authors (to be precise, to their linked accounts). This can create issues with counting posts because we'd count this post both as authored from `admin` and from the to guest authors, although `admin` is not there anymore at all.

There is already a piece of code in `add_coauthors()` that reassigns the post if the original author is not among the current co-authors anymore, but it only accepts `wpuser`s and doesn't check for linked accounts:
```php
foreach ( $coauthor_objects as $coauthor_object ) {
	if ( 'wpuser' == $coauthor_object->type ) {
		$new_author = $coauthor_object;
		break;
	}
}
```
The fix should act in this portion of the code, looking at whether `$coauthor_object->linked_account !== ''` and, in case, retrieving the corresponding `WP_User->ID` and assigning it to `$new_author`.

TheCrowned:
Fixes #582. (See issue for steps to reproduce). 

This ensures that, when the original author is removed as coauthor, the post is reassigned (i.e. the `post_author` value updated) to one of the current coauthors. Before, only `wpuser`s would be considered, now also Guest Authors with a linked account are allowed.

Also, manipulating `$author_name` in the foreach ended up making the subsequent `in_array()` check always fail, hence making the _should-we-reassign-the-post-routine_ run at every post save. I have created a new variable and this is also fixed.


GaryJones:
Only [last commit](https://github.com/Automattic/Co-Authors-Plus/pull/585/commits/717912f717df9f5f78ad10db660150d0eb8779d3) is relevant.

PR #586, opened 2018-08-14
Make posts count not rely on term count

### LINKED ISSUE #583: Mismatch between different post counting methods
When [registering a taxonomy](https://codex.wordpress.org/Function_Reference/register_taxonomy), it's possible to provide a callback method that will be invoked when a term is edited, and that will update that term count. This goes in the arg `update_count_callback`. 

Right now, we have this snippet of code in place:
```php
// If we use the nasty SQL query, we need our custom callback. Otherwise, we still need to flush cache.
if ( apply_filters( 'coauthors_plus_should_query_post_author', true ) ) {
	$args['update_count_callback'] = array( $this, '_update_users_posts_count' );
} else {
	add_action( 'edited_term_taxonomy', array( $this, 'action_edited_term_taxonomy_flush_cache' ), 10, 2 );
}
```
and, as far as I know, it is often desired to filter `coauthors_plus_should_query_post_author` to be `false` for performance benefits.

So, in case the arg `update_count_callback` is omitted (and this step is crucial), it will fallback to [`_update_generic_term_count()`](https://developer.wordpress.org/reference/functions/_update_generic_term_count/). That function runs a super-simple query: `SELECT COUNT(*) FROM $wpdb->term_relationships WHERE term_taxonomy_id = %d"`. It just goes and counts links between term and objects, no fancy anything: any post type, any post status... anything that's linked counts. So we have to ask: what post types will have links with `author` terms? Any in `$coauthors_plus->supported_post_types`, as that is used both to display the coauthor box in the edit page AND when running the CLI command `create-terms-for-posts`.

**However**, this is NOT the same behavior achieved by the custom query in `update_author_term_post_count()`. There, [only the `post` post type](https://github.com/Automattic/Co-Authors-Plus/blob/master/co-authors-plus.php#L603) is included in the query. This creates a mismatch between the two counting methods. For example, my setup also has `page` and `download` post types with published items. Those would be counted by the `_update_generic_term_count()`, but not by our custom callback. 

The simplest solution here would be to simply use `$coauthors_plus->supported_post_types` in our custom callback as well. BUT, this would probably break the `count_user_posts()`, which relies on the term count. At that point, the term count would not only contain posts, but any other allowed post type on the site. This would create all sorts of mismatches on the site (`Mine` count in Posts page being wrong; user posts count in Users page being wrong...).

TheCrowned:
Fixes #583.

```php
if ( apply_filters( 'coauthors_plus_should_query_post_author', true ) ) {
	$args['update_count_callback'] = array( $this, '_update_users_posts_count' );
} else {
	add_action( 'edited_term_taxonomy', array( $this, 'action_edited_term_taxonomy_flush_cache' ), 10, 2 );
}
```
This snippet has been removed. As much as we'd like to use the filter `coauthors_plus_should_query_post_author` for performance, turning it off has nasty consequences. For more details, please see the details in #583. This is not so important now, as we are basically changing the way we do coauthors counts.

Up to now, posts count for coauthors was taken from the `count` property of the coauthor term. The main issue here is that there is only one `count` property, but `count_user_posts()` can be called to retrieve count for any post types, or even an array of them. Also, it can be tweaked to include private content or not. So clearly just one `count` is not enough, given the number of variables that can affect the output.

So we must ditch the term count altogether. This means we do not need `_update_users_posts_count()` nor `update_author_term_post_count()` anymore. I have left just their definitions, and added a `_deprecated_function()` call.

The solution I pursued was to retrieve the coauthor content count when requested with `count_user_posts()` and cache it using all the function args (user id, post types, private content). It is indeed a query with two `join`s, but it gets cached and there is no way around it unfortunately. The code for the query was pretty much already there - I just slightly improved it, and let Core `get_posts_by_author_sql()` build the `WHERE` part.

Also, up to now, the posts count query would look for posts that were linked to the given coauthor term OR had that coauthor as `wp_posts.post_author`. Since we have been relying on terms for a long time now, it does not make sense to keep mixing the two things, so I have ditched the `post_author` OR. Also, this would give rise to weird counting mismatches due to the fact that if a post gets assigned to a guest author that does not have a linked account, the WP `post_author` keeps being the latest real user Bob who owned the post. In that case, if we query for coauthor term OR `post_author`, this post will count for Bob as well, although it should not. And there's no way around it other than ignoring `post_author`s.

Finally, I have updated CLI commands and tests to comply with the changes.

Given that we are fully relying on terms for counts, it will be necessary that sites make sure each post has its term for linked authors. So this PR probably depends on #.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
     include:
         # aliased to a recent 5.6.x version
         - php: "5.6"
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '5.6'
           env:
@@ -20,13 +20,13 @@ matrix:
 
         # aliased to a recent 7.x version
         - php: '7.0'
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '7.0'
           env: WP_VERSION=latest
 
         - php: '7.1'
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '7.1'
           env: WP_VERSION=latest
@@ -53,7 +53,7 @@ before_script:
     - |
         if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then
           composer global require "phpunit/phpunit=4.8.*"
-        elif [[ ${TRAVIS_PHP_VERSION:0:2} == "7." && ${WP_VERSION} == "4.6" ]]; then
+        elif [[ ${TRAVIS_PHP_VERSION:0:2} == "7." && ${WP_VERSION} == "4.8" ]]; then
           composer global require "phpunit/phpunit=5.7.*"
         fi
     - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,38 @@
 language: php
 
-php:
-    - 7.1
-
-env:
-    - WP_VERSION=latest
-
 matrix:
-matrix:
-  include:
-   - php: "5.2"
-     env: WP_VERSION=latest
-   - php: "5.2"
-     env: WP_VERSION=4.6
-   - php: "5.6"
-     env:
-         - WP_VERSION=latest
-         - SNIFF=1
-   - php: "5.6"
-     env: WP_VERSION=4.6
-   - php: "7.0"
-     env: WP_VERSION=latest
-   - php: "7.0"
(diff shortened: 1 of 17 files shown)


GaryJones:
It's hard to see what the relevant commits are for this PR, so for the sake of it already having waited 5 years for someone to look at this, I'm going to close it out. If the issue still exists then a fresh PR is welcome.

Issue #550, opened 2018-07-03
Include a uninstall routine


TheCrowned:
Right now, the plugin does not uninstall itself at all. We may want to create a `uninstall.php` file that removes all plugin records from database to avoid leaving it cluttered with potentially unwanted data.

TheCrowned:
Right now I am using this WP_CLI command to clear the database from all non-necessary CAP entries. Useful for testing, it basically reverts `wp co-authors-plus create-guest-authors` and `wp co-authors-plus create-terms-for-posts`. Nor sure we want it in the repo code though.

```php
/**
 * Subcommand to delete guest authors based on users.
 *
 * @subcommand delete-linked-guest-authors
 */
public function delete_linked_guest_authors( $args, $assoc_args ) {
	global $coauthors_plus;

	WP_CLI::confirm( 'Are you sure you want to delete author terms for all users having a linked account? This includes posts and author terms.' );

	$defaults = array(
		// There are no arguments at this time
	);
	$this->args = wp_parse_args( $assoc_args, $defaults );

	$users = get_users();
	$created = 0;
	$skipped = 0;
	$progress = \WP_CLI\Utils\make_progress_bar( 'Processing guest authors...', count ( $users ) );
	foreach ( $users as $user ) {
		
		$guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'linked_account', $user->data->user_login );
		
		if ( ! empty( $guest_author ) ) {
			$coauthors_plus->guest_authors->delete( $guest_author->ID );
		}			
		$progress->tick();
	}
	$progress->finish();
	WP_CLI::success( sprintf( 'All done! Processed %d users.', count( $users ) ) );
}
```

PR #588, opened 2018-08-14
Add wp-cli command to delete guest authors, with option to reassign their posts.

smistephen:
Currently, there is no way to delete guest authors from the command line. This command fills that need.

It is activated using `delete-guest-authors`, and passing in the path to a CSV file containing a list of two columns: the ID to delete, and the login to assign the posts to (or "false" to not reassign).

The command runs dry by default, showing the end user the guest authors it would have attempted to delete, and the user login it would have attempted to transfer the posts to.

Passing --no-dry-run, and confirming that you really do want to delete guest authors, actively modifies the database.


smistephen:
I'm not sure why the Travis build fails, it passes in my fork:

https://github.com/smistephen/Co-Authors-Plus/runs/12256989

TheCrowned:
This looks great! Could we allow deletion of all guest authors that have linked accounts? See #550. Would be good for testing, because allows to basically revert the install process. Maybe the csv thing could become an option (not compulsory) and in some way incorporate this bulk automatic deletion. Only include guest authors with linked accounts as we wouldn't want to delete the others, as there would be no fallback in post assignment.

smistephen:
Totally :+1:

Could you open a separate issue for those improvements? This PR's already pretty big, haha.

TheCrowned:
It's already in #550

smistephen:
Perfect

smistephen:
Looking at #550, it seems like it could be its own independent PR. Am I reading it right?

sboisvert (commented):
A few questions :)

php/class-wp-cli.php
```diff
@@ -45,6 +45,120 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+					$id,
+					$assignee
+				) );
+			} else {
+				// Have to cast so the function won't literally try and reassign to user "false".
+				if ( 'false' === $assignee ) {
+					$assignee = false;
+				}
+
+				$delete_status = $coauthors_plus->guest_authors->delete( $id, $assignee );
```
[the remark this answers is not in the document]
smistephen: The function returns a WP_Error:
https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-coauthors-guest-authors.php#L1304

php/class-wp-cli.php
```diff
@@ -45,6 +45,120 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+		// Safety first.
+		$filename_status = validate_file( $args[0] );
+
+		if ( 0 !== $filename_status ) {
+			WP_CLI::error( __( 'The provided file cannot be accessed. Please use either relative or absolute pathnames, and avoid using directory traversal such as ../ and other methods.', 'co-authors-plus' ) );
+		}
+
+		$filename = $args[0];
+
+		// Add custom error handling on fopen error.
```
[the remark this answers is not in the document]
smistephen: I felt it was a clearer variable name.

php/class-wp-cli.php
```diff
@@ -45,6 +45,120 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+		}
+
+		while ( ( $line = fgetcsv( $file, 100, ',' ) ) !== false ) {
+			$id       = filter_var( $line[0], FILTER_SANITIZE_NUMBER_INT );
+			$assignee = filter_var( $line[1], FILTER_SANITIZE_STRING );
+
+			if ( true === $dry_run ) {
+				WP_CLI::log( sprintf(
+					/* translators: 1: Guest author ID 2: Intended assignee */
+					__( 'Attempting to delete guest author %1$d, intended assignee %2$s', 'co-authors-plus' ),
```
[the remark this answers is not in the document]
smistephen: Sure, I'll see what I can pluck out :+1:

php/class-wp-cli.php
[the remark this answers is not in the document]
sboisvert: Sorry, I meant the custom error handling, not the variable name :)

php/class-wp-cli.php
```diff
@@ -45,6 +45,120 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+							/* translators: 1: Guest author ID 2: Error message returned from guest author deletion attempt */
+							__( 'Error while attempting to delete guest author %1$d: %2$s', 'co-authors-plus' ),
+							$id,
+							$error_message
+						) );
+					}
+				} else {
+					WP_CLI::log( sprintf(
+						/* translators: %d: Guest author ID */
+						__( 'Successfully deleted guest author %d', 'co-authors-plus' ),
```
sboisvert: If someone looked back at this log file file. Is there more information they would need?

php/class-wp-cli.php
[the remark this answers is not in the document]
sboisvert: Would that error message be enough to debug the problem? Is there more information we could add to help the end user?

php/class-wp-cli.php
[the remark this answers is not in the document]
smistephen: Ah. The custom error handling is there to avoid an ugly PHP warning and stack trace from being thrown to the console if the CSV file can't be opened, doesn't exist, or some other error.

sboisvert:
Nod, it might be useful for the comment to explain the intent behind this
since I wouldn't say it's a common pattern :)

php/class-wp-cli.php
```diff
@@ -45,6 +45,145 @@ public function create_guest_authors( $args, $assoc_args ) {
 		WP_CLI::line( "- {$skipped} users already had guest author profiles" );
 	}
 
+	/**
+	 * Delete guest authors.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <guest-author-ids-file>
+	 * : Provide the location of a CSV file containing a list of two columns: guest author ID to delete, and either the string "false" (which means the deleted guest author's posts will not be reassigned) or a string denoting the login of the user or guest author the deleted guest author's posts are to be reassigned to. No column headers are required or allowed, and the maximum line length including ID, login, and comma is 100 characters.
```
philipjohn: Could we break this into multiple lines for easier readability?

php/class-wp-cli.php
```diff
@@ -45,6 +45,145 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+						$id,
+						$assignee,
+						$guest_author->display_name,
+						$guest_author->user_email,
+						$guest_author->linked_account
+					) );
+				}
+			} else {
+				// Have to cast so the function won't literally try and reassign to user "false".
+				if ( 'false' === $assignee ) {
```
philipjohn: What happens if someone has the username "false"? Or "null" or similar?

php/class-wp-cli.php
```diff
@@ -45,6 +45,145 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+		// Are you really, really sure.
+		if ( false === $dry_run ) {
+			WP_CLI::confirm( __( 'You have provided the --no-dry-run flag. Please confirm you wish to delete guest authors.', 'co-authors-plus' ) );
+		} else {
+			WP_CLI::log( __( 'DRY RUN', 'co-authors-plus' ) );
+		}
+
+		while ( ( $line = fgetcsv( $file, 100, ',' ) ) !== false ) {
+			$id       = filter_var( $line[0], FILTER_SANITIZE_NUMBER_INT );
+			$assignee = filter_var( $line[1], FILTER_SANITIZE_STRING );
```
philipjohn: What happens if this assignee doesn't exist?

philipjohn (commented):
This is looking good - I've left some comment.

One overall question - how can we check this code is going to work as expected in production?

php/class-wp-cli.php
smistephen: The function returns a WP_Error :+1: 
https://github.com/Automattic/Co-Authors-Plus/pull/588#discussion_r210121608

smistephen:
> One overall question - how can we check this code is going to work as expected in production?

Checks I've performed so far:
 - run phpunit _(passes)_
 - run phpcs _(passes)_
 - look for performance problems and security vulnerabilities _(nothing obvious)_
 - test function return values _(added appropriate error handling code)_
 - test on a sample WordPress instance _(seems to work as promised)_
 - try to break it with faulty data _(seems to fail gracefully as predicted)_

And I'm always open to more opportunities :+1:

php/class-wp-cli.php
smistephen: Haha, you tricky trickster. I've added code to deal with that now :+1:

smistephen:
Hooray for @philipjohn pulling the gremlin out of Travis, builds pass now, as they should :+1:

php/class-wp-cli.php
philipjohn: I was referring to the value of `$assignee` - what happens if the assignee (i.e. the username) given in the CSV doesn't exist? What will this code do to that value?

philipjohn:
> run phpunit (passes)

Which unit tests did you look at when trying to assess whether this code works as intended?

> try to break it with faulty data (seems to fail gracefully as predicted)

That's a good idea! What kind of faulty data did you use here?

php/class-wp-cli.php
smistephen: Ah yes. That is an error, and it has now been fixed.

php/class-wp-cli.php
```diff
@@ -45,6 +45,145 @@ public function create_guest_authors( $args, $assoc_args ) {
...
+				if ( false === $guest_author ) {
+					WP_CLI::log( sprintf(
+						/* translators: 1: Guest author ID 2: Intended assignee */
+						__( 'Attempting to delete guest author %d, but they do not exist.', 'co-authors-plus' ),
+						$id
+					) );
+				} else {
+					WP_CLI::log( sprintf(
+						/* translators: 1: Guest author ID 2: Intended assignee 3: Guest author display name 4: Guest author email 5: Guest author linked account */
+						__( 'Attempting to delete guest author %1$d, intended assignee %2$s. Display Name: %3$s Email: %4$s Linked Account: %5$s', 'co-authors-plus' ),
```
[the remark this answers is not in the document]
smistephen: I've rearranged the output, similar to elsewhere in the file: https://github.com/Automattic/Co-Authors-Plus/blob/78ce17c2b820f8baa1f35254c0f03b75184e1755/php/class-wp-cli.php#L162

chrissy-dev:
@rebeccahum I appreciate this is an old PR but is there a reason this PR never got merged? Happy to help get this over the line if possible.

rebeccahum:
@chrissy-dev I think this just slipped the radar. First step would be to resolve any merge conflicts and test if the current PR still works please!

GaryJones:
Overlaps with #696.

GaryJones:
Thanks for this, @smistephen, and apologies it sat for so long.

We're closing this out rather than merging. After eight years of `develop` drift the branch is conflicting across 19 files, so landing it would mean a rewrite rather than a rebase. The design has also moved on: the tracking issue (#682) specifies a more WP-CLI-idiomatic signature — `wp co-authors-plus delete <id> [--force] [--assign-to=...]` — rather than the CSV-file approach here.

The feature itself is still wanted, so #682 stays open as the home for it. Your dry-run-by-default behaviour and post-reassignment logic remain a useful reference for whoever picks it up — this PR is linked from #682 so the code stays one click away.

Closing in favour of #682.

PR #587, opened 2018-08-14
Delaying cap_import_existing_users cron by 15 seconds.

TheCrowned:
Delaying cron event `cap_import_existing_users` spawning by 15 seconds to make sure they don't overlap, as requested in #544.

PR #598, opened 2018-08-27
Fix for "Stop handling wp_terms in search_authors, do it in install and profile update"

TheCrowned:
This is a follow up from PR #544, which broke tests and went un-noticed. **This PR depends on #544, and requires that both are merged together - they should be treated as one single PR.**

The main point of failure is that, previous to that PR, the default CAP behavior was to allow the creation of guest authors for already present WP users, but did not force it. This was still manually enforced most of the times, mostly with WP_CLI command `wp co-authors-plus create-guest-authors`. So all tests created users _and then_ also took care of creating corresponding guest authors profiles, which would throw errors because guest authors were already in place for all users. This explains the various replacements of calls to `create_guest_author_by_id()` with calls to `get_guest_author_by()` in this PR.

Furthermore, template tags function `get_coauthors()` and all its children would expect WP users IDs as output, since the tests for those were not explicitly creating guest authors for WP users. However, all WP users have a guest author by default now, so I had to change that as well. Instead of expecting the WP user ID, they now expect the Coauthor ID (i.e. che Term ID).

I have also had to revert changes from PR #437, as they made a good bunch of tests to fail. The issue is that, again, all accounts now have a linked account, and that code was removing them all, always returning an empty list. I have tested the template tag and no duplicate entries are displayed for me, which was the original reason why that code was in place.

Finally, I have slightly edited some tests/removed some of them where they did not apply to the current setup anymore. One example is the test for `create_guest_author_by_user_id()`. To run it, I would have had to delete an existing guest author and then create it again, which seemed pointless. If the function ever failed, a good 20+ tests would fail altogether, which would be a good red sign I guess.

I have also taken the chance and included PR #587 changes into this one, since it belonged here. It delays the cron event re-spawning by 15 seconds on each run.

With these changes, all tests pass but one:

screenshot from 2018-08-27 16-47-07

That test will pass when PR #585 is merged (I can confirm this is true, I have tested it). See that PR for details, but the issue is again due to the chance in behavior for guest authors being present by default for all WP users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
     include:
         # aliased to a recent 5.6.x version
         - php: "5.6"
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '5.6'
           env:
@@ -20,13 +20,13 @@ matrix:
 
         # aliased to a recent 7.x version
         - php: '7.0'
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '7.0'
           env: WP_VERSION=latest
 
         - php: '7.1'
-          env: WP_VERSION=4.6
+          env: WP_VERSION=4.8
 
         - php: '7.1'
           env: WP_VERSION=latest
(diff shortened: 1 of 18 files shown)


rebeccahum:
Just a note, but #544 was reverted and the revert for that will need to be reverted when this is merged.

GaryJones:
Since #590 reverted #544, and this PR contains fixes for multiple issues that are hard (now) to untangle, I'm closing this one out.

PR #600, opened 2018-09-05
Add CLI argument to assign-coauthors

WPprodigy:
This argument is used in the code already, but isn’t allowed to be passed down since it’s not in the synopsis.

It ends up being passed to `add_coauthors()` and determines if the co-author being added should replace all of the existing coauthors, or if it should be appended to the existing ones.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -121,7 +121,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	 * @since 3.0
 	 *
 	 * @subcommand assign-coauthors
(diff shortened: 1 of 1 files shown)


TheCrowned:
All good, but [it is advised](https://make.wordpress.org/cli/handbook/commands-cookbook/#longdesc) to use `--append_coauthors` alone, without the `<boolean>` marker. This would make the argument boolean by default, and also allow to set it to false using `--no-append_coauthors`. :)

WPprodigy:
It defaults to false, so there shouldn't be a need for the `--no-append_coauthors` option.

PR #602, opened 2018-09-05
Avoid PHP Fatal in calling guest_authors->get_guest_author_by

### LINKED ISSUE #601: Not possible to delete users when the guest authors functionality is disabled
It's not possible to delete users if the guest author functionality is disabled.

I get the following PHP error:

`PHP Fatal error: Call to a member function get_guest_author_by() on null in /wp-content/plugins/co-authors-plus/co-authors-plus.php on line 943`.

The guest author functionality is disabled with the following filter:

`add_filter( ‘coauthors_guest_authors_enabled’, ‘__return_false’ );`

TheCrowned:
This fixes #601.

The issue was that `guest_authors->get_guest_author_by()` was being called without prior checking that guest author functions were enabled, and thus resulting in a PHP Fatal because `guest_authors` was `null`.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -940,15 +940,17 @@ class CoAuthors_Plus {
 			wp_delete_term( $delete_user->user_login, $this->coauthor_taxonomy );
 		}
 
-		// Get the deleted user data by user id.
-		$user_data = get_user_by( 'id', $delete_id );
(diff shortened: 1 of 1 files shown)


rebeccahum:
Thanks for this fix!  Looks great.

Issue #552, opened 2018-07-04
Function to retrieve guest author post count


TheCrowned:
There just isn't one at the moment ([see here](https://wordpress.org/support/topic/co-authors-post-count/))!

rebeccahum:
Closing from #605.

PR #605, opened 2018-09-18
Add a method to fetch the guest author post count

shantanu2704:
Addresses #552 

Extracted the logic to get the post count from the `cloumn_posts` function of the `CoAuthors_WP_List_Table` class. 
Then added a wrapper for the newly created function in the class `CoAuthors_Guest_Authors` so that all guest author functionality remains in one class.
Made use of the new wrapper method in the `view_guest_authors_list` method.

Tested the changes with multiple guest authors with a varying number of posts.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1649,22 +1649,29 @@ class CoAuthors_Plus {
 	/**
 	 * Get the post count for the guest author
 	 *
-	 * @global oebject $coauthors_plus
-	 * @param object $item guest-author object.
+	 * @param object $guest_author guest-author object.
 	 * @return int post count for the guest author
 	 */
-	public function get_guest_author_post_count( $item ) {
-		$term       = $this->get_author_term( $item );
-		$guest_term = get_term_by( 'slug', 'cap-' . $item->user_nicename, $this->coauthor_taxonomy );
-		if ( ! empty( $item->linked_account ) && $guest_term->count ) {
-			return count_user_posts( get_user_by( 'login', $item->linked_account )->ID );
+	public function get_guest_author_post_count( $guest_author ) {
+		if ( ! is_object( $guest_author ) ) {
+			return;
+		}
+
+		$term = $this->get_author_term( $guest_author );
+		$guest_term = get_term_by( 'slug', 'cap-' . $guest_author->user_nicename, $this->coauthor_taxonomy );
+
+		if ( ! $term || ! $guest_term ) {
+			return false;
+		}
+
+		if ( ! empty( $guest_author->linked_account ) && $guest_term->count ) {
+			return count_user_posts( get_user_by( 'login', $guest_author->linked_account )->ID );
 		} elseif ( $term ) {
 			return $term->count;
 		} else {
 			return 0;
 		}
 	}
-
 
 }
 
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1657,14 +1657,12 @@ class CoAuthors_Plus {
 			return;
 		}
 
-		$term = $this->get_author_term( $guest_author );
+		$term       = $this->get_author_term( $guest_author );
 		$guest_term = get_term_by( 'slug', 'cap-' . $guest_author->user_nicename, $this->coauthor_taxonomy );
 
-		if ( ! $term || ! $guest_term ) {
-			return false;
-		}
-
-		if ( ! empty( $guest_author->linked_account ) && $guest_term->count ) {
+		if ( is_object( $guest_term )
+			&& ! empty( $guest_author->linked_account )
+			&& $guest_term->count ) {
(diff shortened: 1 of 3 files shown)


php/class-coauthors-wp-list-table.php
```diff
@@ -249,17 +249,28 @@ function column_linked_account( $item ) {
...
+	}
+
+	/**
+	 * Get the post count for the guest author
+	 *
+	 * @global oebject $coauthors_plus
+	 * @param object $item guest-author object.
+	 * @return int post count for the guest author
+	 */
+	function get_post_count( $item ) {
```
rebeccahum: I would suggest moving this function to the `CoAuthors_Plus` class for wider use, instead of having to instantiate a `CoAuthors_WP_List_Table` class.

co-authors-plus.php
```diff
@@ -1644,8 +1644,28 @@ public function filter_author_archive_title( $title ) {
 		$author = $this->get_coauthor_by( 'user_nicename', $author_slug );
 		
 		return sprintf( __( 'Author: %s' ), $author->display_name );
+	}	
+
+	/**
+	 * Get the post count for the guest author
+	 *
+	 * @global oebject $coauthors_plus
```
rebeccahum: I don't think this phpdoc comment is necessary anymore since we've moved the function into the `CoAuthors_Plus` class.

co-authors-plus.php
```diff
@@ -1644,8 +1644,28 @@ public function filter_author_archive_title( $title ) {
...
+	 * @param object $item guest-author object.
+	 * @return int post count for the guest author
+	 */
+	public function get_guest_author_post_count( $item ) {
```
rebeccahum: I'd suggest renaming this to something more descriptive (e.g. `$guest_author`).

co-authors-plus.php
```diff
@@ -1644,8 +1644,28 @@ public function filter_author_archive_title( $title ) {
...
+		$term       = $this->get_author_term( $item );
```
rebeccahum: What happens if `get_author_term()` returns `NULL` from passing in an invalid variable type?

co-authors-plus.php
```diff
@@ -1644,8 +1644,28 @@ public function filter_author_archive_title( $title ) {
...
+		if ( ! empty( $item->linked_account ) && $guest_term->count ) {
+			return count_user_posts( get_user_by( 'login', $item->linked_account )->ID );
+		} elseif ( $term ) {
+			return $term->count;
+		} else {
+			return 0;
+		}
 	}
 
+
```
rebeccahum: It'd be nice to remove this whitespace from the PR.

co-authors-plus.php
shantanu2704: It'll be caught by the check on line number 1661. Since that'll fail, line 1664 will return 0.

shantanu2704: We could bail early or throw a `WP_Error` - `missing-guest-author`, if the parameter `$item` (renaming to `guest_author`) isn't an object. Is that what you had in mind?

rebeccahum: > It'll be caught by the check on line number 1661. Since that'll fail, line 1664 will return 0.

That's correct, however, it will throw an error though because it will try to grab the non-existent properties of `user_nicename`.  What do you think is the best way to go about?  What are your thoughts in returning `false`?

shantanu2704:
[x] Functional Testing
[x] Negative testing using `wp shell`

co-authors-plus.php
```diff
@@ -1644,6 +1644,33 @@ public function filter_author_archive_title( $title ) {
...
+	 */
+	public function get_guest_author_post_count( $guest_author ) {
+		if ( ! is_object( $guest_author ) ) {
+			return;
+		}
+
+		$term = $this->get_author_term( $guest_author );
+		$guest_term = get_term_by( 'slug', 'cap-' . $guest_author->user_nicename, $this->coauthor_taxonomy );
+
+		if ( ! $term || ! $guest_term ) {
```
rebeccahum: I'm not sure if this check is necessary since it will return 0 in line 1672 if either conditions are met. What are your thoughts?

shantanu2704: I think it is, since if either of those values is `false` like you mentioned in a previous comment, it would throw a notice - `Trying to get property of non-object`

shantanu2704: do you think this check makes the last condition on lines 1671 to 1673 redundant? I think it'd be safe to remove those three lines instead of removing this check

rebeccahum: What if `$guest_term` returns false but `$term` returns a valid value?

rebeccahum: That's correct, that check is still necessary.

shantanu2704:
The commit message is supposed to be - 
`Add an is_object check for $guest_term`.
I think `$guest_term` was not parsed by GH

Issue #266, opened 2015-05-28
Add ability to add classes to coauthors_get_avatar


corysimmons:
Anytime you're outputting an HTML tag, please offer the ability to add classes to it.

https://github.com/Automattic/Co-Authors-Plus/blob/cb21f6b3e7a129efd11e495d80622bcd8dc63561/template-tags.php#L561

corysimmons:
Or add the ability to simply get the src of the image.

PR #610, opened 2018-10-01
Add the ability to add additional classes to coauthors_get_avatar()

shantanu2704:
Addresses #266.

Based on the type of co-author, `coauthors_get_avatar()` calls either `get_guest_author_thumbnail()` or `get_avatar()`.

[`get_avatar()`](https://github.com/WordPress/WordPress/blob/master/wp-includes/pluggable.php#L2533) accepts an `args` array, an element of which is an `array|string` of additional classes.

A similar modification is required for [`get_guest_author_thumbnail()`](https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-coauthors-guest-authors.php#L965) which will also accept an additional parameter `$class`; a string of additional classes. 

[`coauthors_get_avatar()`](https://github.com/Automattic/Co-Authors-Plus/blob/master/template-tags.php#L683) now accepts an optional param `$class` which will be either an array or string of additional classes, the default value of which will be `null`. If it is an array of classes, it will be converted to a space delimited string using `implode()` and passed to either `get_guest_author_thumbnail()` or `get_avatar()`.

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -676,19 +676,31 @@ function coauthors_wp_list_authors( $args = array() ) {
  * This is a replacement for using get_avatar(), which only operates on email addresses and cannot differentiate
  * between Guest Authors (who may share an email) and regular user accounts
  *
- * @param  object   $coauthor The Co Author or Guest Author object
- * @param  int      $size     The desired size
- * @return string             The image tag for the avatar, or an empty string if none could be determined
+ * @param  object        $coauthor The Co Author or Guest Author object.
+ * @param  int           $size     The desired size.
+ * @param  string        $default  Optional. URL for the default image or a default type. Accepts '404'
+ *                                 (return a 404 instead of a default image), 'retro' (8bit), 'monsterid'
+ *                                 (monster), 'wavatar' (cartoon face), 'indenticon' (the "quilt"),
+ *                                 'mystery', 'mm', or 'mysteryman' (The Oyster Man), 'blank' (transparent GIF),
+ *                                 or 'gravatar_default' (the Gravatar logo). Default is the value of the
+ *                                 'avatar_default' option, with a fallback of 'mystery'.
+ * @param  string        $alt      Optional. Alternative text to use in &lt;img&gt; tag. Default false.
+ * @param  array|string  $class    Array or string of additional classes to add to the &lt;img&gt; element. Default null.
+ * @return string                  The image tag for the avatar, or an empty string if none could be determined.
  */
-function coauthors_get_avatar( $coauthor, $size = 32, $default = '', $alt = false ) {
+function coauthors_get_avatar( $coauthor, $size = 32, $default = '', $alt = false, $class = null ) {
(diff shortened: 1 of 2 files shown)


template-tags.php
```diff
@@ -676,19 +676,31 @@ function coauthors_wp_list_authors( $args = array() ) {
...
  */
-function coauthors_get_avatar( $coauthor, $size = 32, $default = '', $alt = false ) {
+function coauthors_get_avatar( $coauthor, $size = 32, $default = '', $alt = false, $class = null ) {
 	global $coauthors_plus;
 
 	if ( ! is_object( $coauthor ) ) {
 		return '';
 	}
 
+	if ( ! empty( $class ) && is_array( $class ) ) {
```
rebeccahum: What if `$class` is passed in as a string datatype?  Since I see in the PHPDocs for this function, string is allowed:

> @param  array|string  $class    Array or string of additional classes to add to the &lt;img&gt; element. Default null.

shantanu2704: If `$class` is passed as a string, we don't have to convert it to anything. Just pass it as-is to `get_guest_author_thumbnail()` or `get_avatar()`. That string will then be appended to the existing classes in those functions.

template-tags.php
```diff
@@ -697,7 +709,7 @@ function coauthors_get_avatar( $coauthor, $size = 32, $default = '', $alt = fals
 
 	// Make sure we're dealing with an object for which we can retrieve an email
 	if ( isset( $coauthor->user_email ) ) {
-		return get_avatar( $coauthor->user_email, $size, $default, $alt );
+		return get_avatar( $coauthor->user_email, $size, $default, $alt, array( 'class' => $class ) );
```
rebeccahum: I'm seeing that for `get_avatar()`, you can pass the `$class` in as an array, so would it be necessary to `implode()` it on line 699? https://codex.wordpress.org/Function_Reference/get_avatar Related to comment above ^^

template-tags.php
rebeccahum: Do you think it would be better to move this check into `get_guest_author_thumbnail()`, so then we can pass in `$class` as either an `array` or `string` datatype?

template-tags.php
shantanu2704: The value in `$args['class']` for `get_avatar()` can be an array or a string. So I don't think we need to `implode()`.
https://github.com/WordPress/WordPress/blob/master/wp-includes/pluggable.php#L2533

However, I like your previous suggestion of moving the check on line 698 to `get_guest_author_thumbnail()`. That way, there will be minimal changes to this function and the logic for differentiating between an array and string will be the responsibility of  `get_guest_author_thumbnail` and `get_avatar()`

template-tags.php
shantanu2704: Yes, I agree. That way we can pass in the $class parameter as is, and `get_guest_author_thumbnail()` and `get_avatar()` will check if it's an array or string.

PR #617, opened 2018-11-05
Enable CAP Query modifications for "any" post_type

rebeccahum:
Currently the CAP plugin does a taxonomy check to determine if it should modify the queries groupby, where and join filters. However the filters are not setup to handle "any" post_type. This patch uses the logic from WP_Query to convert "any" to an array of post_types so that it can be properly tested if modifications should be applied to the query.


mdbitz:
I remember this code :)

Issue #452, opened 2017-09-06
Filtering the get_avatar hook does not change avatars in author selection widget


fschroiff:
With javascript active the author selection widget seems to only show gravatar avatars. I filter `get_avatar` to show some other image. Would be nice if that could be reflected in the author selection widget as well.

mostafaabd:
Hello, i'm a new contributor,  so can i fix this issue, and also there is no more documentation to test my changes before send pull request for php and jquery code.

rebeccahum:
Closed per #621.

PR #621, opened 2018-11-10
fix issue 452

mostafaabd:
fix issue #452

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -386,7 +386,6 @@ class CoAuthors_Plus {
 					}
 					?>
 					<li>
-						<?php echo get_avatar( $coauthor->user_email, $this->gravatar_size ); ?>
 						<span id="<?php echo esc_attr( 'coauthor-readonly-' . $count ); ?>" class="coauthor-tag">
 							<input type="text" name="coauthorsinput[]" readonly="readonly" value="<?php echo esc_attr( $coauthor->display_name ); ?>" />
 							<input type="text" name="coauthors[]" value="<?php echo esc_attr( $coauthor->user_login ); ?>" />
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -380,6 +380,10 @@ class CoAuthors_Plus {
 				<?php
 				foreach ( $coauthors as $coauthor ) :
 					$count++;
+					$avatar_url = '';
+					if ( has_post_thumbnail( $coauthor->ID ) ) {
+						$avatar_url = get_the_post_thumbnail_url( $coauthor->ID, $this->gravatar_size );
+					}
 					?>
 					<li>
 						<?php echo get_avatar( $coauthor->user_email, $this->gravatar_size ); ?>
@@ -388,6 +392,7 @@ class CoAuthors_Plus {
 							<input type="text" name="coauthors[]" value="<?php echo esc_attr( $coauthor->user_login ); ?>" />
 							<input type="text" name="coauthorsemails[]" value="<?php echo esc_attr( $coauthor->user_email ); ?>" />
 							<input type="text" name="coauthorsnicenames[]" value="<?php echo esc_attr( $coauthor->user_nicename ); ?>" />
+							<input type="text" name="coauthorsavaters[]" value="<?php echo esc_attr( $avatar_url ); ?>" />
 						</span>
 					</li>
 					<?php
@@ -1168,7 +1173,11 @@ class CoAuthors_Plus {
 		if( empty( $authors ) ) echo apply_filters( 'coauthors_no_matching_authors_message', 'Sorry, no matching authors found.');
 
 		foreach ( $authors as $author ) {
-			echo esc_html( $author->ID . ' | ' . $author->user_login . ' | ' . $author->display_name . ' | ' . $author->user_email . ' | ' . rawurldecode( $author->user_nicename ) ) . "\n";
+			$avatar_url = '';
+			if ( has_post_thumbnail( $author->ID ) ) {
+				$avatar_url = get_the_post_thumbnail_url($author->ID, $this->gravatar_size );
+			}
+			echo esc_html( $author->ID . ' | ' . $author->user_login . ' | ' . $author->display_name . ' | ' . $author->user_email . ' | ' . rawurldecode( $author->user_nicename ) ) . ' | ' . esc_url( $avatar_url ) . "\n";
 		}
 
 		die();
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -224,7 +224,6 @@ class CoAuthors_Plus {
 
 		// Apply some targeted filters
 		add_action( 'load-edit.php', array( $this, 'load_edit' ) );
-		
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


mostafaabd:
is there is wrong in my PR @rebeccahum

rebeccahum:
@mostafaabd Hi!  Unfortunately, this PR doesn't work for me when you filter `get_avatar`.  
screen shot 2018-12-27 at 5 09 50 pm

mostafaabd:
@rebeccahum 
To test it, author need to have assign avatar in guest author profile page see image 1, 2
What i do 
1- check if user assign avatar image in guest author profile page.
2- If [**Yes**] Display it.
2- if **[No**] Return to main code that get avatar by email if not found display default avatar.

**to Improve this solution in feature ** 
- Remove jquery code in co-author-plus.js MD5 (Message-Digest Algorithm) by WebToolkit -- needed for gravatars.
- In PHP need to do it
1 - if user assign avatar image in guest author profile page. [Display it]
2 - Check If user has avatar image using get_avatar_url() function , if not found using php code [here ](https://en.gravatar.com/site/implement/images/) to display default avatar image

avatar-1

avatar-2

rebeccahum:
@mostafaabd Yep, there is an avatar assigned to the guest author and it still does not show the expected avatar for me.  Does this PR work for you on a fresh local WP install with just CAP?

mostafaabd:
@rebeccahum , yes i test it and you can test it please ctrl+f5 to load new js file not cached file.
and i update PR because there is one line not need in function that display meta for this reasons:
1- this line display avatar image but plugin use jquery to display all fields related to author.

co-authors-plus.php
```diff
@@ -380,14 +380,18 @@ public function coauthors_meta_box( $post ) {
 				<?php
 				foreach ( $coauthors as $coauthor ) :
 					$count++;
+					$avatar_url = '';
+					if ( has_post_thumbnail( $coauthor->ID ) ) {
+						$avatar_url = get_the_post_thumbnail_url( $coauthor->ID, $this->gravatar_size );
```
rebeccahum: What happens if we filter `get_avatar()`?  Should that be displayed instead?

rebeccahum: What happens if we filter `get_avatar_url`? Should that be displayed instead?

co-authors-plus.php
```diff
@@ -1168,7 +1172,11 @@ public function ajax_suggest() {
 		if( empty( $authors ) ) echo apply_filters( 'coauthors_no_matching_authors_message', 'Sorry, no matching authors found.');
 
 		foreach ( $authors as $author ) {
-			echo esc_html( $author->ID . ' | ' . $author->user_login . ' | ' . $author->display_name . ' | ' . $author->user_email . ' | ' . rawurldecode( $author->user_nicename ) ) . "\n";
+			$avatar_url = '';
+			if ( has_post_thumbnail( $author->ID ) ) {
+				$avatar_url = get_the_post_thumbnail_url($author->ID, $this->gravatar_size );
```
rebeccahum: What happens if we filter `get_avatar_url`? Should that be displayed instead?

co-authors-plus.php
```diff
@@ -221,7 +224,7 @@ public function admin_init() {
 
 		// Apply some targeted filters
 		add_action( 'load-edit.php', array( $this, 'load_edit' ) );
-
+
```
rebeccahum: Looks like whitespace got addeded unintentionally, please remove it.

co-authors-plus.php
```diff
@@ -380,14 +383,15 @@ public function coauthors_meta_box( $post ) {
 				<?php
 				foreach ( $coauthors as $coauthor ) :
 					$count++;
+					$avatar_url = get_avatar_url( $coauthor->ID, array( 'default' => 'gravatar_default' ) );
 					?>
 					<li>
-						<?php echo get_avatar( $coauthor->user_email, $this->gravatar_size ); ?>
```
rebeccahum: Should we keep the `get_avatar()` being `echo`'d because of `filter_get_avatar` already existing in CAP and incorporate the logic of the changes in that function? https://github.com/Automattic/Co-Authors-Plus/blob/f43f5034816ba59a60edb4abe85da0af13eb1912/php/class-coauthors-guest-authors.php#L1428

js/co-authors-plus.js
```diff
@@ -389,6 +396,7 @@ jQuery( document ).ready(function () {
 		var $post_coauthor_names = jQuery( 'input[name="coauthorsinput[]"]' );
 		var $post_coauthor_emails = jQuery( 'input[name="coauthorsemails[]"]' );
 		var $post_coauthor_nicenames = jQuery( 'input[name="coauthorsnicenames[]"]' );
+		var $post_coauthoravatar = jQuery( 'input[name="coauthorsavaters[]"]' );
```
rebeccahum: Not a big deal, but can we fix the typo here of `avatar`? It should be `coauthorsavatars`.

co-authors-plus.php
```diff
@@ -380,14 +380,18 @@ public function coauthors_meta_box( $post ) {
...
+					}
 					?>
 					<li>
-						<?php echo get_avatar( $coauthor->user_email, $this->gravatar_size ); ?>
 						<span id="<?php echo esc_attr( 'coauthor-readonly-' . $count ); ?>" class="coauthor-tag">
 							<input type="text" name="coauthorsinput[]" readonly="readonly" value="<?php echo esc_attr( $coauthor->display_name ); ?>" />
 							<input type="text" name="coauthors[]" value="<?php echo esc_attr( $coauthor->user_login ); ?>" />
 							<input type="text" name="coauthorsemails[]" value="<?php echo esc_attr( $coauthor->user_email ); ?>" />
 							<input type="text" name="coauthorsnicenames[]" value="<?php echo esc_attr( $coauthor->user_nicename ); ?>" />
+							<input type="text" name="coauthorsavaters[]" value="<?php echo esc_attr( $avatar_url ); ?>" />
```
rebeccahum: Not a big deal, but can we fix the typo here of `avatar`? It should be `coauthorsavatars`.

co-authors-plus.php
```diff
@@ -1689,7 +1694,21 @@ public function get_guest_author_post_count( $guest_author ) {
...
+	 * @param $url string Avatar URL
+	 * @param $id  int Author ID
+	 *
+	 * @return string Avatar URL
+	 */
+	public function filter_get_avatar_url( $url, $id ) {
+		if ( has_post_thumbnail( $id ) ) {
+			$url = get_the_post_thumbnail_url( $id, $this->gravatar_size );
+		}
+		return esc_url( $url );
```
rebeccahum: It won't be necessary to escape `$url` since it's not being outputted directly.

co-authors-plus.php
rebeccahum: Since we're filtering `get_avatar_url`, we should still be able to use `get_avatar`.

rebeccahum: We can still use `get_avatar()` here for display in the case that there is no JS.  I'd recommend passing in the coauthor `ID` instead of the `user_email` though since guest authors may not always have an email assigned.

co-authors-plus.php
```diff
@@ -380,14 +383,15 @@ public function coauthors_meta_box( $post ) {
...
+					$avatar_url = get_avatar_url( $coauthor->ID, array( 'default' => 'gravatar_default' ) );
 					?>
 					<li>
-						<?php echo get_avatar( $coauthor->user_email, $this->gravatar_size ); ?>
 						<span id="<?php echo esc_attr( 'coauthor-readonly-' . $count ); ?>" class="coauthor-tag">
 							<input type="text" name="coauthorsinput[]" readonly="readonly" value="<?php echo esc_attr( $coauthor->display_name ); ?>" />
 							<input type="text" name="coauthors[]" value="<?php echo esc_attr( $coauthor->user_login ); ?>" />
 							<input type="text" name="coauthorsemails[]" value="<?php echo esc_attr( $coauthor->user_email ); ?>" />
 							<input type="text" name="coauthorsnicenames[]" value="<?php echo esc_attr( $coauthor->user_nicename ); ?>" />
+							<input type="text" name="coauthorsavaters[]" value="<?php echo esc_attr( $avatar_url ); ?>" />
```
rebeccahum: If we keep the `get_avatar()` call above, we can change the input type to `hidden` instead of `text` since it may not be necessary need to display the URL to the `avatar_url` in a text field.

Also, not a big deal, but we should fix the typo here of `avatar` so that the `name` attribute should be `coauthorsavatars`.

mostafaabd: I do all changes, and there is no option to hide `avatar_url ` in a text field because this plugin use [ Jquery suggest](https://web.archive.org/web/20111017233444/http://plugins.jquery.com/project/suggest) that not support to render output like jQuery Autocomplete.

Issue #625, opened 2018-11-26
Custom query for Author results in a WordPress database error


hyperionjrw:
With a custom query for an Author's posts WordPress will throw a DB error due to malformed SQL.

Before your post_where filter runs the query looks like this:
`' AND   wp_posts.post_author IN (10202675)  AND wp_posts.post_type = \'post\' AND ((wp_posts.post_status = \'publish\'))'`

After it looks like this:
`' AND ( wp_posts.post_author IN (10202 OR (wp_term_taxonomy.taxonomy = \'author\' AND wp_term_taxonomy.term_id = \'48333\'))675)  AND wp_posts.post_type = \'post\' AND ((wp_posts.post_status = \'publish\'))'`

You'll notice that the user ID goes from 10202675 to 10202--SQL--675

The problem is down to a bug in the posts_where_filter method when setting $id.
If it's not an author page it is set to '\d+' to look for digits when passed to the regex but before then that $id is passed to get_user_by which will fail and then the value is set to 0. That 0 is then passed into the regex and ends up splitting longer user_ids at one position after the last zero. 

regex: `post_author\s*IN\s*(.*0(?:.*private\')?.)`

So:
`$id = is_author() ? get_queried_object_id() : '\d+';`
`if ( false === get_user_by( 'id', $id ) ){`
`	$id = 0;`
`}`

Needs to become:
`$id = is_author() ? get_queried_object_id() : '\d+';`
`if ( false === get_user_by( 'id', $id ) ){`
`	$id = '\d+';`
`}`

Although I think there will still be a problem if you try to query a different author's posts on another author's page as the id will always come from the page rather than the custom query.

rebeccahum:
@hyperionjrw Do you have any steps on how to reproduce the issue?

hyperionjrw:
With a clean WordPress install do the following:

- Run this SQL to make sure the next user created has an ID that'll cause a problem:
    `ALTER TABLE wp_users AUTO_INCREMENT = 9099999999;`
- Create a new users
- Create a post with that user account or change an existing post over to that user. 
    We basically need the id in the post_author column.
- Activate the plugin
- Make sure to add this user as a secondary user on a post so that terms are created. 
- Run the following query in index.php replacing the author ID with your new one.
    `$x = new WP_Query( array (
        'post_status' => 'publish',
        'author' => 9099999999
    ) );`

[...]

PR #628, opened 2018-11-27
Fix for issue #625

hyperionjrw:
A previous change #624 broke the regEx which then broke the SQL see #625 for full details.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -660,13 +660,14 @@ class CoAuthors_Plus {
      * @param string $where
      * @param WP_Query $query
      *
-     * @return string|string[]|null
+     * @return string
      */
 	function posts_where_filter( $where, $query ) {
 		global $wpdb;
 
 		if ( $query->is_author() ) {
 			$post_type = $query->query_vars['post_type'];
+			if ( 'any' === $post_type ) {
 				$post_type = get_post_types( array( 'exclude_from_search' => false ) );
 			}
 
@@ -717,14 +718,16 @@ class CoAuthors_Plus {
 				}
 				$terms_implode = rtrim( $terms_implode, ' OR' );
 
-				$id = is_author() && $query->is_main_query() ? get_queried_object_id() : '\d+';
+				// We need to check the query is the main query as a new query object would result in the wrong ID
+				$id = $query->is_main_query() && is_author() ? get_queried_object_id() : '\d+';
 
 				//If we have an ID but it's not a "real" ID that means that this isn't the first time the filter has fired and the object_id has already been replaced by a previous run of this filter. We therefore need to replace the 0
 				// This happens when wp_query::get_posts() is run multiple times.
+				// If previous condition resulted in this being a string there's no point wasting a db query looking for a user.
 				if ( $id !== '\d+' && false === get_user_by( 'id', $id ) ){
 					$id = '\d+';
 				}
-				
+
 				// When WordPress generates query as 'post_author IN (id)'.
 				if ( false !== strpos( $where, "{$wpdb->posts}.post_author IN " ) ) {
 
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -654,15 +654,19 @@ function posts_join_filter( $join, $query ) {
...
+     * @param WP_Query $query
+     *
+     * @return string|string[]|null
+     */
 	function posts_where_filter( $where, $query ) {
 		global $wpdb;
 
 		if ( $query->is_author() ) {
 			$post_type = $query->query_vars['post_type'];
-			if ( 'any' === $post_type ) {
```
rebeccahum: Hi there, I'm wondering why this line is being removed?

co-authors-plus.php
```diff
@@ -713,12 +717,12 @@ function posts_where_filter( $where, $query ) {
 				}
 				$terms_implode = rtrim( $terms_implode, ' OR' );
 
-				$id = is_author() ? get_queried_object_id() : '\d+';
+				$id = is_author() && $query->is_main_query() ? get_queried_object_id() : '\d+';
```
rebeccahum: I don't think we need to add the check for `$query->is_main_query()`

co-authors-plus.php
```diff
@@ -713,12 +717,12 @@ function posts_where_filter( $where, $query ) {
...
 
 				//If we have an ID but it's not a "real" ID that means that this isn't the first time the filter has fired and the object_id has already been replaced by a previous run of this filter. We therefore need to replace the 0
 				// This happens when wp_query::get_posts() is run multiple times.
-				if ( false === get_user_by( 'id', $id ) ){
-					$id = 0;
+				if ( $id !== '\d+' && false === get_user_by( 'id', $id ) ){
```
rebeccahum: The check for `$id !== '\d+'` is also not necessary.  I think we can just have it assigned to `'\d+'` if `get_user_by( 'id', $id )` returns `false` since `$id` will not equal `'\d+'` if `$id` ends up equating to `get_queried_object_id()` which will not be `\d+`.

co-authors-plus.php
rebeccahum: How come we are removing this line here?  This will cause a syntax error and throw off the support for `'any'` post types.

co-authors-plus.php
rebeccahum: The check for `$id !== '\d+'` is probably not necessary. I think we can just have `$id` assigned to `'\d+'` if `get_user_by( 'id', $id )` returns `false` since `$id` will not equal `'\d+'` if `$id` ends up equating to `get_queried_object_id()` (which will not be `'\d+'`).

co-authors-plus.php
rebeccahum: This fix works without the check for `$query->is_main_query()` so I think we don't need to add it for now.

co-authors-plus.php
hyperionjrw: The check for the string was to avoid doing an unnecessary user query that's obviously going to fail. String check is far quicker that a db query.

co-authors-plus.php
hyperionjrw: I didn't think I was. 🤔
I think I may have messed up when selecting which bits to commit as I had a heap of white space changes show up due to my automatic tendency to run a code formatting shortcut after almost all edits and also doing this between other things.

co-authors-plus.php
hyperionjrw: If you query a second user from author.php to show something like, people in same dept or other authors you may like, you won't get the correct posts, you'll only get those where the user is in the post_author column.

e.g. If I add `$x = new WP_Query( array ( 'post_status' => 'publish', 'author' => 9099999999 ) );` to author.php after the main loop. Go to `/?author=1` the ID that's passed to the regex is **1** not **9099999999** so our custom query is missing any posts where the queried user isn't in the `post_author` column.

hyperionjrw: Thinking this through you also need this check to avoid further risk of WordPress throwing db errors. If your custom queried user ID contains the author page user ID somewhere before the 3rd char from the end then we'd end up with a split user ID and invalid SQL. 

e.g Go to /?author=1 and sub-query for author 10123456. That'd produce regex that's looking through the the SQL for `\s*(.*1.)/` which would match 1012 from our ID and put the new SQL in the middle of the ID and thus a db error would be thrown.

Issue #614, opened 2018-10-15
Gutenberg does not use the remove_meta_box command


jonathanstegall:
I was testing Gutenberg a bit today and I noticed that the default author dropdown appears in the Document panel even when Co-Authors-Plus is active.

Of course currently [this command](https://github.com/Automattic/Co-Authors-Plus/blob/master/co-authors-plus.php#L325) removes the box from the classic editor.

I've been looking around for what the equivalent is in Gutenberg. Most of the support threads I've seen end up pointing me to this [GitHub issue](https://github.com/WordPress/gutenberg/issues/6533), which started out specific to featured images, but also mentions others (such as in [this comment](https://github.com/WordPress/gutenberg/issues/6533#issuecomment-401137263)).

While this issue is not closed in Gutenberg, I thought it would be good to flag it here in case they come up with a usable solution before launch that can be implemented here.

jonathanstegall:
I tested 4.3 of Gutenberg and this issue was still present. But then I tried the WordPress 5.0 Beta 4 that was released today, and it appears to have been resolved.

The [release post](https://wordpress.org/news/2018/11/wordpress-5-0-beta-4/) has this info:

> Additionally, there have been some bugs fixed that popped up in beta 3:
> 
> - Better support for plugins that have more advanced meta box usage.
> - Script concatenation is now supported.
> - Ajax calls could occasionally cause PHP errors.

Presumably the advanced meta box usage is what fixed this issue.

jonathanstegall:
I tested 4.3 of Gutenberg and this issue was still present. When I tried the WordPress 5.0 Beta 4 that was released today without the Gutenberg plugin the issue was still present. But then, when I activate the Gutenberg plugin (still version 4.3), this issue appears to have been resolved.

The [release post](https://wordpress.org/news/2018/11/wordpress-5-0-beta-4/) has this info:

> Additionally, there have been some bugs fixed that popped up in beta 3:
> 
> - Better support for plugins that have more advanced meta box usage.
> - Script concatenation is now supported.
> - Ajax calls could occasionally cause PHP errors.

Presumably the advanced meta box usage is what fixed this issue, but only if the Gutenberg plugin is used instead of just Core.

jonathanstegall:
I saw a [recent update](https://github.com/WordPress/gutenberg/issues/6533#issuecomment-441816757) in Gutenberg about this:

> To my knowledge, this Author field in Gutenberg can't be disabled:
> 
[...]

PR #630, opened 2018-12-07
Use CSS to hide the Block Editor's Author label + select

mjangda:
This is a pretty big hack :)

With 5.0, we end up with both Core's author input and the Co-Authors Plus metabox which is going to be cause for confusion (see #614).

It would be better to remove the inputs completely rather than hiding them (via javascript or hooking directly into Block Editor [1]) but this is a good start to help reduce that confusion.

[1] May need some new hooks in the editor to make this possible easily (i.e. without hacks :)).

### DIFF
--- a/css/co-authors-plus.css
+++ b/css/co-authors-plus.css
@@ -109,3 +109,9 @@
 			margin-left: 30px;
 			font-size: 13px;
 		}
+
+/** Block Editor Hack for 5.0: Hide the core author input **/
+.block-editor label[for^="post-author-selector-"],
+.block-editor select[id^="post-author-selector-"] {
+	display: none;
+}


mjangda:
Even better fix in the future would be to replace the core input with the co-authors plus input and have it nicely integrated with the Block Editor.

scofennell:
This CSS no longer targets the post_author selector.

`/** Block Editor Hack for 5.0: Hide the core author input **/
.block-editor label[for^="post-author-selector-"],
.block-editor select[id^="post-author-selector-"] {
	display: none;
}
`

The markup in Gutenberg has changed:

Screen Shot 2021-05-20 at 10 36 44 AM

PR #632, opened 2018-12-07
Prepare 3.3.1

mjangda:
Cherrypicks #630 and bumps readme + versions.

We're doing a small focused release bump since there have been a significant number of changes since 3.3.0 and we need to do some extensive testing before releasing them all.

With this release we get basic 5.0 compat.


mjangda:
Note: Need to merge the readme and version changes to master after.

PR #653, opened 2019-03-11
VIP: Enable co-authors plus filter that disabled the querying by post_author on VIP Go

shantanu2704:
This speeds up the SQL query. This filter has been enabled on WordPress.com VIP. This filter is active on WordPress.com VIP but not on VIP Go.

We'd need to run the backfill CLI before this can be merged in.


GaryJones:
@shantanu2704 Is this still needed for VIP?

shantanu2704:
I don't know, @GaryJones :(
I'm looking back at VIP Intern Shantanu in disappointment for not including any further details.
Thanks to GitHub, [here's some more context ](https://github.com/Automattic/vip-issues/issues/366#issuecomment-457293087)around why we considered adding this filter.

GaryJones:
Adding to a future milestone to investigate later on. @shantanu2704 Do you fancy checking to see if the query by post_author is coming up as slow / needed any more?

GaryJones:
Thank you for this, @shantanu2704, and apologies that it sat open for so long.

The behaviour you were after — being able to disable querying by `post_author` on VIP — is now available on `develop` through the `coauthors_plus_should_query_post_author` filter, which gates exactly that path. Between that and several years of changes to the query and helper layout, this branch has drifted well past a clean rebase (it also bundles some VIP/Travis/README artefacts that no longer fit the current repo).

Rather than leave it lingering, we're going to close it out — but the underlying need is covered by the filter on develop. Thanks again for the contribution; it helped inform where we ended up.

shantanu2704:
Thanks for closing the loop on this, @GaryJones !

PR #659, opened 2019-03-18
Fix author archive title and description

### LINKED ISSUE #643: Wrong user description showing on Author archive
It seems that the author description that's being shown on the author archive is generated by pulling the author of the first post in the result set. However, if the first post contains co-authors, and the author in question is NOT the primary author of the first post, the description appears wrong.

Using the following filter works around the issue:

function mp_author_description_filter( $description ) {

	if ( ( is_author() ) && ( function_exists( 'coauthors_posts_links' ) ) ) {

		$a = get_user_by( 'slug', get_query_var( 'author_name' ) );
		if ( $a ) {
			return $a->description;
		}

	}
	
	return $description;
} 
add_filter( 'get_the_archive_description', 'mp_author_description_filter' );

### LINKED ISSUE #658: Author archive title and description is shown wrong
Since author archive title is fixed using `get_the_archive_title` filter, it gets overridden by twentynineteen which uses the same filter. So the correction filter in Co-Authors-Plus needs to be delayed.

Similarly, the archive description is shown wrong if on an author archive, the first story has co-authors and the first author is NOT the same as the author for the archive. This needs to be fixed by `get_the_archive_description` filter.

#643 is a bit similar

manzoorwanijk:
Fixes #643 and #658


robmaurizi:
Is this slated for a specific release yet?

rebeccahum:
@manzoorwanijk Can you please close and re-open this PR? We needed to rewrite history.

PR #661, opened 2019-03-19
#651 fix sql where clause when custom post status exist

### LINKED ISSUE #651: The posts_where regex causes invalid SQL when there are custom post statuses
This is related to https://github.com/Automattic/Co-Authors-Plus/issues/508 and caused by https://github.com/Automattic/Co-Authors-Plus/pull/576/.

 ## Problem
When there is a custom post status in the query ([such as the one added by bbPress here](https://github.com/wp-plugins/bbpress/blob/master/bbpress.php#L618-L627)), it can make the current regex do unintended things on this line: https://github.com/Automattic/Co-Authors-Plus/blob/b3f34ce6f989f26d25dcad729767bf72be893677/co-authors-plus.php#L740

Example final query after our filter:

```
SELECT SQL_CALC_FOUND_ROWS wp_posts.ID
FROM wp_posts 
LEFT JOIN wp_term_relationships AS tr1
ON (wp_posts.ID = tr1.object_id)
LEFT JOIN wp_term_taxonomy
ON ( tr1.term_taxonomy_id = wp_term_taxonomy.term_taxonomy_id )
WHERE 1=1 
AND ( (wp_term_taxonomy.taxonomy = 'author'
AND wp_term_taxonomy.term_id = '102193'))OR wp_posts.post_status = 'hidden')
GROUP BY wp_posts.ID
HAVING MAX( IF ( wp_term_taxonomy.taxonomy = 'author', IF ( wp_term_taxonomy.term_id = '102193',2,1 ),0 ) ) <> 1 
ORDER BY wp_posts.post_date DESC
LIMIT 0, 20;
```

Note the `OR wp_posts.post_status = 'hidden')` which makes this SQL invalid.

Before the regex replace, the where clause looks like this:

```
" AND wp_posts.post_author IN (456)  AND wp_posts.post_type = 'post' AND (wp_posts.post_status = 'publish' OR wp_posts.post_status = 'closed' OR wp_posts.post_status = 'future' OR wp_posts.post_status = 'draft' OR wp_posts.post_status = 'pending' OR wp_posts.post_status = 'private' OR wp_posts.post_status = 'hidden')"
```

After:

```
" AND ( (wp_term_taxonomy.taxonomy = 'author' AND wp_term_taxonomy.term_id = '102193'))OR wp_posts.post_status = 'hidden')"
```

## Solution

So as you can see, the flaw is assuming that `private` will be the last status in the list. It's not always going to be, as seen here when using bbPress.

We could either be more flexible with the regex and target the last `')`, or perhaps get all of the post_status_types from WP that will be added here and target each of them.

Context: 89781-zd.

davidsword:
Fixes #651 

If you have a custom `post_status` on your site, the following regex pattern will not work as intended:

https://github.com/Automattic/Co-Authors-Plus/blob/4e1b2d4c80bf28169bef5d8a1c77e95489acf042/co-authors-plus.php#L740

I've beautified the snippets, but our `$where` query originally enters the function looking like

```sql
AND
wp_posts.post_author IN (3)
AND
wp_posts.post_type = 'post'
AND
(
    wp_posts.post_status = 'publish'
    OR wp_posts.post_status = 'closed'
    OR wp_posts.post_status = 'future'
    OR wp_posts.post_status = 'draft'
    OR wp_posts.post_status = 'pending'
    OR wp_posts.post_status = 'private'
    OR wp_posts.post_status = 'hidden'
)
```

Currently, with an extra `post_status` of `hidden`, this regex replace changes the SQL query into:

```sql
AND 
( 
    wp_posts.post_author IN ( 3 )
    AND wp_posts.post_type = 'post'
    AND 
    ( 
        wp_posts.post_status = 'publish'
        OR wp_posts.post_status = 'closed'
        OR wp_posts.post_status = 'future'
        OR wp_posts.post_status = 'draft'
        OR wp_posts.post_status = 'pending'
        OR wp_posts.post_status = 'private'
        OR 
        ( 
            wp_term_taxonomy.taxonomy = 'author'
            AND wp_term_taxonomy.term_id = '200' 
        )
        OR 
        ( 
            wp_term_taxonomy.taxonomy = 'author'
            AND wp_term_taxonomy.term_id = '197' 
        ) 
    )
    OR wp_posts.post_status = 'hidden' 
)  
```
This PR removes the need for `private` and just looks for the first occurrence of a closing parentheses `')`, which gives us a `$where` value of:

```sql
AND 
(
    wp_posts.post_author IN ( 3 )
    AND wp_posts.post_type = 'post'
    AND 
    ( 
        wp_posts.post_status = 'publish'
        OR wp_posts.post_status = 'closed'
        OR wp_posts.post_status = 'future'
        OR wp_posts.post_status = 'draft'
        OR wp_posts.post_status = 'pending'
        OR wp_posts.post_status = 'private'
        OR wp_posts.post_status = 'hidden' 
    )
    OR 
    ( 
        wp_term_taxonomy.taxonomy = 'author'
        AND wp_term_taxonomy.term_id = '200' 
    )
    OR 
    ( 
        wp_term_taxonomy.taxonomy = 'author'
        AND wp_term_taxonomy.term_id = '197' 
    )
)  
```

---

To replicate this, in `/wp-admin/edit.php` make a request for `?author=<id>`. This fires the `post_author IN` where clause (as per: https://github.com/WordPress/WordPress/blob/04dff0e2a969228bc80d9b1961136eef4edc583b/wp-includes/class-wp-query.php#L2180-L2196 ). You will also need a custom `post_status` registered by either creating one, or using a plugin like bbpress (which adds `hidden`).

Note that I wasn't able to replicate the symptoms of the problem itself (no results), but I was able to confirm the above differences in the `$where` clause.

#508 related.
/cc @rebeccahum


Venutius:
An example of a plugin that uses a custom post status is https://wordpress.org/plugins/lh-logged-in-post-status/ My own plugin https://wordpress.org/plugins/bp-post-status/ also users custom post statuses and I can confirm my author based posts templates fail with Co-Authors plus loaded with SQL error. I think either of these plugins should be good to test this.

WPprodigy:
Just tested this with the replication steps in https://github.com/Automattic/Co-Authors-Plus/issues/651#issuecomment-475397392.

It does solve the problem. Though perhaps worth noting that in 3.3.1, just one post would show up. After updating this line, trashed posts also showed up in the `mine` column. Unsure if this is intended, but seems like different behavior from the last release.

rebeccahum:
Closing in favour of #663.

PR #664, opened 2019-03-22
Performance improvements for assign_coauthors command

mikeyarce:
We’re doing a few things here to improve the performance of assign_coauthors:

1. `'no_found_rows'	   => true,`  Since we don’t care about pagination here, we can dramatically speed up the query by adding this argument.  This removes the intensive SQL `SQL_CALC_FOUND_ROWS` query modifier.

2. We also are removing the `posts_groupby` argument because we don’t need to actually group by Post ID.  You should only need to group by Post ID when you have multiple results that could have the same Post ID (like if you had multiple meta keys with the same Post ID).

3. We’re doing some deferring of term counting here too, and basically deferring the coutning until AFTER each foreach runs, so that we’re not doing term counting on each individual item in the loop.

In some use cases, this sped up the command from taking a few weeks of time to a few hours.


mikeyarce:
This PR got messy, going to make a new one

PR #670, opened 2019-04-15
Refreshing meta box nonce after re-authenticating following a wp_auth_check()

### LINKED ISSUE #669: Refresh meta box nonce after re-authenticating to prevent data loss
Users are currently losing data when they try to save their post after re-authenticating due to their session expiring.

Once the user has re-authenticated via the wp_auth_check modal everything appears to be okay until they save the post, at which point they are greeted with the 'Link has expired' `wp_die()` screen, due to the meta box nonce no longer being valid. This causes the user to lose any unsaved data.

The fix is to refresh the nonce via the `wp_refresh_nonces` filter. I will open a PR that addresses this issue.

hrkhal:
Fixes #669 

Filters `wp_refresh_nonces` to refresh the coauthors-edit nonce to prevent data loss after re-authenticating an expired session.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -103,6 +103,9 @@ class CoAuthors_Plus {
 		add_action( 'add_meta_boxes', array( $this, 'add_coauthors_box' ) );
 		add_action( 'add_meta_boxes', array( $this, 'remove_authors_box' ) );
 
+		// Refresh the nonce after the user re-authenticates due to a wp_auth_check() to avoid failing check_admin_referrer()
+		add_action( 'wp_refresh_nonces', array( $this, 'refresh_coauthors_nonce' ), 20, 1 );
+
 		// Removes the co-author dropdown from the post quick edit
 		add_action( 'admin_head', array( $this, 'remove_quick_edit_authors_box' ) );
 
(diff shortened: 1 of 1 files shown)


rebeccahum:
Thanks for the PR -- can we get some steps on how to reproduce the issue?

hrkhal:
@rebeccahum To reproduce the issue you'll need an install with custom fields and meta boxes in the post editor and just the co-authors-plus plugin activated. Then:

1. Open up the post edtior and add some content and fill in custom meta Save as normal.
2. Open up wp-admin in a new tab and log out of wp-admin from it.
3. Return to post editor tab and wait for the auth check modal and then reauthenticate by logging in
4. Update the content and edit the custom meta.
5. Now attempt to save the post and you'll hit the `Link has expired` screen instead of successful post save.
6. Follow the `please try again` link and you’ll notice that post content has likely saved (due to autosave) but all the custom meta changes have been lost.

We get the `Link has expired` die screen because the `check_admin_referrer` check in co-authors-plus fails due to the expired nonce that wasn’t refreshed after re-authenticating. Thus any save_post actions after never fired and thus data lost.

By implementing this PR the main co-authors nonce is refreshed after re-authenticating which will prevent the data loss described above.

You may also notice that at step 4, you’ll be unable to search for new authors to add via the co-authors-plus meta box. This is because the nonce stored in the JS var `coAuthorsPlus_ajax_suggest_link` is also now invalid because of the re-authenticating. Due to the way it's stored `wp_refresh_nonces` doesn't fix it. This is a seperate issue I'll likely open an issue/pr for.

GaryJones (approved):
✅  Wonderful recreation instructions @hrkhal - much appreciated!

With the Classic Editor plugin active in WP 6.2.2, I was able to recreate this (though after "Please try again" the post meta did seem to be saved); the patch worked to remove the Link has expired die screen.

With the block editor in WP 6.2.2, this doesn't seem to be an issue. With other plugins activated, I got an "Updating failed" while following the procedure, but this persisted with the PR applied as well. On `master` and with this patch, after logging back in and editing content and post meta, the save worked correctly.

PR #671, opened 2019-04-18
Remove PHP notice when viewing guest author archive

tkiehne:
When viewing an author archive that is not tied to an actual Wordpress user (i.e.: a guest author), PHP throws a Notice in filter_count_user_posts().  The $user object doesn't exist but the function tries to retrieve data from the object anyway.

This change checks for a user before continuing.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1037,14 +1037,15 @@ class CoAuthors_Plus {
 	 */
 	function filter_count_user_posts( $count, $user_id ) {
 		$user = get_userdata( $user_id );
-		$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
+		if ( $user ) {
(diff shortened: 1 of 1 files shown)


rebeccahum:
@tkiehne Can we get a copy of the notice?

tkiehne:
> Notice:  Trying to get property of non-object in /var/www/html/web/app/plugins/co-authors-plus/co-authors-plus.php on line 1041

lubieowoce:
getting the same thing, triggered by Yoast's "only show authors who have >=1 posts in search results" setting (obv. because it's counting user posts)

nielslange:
Hello @tkiehne 👋 Can you still reproduce this problem on your end?

GaryJones:
Closing for lack of response.

GaryJones:
Closing for lack of feedback.

PR #674, opened 2019-05-01
Use accurate counts for both users and guest authors

### LINKED ISSUE #228: Recent changes to bylines/author assignments don't get reflected properly in post counts in admin UI

rebeccahum:
This resolves #228.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1042,7 +1042,12 @@ class CoAuthors_Plus {
 		$term = $this->get_author_term( $user );
 
 		if ( $term && ! is_wp_error( $term ) ) {
-			$count = $term->count;
+			if ( 'guest-author' === $user->type ) {
+				// If using guest author term count, add on linked user count. 
+				$count = (int) $count + $term->count;
+			} else {
+				$count = $term->count;
+			}
 		}
 
 		return $count;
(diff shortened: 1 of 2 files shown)


WPprodigy:
Hmm, so how should this work exactly?

In the current master branch, after following the steps in https://github.com/Automattic/Co-Authors-Plus/issues/228, both the Guest Author and the User show 0 posts.

After pulling this change, my Guest Author still shows 0 and the User's count comes back. Is that intended, or should they be synced since they are linked?

rebeccahum:
@WPprodigy Should be `1` for both linked accounts since we are taking the total.

Exact steps to reproduce:
1) Create user 'A'
2) Sign on from user 'A' and make a post
3) Create guest author 'B'
4) Link guest author 'B' to user 'A'
5) Go to both the User and Guest Author listing and you should see a total of `1` posts for both

WPprodigy:
Ah, found the problem. I cheated on step 1 and didn't create a new user. Since my user existed since before CAP, it didn't have a term. After running `wp co-authors-plus create-terms-for-posts` it started working correctly.

That said, I think it should be more prepared for this possible behavior. Otherwise, you also get this error:

> PHP Notice:  Trying to get property 'count' of non-object in /Co-Authors-Plus/php/class-coauthors-wp-list-table.php on line 260

PR #678, opened 2019-05-24
#676 - Avatar Bug

### LINKED ISSUE #676: Avatar images are incorrect in the admin
After updating to the latest version, all of the avatars that display in the admin are incorrect. After some troubleshooting it appears the images are being pulling from the media library instead. For example, user's avatar with ID 100 is using the media image that was uploaded to post 100. This does not affect the front-end, just the admin. This is for all instances of the avatar displayed in the admin, including the top admin tool bar and user edit screens.

I am unable to replicate this outside of the production and staging environments. I've tried replicating using a different WP installation using the same theme and same set of plugins on my local environment.

Commenting out this line in the plugin appears to fix the issue https://github.com/Automattic/Co-Authors-Plus/blob/master/co-authors-plus.php#L1726

jasonbahl:
This fixes a bug where the Avatar images are being replaced with the post's featured image instead of the Author's featured image

Closes #676

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1722,9 +1722,17 @@ class CoAuthors_Plus {
 	 * @return string Avatar URL
 	 */
 	public function filter_get_avatar_url( $url, $id ) {
-		if ( has_post_thumbnail( $id ) ) {
+
+		$guest_author = true;
+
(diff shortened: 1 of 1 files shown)


jasonbahl:
Looks like this was causing some fatal errors. Closing this until we have a better solution

PR #685, opened 2019-07-08
alllow adding of coauthors by ID

paulschreiber:


paulschreiber:
@whyisjake can we take this PR, or suggest an equivalent fix?

PR #689, opened 2019-07-11
Fixed array syntax for older versions of PHP.

samyerkes:
This was causing a 500 error for a site running an older version of PHP.


nielslange:
Hello @samyerkes 👋 Thanks for your contribution. While it's great to remain backwards compatible, the short array syntax had been [introduced with PHP 5.4](https://www.php.net/releases/5_4_0.php). WordPress officially [recommends using PHP 7.4 or greater](url). The minimum supported version is currently PHP 5.6.20, which already supports the PHP short syntax. Therefore, I'd tend not to change the code towards outdated PHP versions. 

@rebeccahum Can I have you second opinion in this case?

rebeccahum:
Agreed with @nielslange. Closing.

PR #690, opened 2019-07-16
Fix wrong avatar url getting returned when a users id matches a post id

ivanlopez:
This fixes a bug where the wrong image was getting returned for WordPress users that had the same ID as post ID. If the post had a featured image then it was getting returned as the user's avatar. This adds an additional check to make sure that the ID belongs to a post with a post type of `guest-author`.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1722,7 +1722,7 @@ class CoAuthors_Plus {
 	 * @return string Avatar URL
 	 */
 	public function filter_get_avatar_url( $url, $id ) {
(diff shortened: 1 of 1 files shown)


davidvexel:
Yeah but what if the user id matches an post id with type guest author? the author could not belong to that user.

GaryJones:
This will need conflicts fixing - CAP has changed what filter the callback is hooked to, for instance. 

I'm closing it out, but feel free to refresh it if the wrong avatar URL is being used.

cc @katag9k

PR #698, opened 2019-08-05
WIP: Specify user by ID in assign-user-to-coauthor

Floppy:
If a user has already been deleted, we need to specify just the user ID when assigning coauthors, instead of the login. This PR adds a second command-line option. One of either `--user_id` or `--user_login` is required. Existing behaviour should be unchanged.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -210,22 +210,28 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	 * @since 3.0
 	 *
 	 * @subcommand assign-user-to-coauthor
-	 * @synopsis --user_login=<user-login> --coauthor=<coauthor>
+	 * @synopsis --user_login=<user-login> --user_id=<user-id> --coauthor=<coauthor>
 	 */
 	public function assign_user_to_coauthor( $args, $assoc_args ) {
 		global $coauthors_plus, $wpdb;
 
 		$defaults = array(
-				'user_login'        => '',
+				'user_login'        => NULL,
+				'user_id'           => NULL,
 				'coauthor'          => '',
 			);
 		$assoc_args = wp_parse_args( $assoc_args, $defaults );
 
-		$user = get_user_by( 'login', $assoc_args['user_login'] );
+		$user_id = $assoc_args['user_login'];
+		if ( $assoc_args['user_login'] ) {
+			$user = get_user_by( 'login', $assoc_args['user_login'] );
+			$user_id = $user->ID;
+		}
+
 		$coauthor = $coauthors_plus->get_coauthor_by( 'login', $assoc_args['coauthor'] );
 
-		if ( ! $user ) {
-			WP_CLI::error( __( 'Please specify a valid user_login', 'co-authors-plus' ) );
+		if ( ! $user_id ) {
+			WP_CLI::error( __( 'Please specify a valid user_login or user_id', 'co-authors-plus' ) );
 		}
 
 		if ( ! $coauthor ) {
@@ -233,7 +239,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 		}
 
 		$post_types = implode( "','", $coauthors_plus->supported_post_types );
-		$posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('$post_types')", $user->ID ) );
+		$posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('$post_types')", $user_id ) );
 		$affected = 0;
 		foreach ( $posts as $post_id ) {
 			$coauthors = cap_get_coauthor_terms_for_post( $post_id );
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -210,30 +210,36 @@ public function assign_coauthors( $args, $assoc_args ) {
...
 		$defaults = array(
-				'user_login'        => '',
+				'user_login'        => NULL,
+				'user_id'           => NULL,
 				'coauthor'          => '',
 			);
 		$assoc_args = wp_parse_args( $assoc_args, $defaults );
 
-		$user = get_user_by( 'login', $assoc_args['user_login'] );
+		$user_id = $assoc_args['user_login'];
```
RobjS: Is this meant to be:
```
$user_id = $assoc_args['user_id']
```

Currently it doesn't look like a specified user_id actually gets used anywhere.

Floppy: Yes, it is. 🙄

Floppy:
```suggestion
		$user_id = $assoc_args['user_id'];
```

php/class-wp-cli.php
```diff
@@ -210,30 +210,36 @@ public function assign_coauthors( $args, $assoc_args ) {
 	 * @since 3.0
 	 *
 	 * @subcommand assign-user-to-coauthor
-	 * @synopsis --user_login=<user-login> --coauthor=<coauthor>
+	 * @synopsis --user_login=<user-login> --user_id=<user-id> --coauthor=<coauthor>
```
Floppy: I'm not at all sure how to specify that *one* of `user_login` or `user_id` is required. Anyone else?

php/class-wp-cli.php
```diff
@@ -210,30 +210,36 @@ public function assign_coauthors( $args, $assoc_args ) {
...
+		if ( $assoc_args['user_login'] ) {
+			$user = get_user_by( 'login', $assoc_args['user_login'] );
+			$user_id = $user->ID;
+		}
+
 		$coauthor = $coauthors_plus->get_coauthor_by( 'login', $assoc_args['coauthor'] );
 
-		if ( ! $user ) {
-			WP_CLI::error( __( 'Please specify a valid user_login', 'co-authors-plus' ) );
+		if ( ! $user_id ) {
```
RobjS: It might be worth checking if the `$user_id` value is an int at this point, and throwing a relevant WP_CLI error if not. Otherwise, the user could pass a non-int value in `--user_id`, and the `$wpdb->prepare()` call in l.242 could throw a non-user-friendly error when it tries to treat it as an int.

RobjS: Actually, perhaps just checking that the specified user_id corresponds to an existing user would work?

php/class-wp-cli.php
```diff
@@ -210,30 +210,36 @@ public function assign_coauthors( $args, $assoc_args ) {
...
 		$defaults = array(
-				'user_login'        => '',
+				'user_login'        => NULL,
+				'user_id'           => NULL,
 				'coauthor'          => '',
 			);
 		$assoc_args = wp_parse_args( $assoc_args, $defaults );
 
-		$user = get_user_by( 'login', $assoc_args['user_login'] );
+		$user_id = $assoc_args['user_id'];
```
Floppy: @RobJS I think the sensible way to check for int user IDs is to parse it properly here, which I probably needed to do anyway.

```suggestion
		$user_id = intval($assoc_args['user_id']);
```

php/class-wp-cli.php
```diff
@@ -210,30 +210,36 @@ public function assign_coauthors( $args, $assoc_args ) {
...
+		if ( $assoc_args['user_login'] ) {
+			$user = get_user_by( 'login', $assoc_args['user_login'] );
+			$user_id = $user->ID;
```
RobjS: I think this'll fail if the `user_login` doesn't correspond to an actualy user, because `get_user_by()` will just return false.

php/class-wp-cli.php
Floppy: @RobjS I've added a suggested change to coerce it earlier on, which should fix this.

php/class-wp-cli.php
Floppy: Ah yes, this would have been guarded against before.
```suggestion
			if ($user) {
				$user_id = $user->ID;
			}
```

php/class-wp-cli.php
Floppy: Can't check that it's an existing user, as this is specifically for use in the case where the user with that ID doesn't exist any more in the database.

GaryJones:
Can you please refresh and fix the conflicts?

GaryJones:
Hi @Floppy — wanted to leave a note here for posterity.

The use case you raised in this PR (assigning posts to a co-author when the original WordPress user has been deleted, so a `--user_login` lookup is impossible) was a real and valid gap, and the design you landed on — accepting `--user_id` as an alternative — was the right call.

I picked up the thread in #1266, rebased against current `develop` (which has since gained `--append_coauthors`), incorporated the review feedback from yourself, @RobjS and @mallorydxw, and added Behat coverage for the validation paths and the orphan-post happy path. That has now been merged.

Thanks to all three of you for the original work and review — none of the design decisions in #1266 are mine; I just carried the baton over the line. Ship credit belongs upstream.

Floppy:
Nice, one thanks for letting me know! I'd completely forgotten about this PR, nice to see it fulfilling its destiny eventually ❤️

Issue #722, opened 2020-02-18
Add normal users?


strarsis:
Is it possible to add "normal" WordPress users as author of a post?
The search suggestion doesn't seem to search for normal users but rather in the `author` taxonomy,
hence it won't find normal WordPress users.
The existing owner of the post who is a normal WordPress user, is already listed there, 
but can't be found by the search either.

rebeccahum:
> Is it possible to add "normal" WordPress users as author of a post?

Yes, this should be possible.  Are you able to provide any further details on how to reproduce on a vanilla install?

strarsis:
@rebeccahum: Well, I have to create a guest user per normal WordPress user, also with different username. This can become very complicated.

rebeccahum:
I'm thinking that you may have to run a backfill on authors created before CAP was installed to create their respective terms: https://github.com/Automattic/Co-Authors-Plus/blob/221e4df88d65b1ba2eb4467d16e4ba74d82f37c9/php/class-wp-cli.php#L53

`wp co-authors-plus create-terms-for-posts`

PR #731, opened 2020-04-23
Change permission

bhsjacket:
For many newsrooms, including ours, we have users that we don't want to have edit access. For example, our photographers have accounts because they use our backend to submit their photos. Sometimes though, we want to assign them as the author of a post. For example, if they make a photo essay.

I don't see any ready for why the user has to be able to edit posts. The guest author system is nice, but it's just not very convenient to have a custom post type that acts like a user.

It seems like users have had this issue before:
- [Bram Peerlings on WordPress.org](https://wordpress.org/support/topic/allow-subscribers-as-authors/)
- [strarsis on GitHub](https://github.com/Automattic/Co-Authors-Plus/issues/722)

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1270,7 +1270,7 @@ class CoAuthors_Plus {
 			// Make sure the user is contributor and above (or a custom cap)
 			if ( in_array( $found_user->user_nicename, $ignored_authors, true ) ) { // AJAX sends a list of already present *users_nicenames*
 				unset( $found_users[ $key ] );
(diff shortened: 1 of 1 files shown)


rebeccahum:
@bhsjacket Is there any reason why you don't use the filter for changing the capabilities?

bhsjacket:
> @bhsjacket Is there any reason why you don't use the filter for changing the capabilities?

I just think that it should be the default behavior. You can close it if you disagree.

For anyone who finds this in the future, here's the code I used:
```php
function cap_callback() {
    return 'read';
}
add_filter('coauthors_edit_author_cap', 'cap_callback');
```

rebeccahum:
Thanks for the PR, but I think it would be better to keep the permissions as-is to not break any sites that currently rely on the `edit_posts` permission.

PR #744, opened 2020-09-09
Search by term name rather than slug due

### LINKED ISSUE #743: Double prefix with 'cap-cap'
## The issue:

Authors with slug `cap-cap-` don't appear in the Ajax search `ajax_suggest` and their post counts are zero.

### Method 1

Create new Author from WP Dashboard with Display Name set to `Cap America`. 

Their unique slug is `cap-america`, but their `wp_terms` slug is `cap-cap-america`.

### Method 2

Import users with `create-guest-authors-from-csv` or `create-guest-authors-from-wxr` and include `cap-` in the `user_login`. Once again their their `wp_terms` slug starts with `cap-cap-`.

## Suggested/Possible solution

Remove `cap-` before creating/importing Authors. 

Change https://github.com/Automattic/Co-Authors-Plus/blob/master/co-authors-plus.php#L1524 

`$coauthor_slug = 'cap-' . $coauthor->user_nicename;` to `$coauthor_slug = 'cap-' . str_replace( 'cap-', '', $coauthor->user_nicename );`

Change https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-wp-cli.php#L863

`'user_login' => $author['user_login'],` to `'user_login' => str_replace( 'cap-', '', $author['user_login'] ),`

If this suggested solution is approved, happy to create PR.

rebeccahum:
Fixes #743.

Due to guest author terms being created with a prefix of `cap-`, we should search by the term name which is akin to the guest author "unique slug".  If we create a guest author called "Cap Underpants", the slug will be `cap-cap-underpants` and therefore, will not return any results from the query when `get_guest_author_by()` is called by `get_coauthor_by()`.

Testing Steps:
1) Create guest author with Display Name of "Cap Underpants".
2) Try to assign that guest author to a post by typing "cap" in the ajax box.
3) You will not see that the expected guest author pop up.
4) Apply patch.
5) Repeat step 2.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1258,7 +1258,7 @@ class CoAuthors_Plus {
 		// Get the guest author objects
 		$found_users = array();
 		foreach ( $found_terms as $found_term ) {
-			$found_user = $this->get_coauthor_by( 'user_nicename', $found_term->slug );
+			$found_user = $this->get_coauthor_by( 'user_nicename', $found_term->name );
 			if ( ! empty( $found_user ) ) {
 				$found_users[ $found_user->user_login ] = $found_user;
 			}


geoffguillain (approved):
Tested and works as expected

PR #751, opened 2021-02-11
Update and improve Travis build for default Xenial dist

rebeccahum:
[Prior to switching to Xenial](https://docs.travis-ci.com/user/trusty-to-xenial-migration-guide) by default, TravisCI was using Trusty builds, which were working as expected. This fixes the Travis build file by:
- enabling mysql as a service
- bumping WP version to recent
- installing VIPCS sniffs and using `WordPress-VIP-Go` ruleset
- requiring PHPUnit of 5.7 since there are incompatibilities with PHP and the default PHPUnit versions
- removing inaccurate PHP syntax error check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 language: php
+services: mysql
 
 matrix:
     include:
         # aliased to a recent 5.6.x version
         - php: "5.6"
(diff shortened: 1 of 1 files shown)


rogertheriault (approved):
Approved

Issue #761, opened 2021-03-27
Some Contributors and Guest Author names are not loading


iasgharalikhan:
Hi, 
We have some guest authors who are not reflecting in the plugin block.

Can you please look into this.

lukaszange:
I also have this error on my productive version, but not yet on the local one to reproduce... The error has been occurring since the update to version #760 (Version 3.4.4).

**More detailed description**
1. Install CAP Version 3.4.4
2. Edit a existing post or create a new one
3. Try to add a co author to this specific post
4. The AJAX suggest box returns the error "Sorry, no matching authors found."

**Circumstances already tested**
The error seems only to be happening on "real" users - guest authors can still be selected.

rebeccahum:
@lukaszange Does rolling back to 3.4.2 fix the issue?

iasgharalikhan:
> @lukaszange Does rolling back to 3.4.2 fix the issue?

Roll back fixed it for me

rebeccahum:
I'm not able to reproduce on a fresh install, whether the real user has been created before CAP was installed or after. Any details on reproducing locally would be great.

iasgharalikhan:
> I'm not able to reproduce on a fresh install, whether the real user has been created before CAP was installed or after. Any details on reproducing locally would be great.

So, we have been using CAP for almost 6 months and all the versions in that time period.

The problem is not specific to older or newer profiles, cause some of the newer users are displaying in the updated version.

But only a few of users are left out.

I can share screenshots if you need some details

rebeccahum:
@iasgharalikhan If it's a few of them being left out, are you able to provide the usernames as examples?

iasgharalikhan:
M A Siraj
PTI
IANS
ANI

Are some user names we had which were not displaying, since we rolled back to the previous version. I don't think I have any more details on this.

lethalmiko:
I installed plugin months after users were created and already existing. Plugin has worked fine for over 4 years until last week Friday 26th March 2021 when it stopped. I searched for 4 different existing users but no results. Rolling back to 3.4.3 solved the problem. One user on the thread I started on wordpress.org wrote this:
-------
giuliovale (@giuliovale)
20 hours, 57 minutes ago
Problem came from change of:

$found_user = $this->get_coauthor_by( ‘user_nicename’, $found_term->slug );

to

$found_user = $this->get_coauthor_by( ‘user_nicename’, $found_term->name );

[...]

PR #764, opened 2021-03-28
Account for guest author terms that begin with "cap-"

### LINKED ISSUE #743: Double prefix with 'cap-cap'

rebeccahum:
#744 introduced the bug #761.

This should fix #743, which was what #744 was intended to do.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1259,6 +1259,12 @@ class CoAuthors_Plus {
 		$found_users = array();
 		foreach ( $found_terms as $found_term ) {
 			$found_user = $this->get_coauthor_by( 'user_nicename', $found_term->slug );
+			if ( ! $found_user && 0 === strpos( $found_term->slug, 'cap-cap-' ) ) {
+				// Account for guest author terms that start with 'cap-'.
+				// e.g. "Cap Ri" -> "cap-cap-ri".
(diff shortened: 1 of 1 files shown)


jeffersonrabb (approved):
🚢

PR #767, opened 2021-03-29
🐛 FIX: Show Gravatar of guest user in editor

### LINKED ISSUE #758: Gravatar of gues author not visible
**Steps to reproduce**

1. Go to `/wp-admin/post-new.php?post_type=guest-author`
2. Add a new guest author
3. Go to `/wp-admin/users.php?page=view-guest-authors`
4. Ensure that Gravatar for guest author is visible
5. Go to `/wp-admin/post-new.php`
6. Add new guest author
7. See that Gravatar is not visible

**Current behaviour**

When adding a guest author, the gravatar is not visible in the editor.

**Expected behaviour**

When adding a guest author, the gravatar is visible in the editor.

**Screenshots**

<table>
<tr>
<td>

users

User gravatar is visible on the user page.
</td>
<td>

guest-authors

Guest author gravatar is visible on the guest author page.
</td>
<td>

editor

Admin gravatar is visible, but guest author gravatar not visible in the editor.
</td>
</tr>
</table>

nielslange:
Fixes #758 

**Change**

Show Gravatars of guest authors on in the editor.

**Steps to test**

1. Check out this PR
2. Go to /wp-admin/post-new.php?post_type=guest-author
3. Add a new guest author
4. Go to /wp-admin/users.php?page=view-guest-authors
5. Ensure that Gravatar for guest author is visible
6. Go to /wp-admin/post-new.php
7. Add new guest author
8. See that Gravatar is visible for both the author and the guest author

**Screenshots**

<table>
<tr>
<td>Before:
<br><br>

Screenshot 2021-03-29 at 10 55 06
</td>
<td>After:
<br><br>

Screenshot 2021-03-29 at 10 54 44
</td>
</tr>
</table>

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -382,7 +382,11 @@ class CoAuthors_Plus {
 				<?php
 				foreach ( $coauthors as $coauthor ) :
 					$count++;
-					$avatar_url = get_avatar_url( $coauthor->ID );
+					if ( isset( $coauthor->type ) && 'guest-author' === $coauthor->type ) {
+					    $avatar_url = get_avatar_url( $coauthor->user_email );
+					} else {
+					    $avatar_url = get_avatar_url( $coauthor->ID );
+					}
 					?>
 					<li>
 						<?php echo get_avatar( $coauthor->ID, $this->gravatar_size ); ?>


htdat:
This works as expected for cases without a custom avatar like the following one: 

Markup on 2021-03-29 at 14:00:18

### However, there are some fewer issues: 

1. With a custom avatar, I think it's expected that the custom avatar will be used but with this patch, Gravar is still being used. 
2. Also, `get_avatar_url` is used in a few other places within the plugin. If you fix them, you will need to fix it too. 

https://github.com/Automattic/Co-Authors-Plus/blob/a5953a245caf2e26f38ac18ce99868f50923eaa3/co-authors-plus.php#L478

https://github.com/Automattic/Co-Authors-Plus/blob/a5953a245caf2e26f38ac18ce99868f50923eaa3/co-authors-plus.php#L1207-L1208

### I notice that there is a filter regarding this here: 

https://github.com/Automattic/Co-Authors-Plus/blob/a5953a245caf2e26f38ac18ce99868f50923eaa3/co-authors-plus.php#L1751-L1765

I have run a quick test and it seems updating L1760 like similar to you fix would help this issue. 
Updating this filter may not require to update all other plugins using `get_avatar_url`.

nielslange:
Thanks for your review, @htdat. Somehow, the filter `filter_pre_get_avatar_data_url` only gets triggered for the regular users, but not for the guest authors. My hunch is because they're stored as CPT and not as regular users. That said, I'll look into this and of course the Gravatar should only be used, if no avatar is given.

htdat:
> Somehow, the filter filter_pre_get_avatar_data_url only gets triggered for the regular users, but not for the guest authors.

In theory, `filter_pre_get_avatar_data_url` should be triggered when `get_avatar_url` is run. Trace: 

- [get_avatar_url()](https://github.com/WordPress/WordPress/blob/67d6f5f6c6a77b16c154a9cfbc9f555a2787b280/wp-includes/link-template.php#L4141-L4141) actually extracts `url` from `get_avatar_data()`.
- [get_avatar_data()](http://github.com/WordPress/WordPress/blob/67d6f5f6c6a77b16c154a9cfbc9f555a2787b280/wp-includes/link-template.php#L4282) will call this filter `pre_get_avatar_data`. 

If it does not work in some specific cases, that may be `filter_pre_get_avatar_data_url` is not hooked with the proper time?

nielslange:
> In theory, filter_pre_get_avatar_data_url should be triggered when get_avatar_url is run.

That's what I thought as well, but I'm seeing a different behaviour.

> If it does not work in some specific cases, that may be filter_pre_get_avatar_data_url is not hooked with the proper time?

That's possible and I'm proceeding with this PR now.

nielslange:
Not sure what happened yesterday, but adjusting the filter worked like charm today, @htdat. Thanks for pointing out the filter. 🙌

co-authors-plus.php
```diff
@@ -1753,9 +1753,11 @@ public function filter_pre_get_avatar_data_url( $args, $id ) {
 			return $args;
 		}
 		$coauthor = $this->get_coauthor_by( 'id', $id );
-		if ( false !== $coauthor && isset( $coauthor->type ) && 'guest-author' === $coauthor->type ) { 
+		if ( false !== $coauthor && isset( $coauthor->type ) && 'guest-author' === $coauthor->type ) {
 			if ( has_post_thumbnail( $id ) ) {
 				$args['url'] = get_the_post_thumbnail_url( $id, $this->gravatar_size );
+			} elseif ( isset( $coauthor->type ) && 'guest-author' === $coauthor->type ) {
```
htdat, line 1759: With L1756, this L1957 will be always true. This line might check `$coauthor->user_email` instead?

htdat (changes requested):

co-authors-plus.php
nielslange: Good catch. What was I thinking here? 🤭 Just fixed it as you're 100% correct. `isset( $coauthor->type ) && 'guest-author' === $coauthor->type` will always be true at this line.

rebeccahum (approved):

PR #768, opened 2021-03-29
🐛 FIX: Fix behaviour of filter button on guest author page

### LINKED ISSUE #715: Nonfunctional "filter" button on Contributor Profiles page
On the `/wp-admin/users.php?page=view-guest-authors&orderby=last_name&order=asc` page, there is a nonfunctional `Filter` button:

image

nielslange:
Fixes #715

**Change**

Fix behaviour of filter button on guest author page

**Steps to test**

1. Check out this PR
2. Head over to `/wp-admin/users.php?page=view-guest-authors`
3. Ensure that the filter button is working as expected and that filter options are visible

**Screenshots**

<table>
<tr>
<td>Before:
<br><br>

#715-before
</td>
<td>After:
<br><br>

#715-after-1
</td>
</tr>
</table>

### DIFF
--- a/php/class-coauthors-wp-list-table.php
+++ b/php/class-coauthors-wp-list-table.php
@@ -276,8 +276,8 @@ if ( 'top' == $which ) {
 			echo '<option value="' . esc_attr( $key ) . '" ' . selected( $this->active_filter, $key, false ) . '>' . esc_attr( $value ) . '</option>';
 		}
 		echo '</select>';
+	    submit_button( __( 'Filter', 'co-authors-plus' ), 'secondary', false, false );
 	}
-	submit_button( __( 'Filter', 'co-authors-plus' ), 'secondary', false, false );
 }
 		?></div><?php
 	}
--- a/php/class-coauthors-wp-list-table.php
+++ b/php/class-coauthors-wp-list-table.php
@@ -100,8 +100,6 @@ class CoAuthors_WP_List_Table extends WP_List_Table {
         switch ( $this->active_filter ) {
             case 'with-linked-account':
                 $args['meta_query'] = array(
-                    'relation' => 'AND',
-                    array( 'key' => $key, 'compare' => 'EXISTS' ),
                     array( 'key' => $key, 'compare' => '!=', 'value' => '' ),
                 );
                 break;
--- a/php/class-coauthors-wp-list-table.php
+++ b/php/class-coauthors-wp-list-table.php
@@ -10,6 +10,8 @@ require_once( ABSPATH . 'wp-admin/includes/class-wp-list-table.php' );
 class CoAuthors_WP_List_Table extends WP_List_Table {
 
 	var $is_search = false;
+    var $filters = array();
+    var $active_filter = '';
 
 	function __construct() {
 		if ( ! empty( $_REQUEST['s'] ) ) {
--- a/php/class-coauthors-wp-list-table.php
+++ b/php/class-coauthors-wp-list-table.php
@@ -96,39 +96,23 @@ class CoAuthors_WP_List_Table extends WP_List_Table {
 			$this->active_filter = 'show-all';
 		}
 
-		switch ( $this->active_filter ) {
-			case 'with-linked-account':
-			case 'without-linked-account':
-				$key = $coauthors_plus->guest_authors->get_post_meta_key( 'linked_account' );
-				if ( 'with-linked-account' == $this->active_filter ) {
-                    $args['meta_query'] = array(
-                        'relation' => 'AND',
-                        array(
-                            'key'     => $key,
-                            'compare' => 'EXISTS',
-                        ),
-                        array(
-                            'key'     => $key,
-                            'value'   => '',
-                            'compare' => '!=',
-                        ),
-                    );
-				} else {
-                    $args['meta_query'] = array(
-                        'relation' => 'OR',
-                        array(
-                            'key'     => $key,
-                            'compare' => 'NOT EXISTS',
-                        ),
-                        array(
-                            'key'     => $key,
-                            'value'   => '',
-                            'compare' => '=',
-                        ),
-                    );
-				}
+        $key = $coauthors_plus->guest_authors->get_post_meta_key( 'linked_account' );
+        switch ( $this->active_filter ) {
+            case 'with-linked-account':
+                $args['meta_query'] = array(
+                    'relation' => 'AND',
+                    array( 'key' => $key, 'compare' => 'EXISTS' ),
+                    array( 'key' => $key, 'compare' => '!=', 'value' => '' ),
+                );
                 break;
-		}
+            case 'without-linked-account':
+                $args['meta_query'] = array(
+                    'relation' => 'OR',
+                    array( 'key' => $key, 'compare' => 'NOT EXISTS' ),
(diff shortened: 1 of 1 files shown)


rebeccahum:
What are the pros and cons of removing it versus making it functional?

nielslange:
> What are the pros and cons of removing it versus making it functional?

Good question, @rebeccahum. I looked up this problem again and saw that the filter was no loner working due to missing class attibutes. I just adjusted this and the filter is now working as expected.

**Screenshots**

<table>
<tr>
<td>Before:
<br><br>

#715-before
</td>
<td>After:
<br><br>

#715-after-1
</td>
</tr>
</table>

rebeccahum:
Thanks for this! On testing, I'm seeing that there is a bug where guest authors with previously linked accounts (now unlinked) are showing up when you filter by linked account:

Screen Shot 2021-03-29 at 10 52 24 AM

Can you please investigate so the button is fully functional?

php/class-coauthors-wp-list-table.php
```diff
@@ -10,6 +10,8 @@
 class CoAuthors_WP_List_Table extends WP_List_Table {
 
 	var $is_search = false;
+    var $filters = array();
+    var $active_filter = '';
```
htdat, line 14: Nitpick: these lines may use tab instead of space for indentation.

php/class-coauthors-wp-list-table.php
```diff
@@ -276,8 +278,8 @@ function extra_tablenav( $which ) {
 			echo '<option value="' . esc_attr( $key ) . '" ' . selected( $this->active_filter, $key, false ) . '>' . esc_attr( $value ) . '</option>';
 		}
 		echo '</select>';
+	    submit_button( __( 'Filter', 'co-authors-plus' ), 'secondary', false, false );
```
htdat, line 281: Similarly, here it looks like 1 tab + 4 spaces. Maybe 2 tabs for it.

php/class-coauthors-wp-list-table.php
nielslange: That's exactly why I created https://github.com/Automattic/Co-Authors-Plus/issues/769. Manually checking the formatting is just a waste of time IMO.

php/class-coauthors-wp-list-table.php
nielslange: See https://github.com/Automattic/Co-Authors-Plus/pull/768#discussion_r603798628

nielslange:
@rebeccahum Thanks for bringing this up. I fixed this problem and updated this PR.

php/class-coauthors-wp-list-table.php
```diff
@@ -97,14 +99,35 @@ function prepare_items() {
 		switch ( $this->active_filter ) {
 			case 'with-linked-account':
 			case 'without-linked-account':
-				$args['meta_key'] = $coauthors_plus->guest_authors->get_post_meta_key( 'linked_account' );
+				$key = $coauthors_plus->guest_authors->get_post_meta_key( 'linked_account' );
 				if ( 'with-linked-account' == $this->active_filter ) {
-					$args['meta_compare'] = '!=';
+                    $args['meta_query'] = array(
```
htdat, line 104: Looks like this does not get advantages of `switch case`. Instead of using this condition `if ( 'with-linked-account' == $this->active_filter )`, what's about moving the value of `$args['meta_query']` to each `case`?

htdat (commented):
WFM but there is an opportunity to improve the code clarity as mentioned above.

php/class-coauthors-wp-list-table.php
nielslange: > Looks like this does not get advantages of switch case. Instead of using this condition if ( 'with-linked-account' == $this->active_filter ), what's about moving the value of $args['meta_query'] to each case?

Absolutely. I was so focused on the query, that I didn't notice the strange switch case statement. I just updated that section and used a compact version to increase the readability. At least, I think the code is more readable with the compact version. Let me know what you think about it.

php/class-coauthors-wp-list-table.php
```diff
@@ -94,18 +96,23 @@ function prepare_items() {
...
-				}
-				$args['meta_value'] = '0';
-				break;
-		}
+        $key = $coauthors_plus->guest_authors->get_post_meta_key( 'linked_account' );
+        switch ( $this->active_filter ) {
+            case 'with-linked-account':
+                $args['meta_query'] = array(
+                    'relation' => 'AND',
+                    array( 'key' => $key, 'compare' => 'EXISTS' ),
```
rebeccahum, line 104: What's the reasoning for this meta query check if we already have the one below where it checks for a key that does not equate to an empty value?

rebeccahum (changes requested):

php/class-coauthors-wp-list-table.php
htdat: LGTM 👍

php/class-coauthors-wp-list-table.php
```diff
@@ -276,8 +283,8 @@ function extra_tablenav( $which ) {
 			echo '<option value="' . esc_attr( $key ) . '" ' . selected( $this->active_filter, $key, false ) . '>' . esc_attr( $value ) . '</option>';
 		}
 		echo '</select>';
+	    submit_button( __( 'Filter', 'co-authors-plus' ), 'secondary', false, false );
```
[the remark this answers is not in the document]
nielslange: I moved the `submit_button()` into the `if ( ! empty( $this->filters ) )` block to ensure that the button is not visible in case there are no filters available, e.g. in case a client tried to overwrite/remove the filters.

php/class-coauthors-wp-list-table.php
nielslange: Good catch, @rebeccahum. There's no reason having `array( 'key' => $key, 'compare' => 'EXISTS' )` in this case. I just updated that statement and removed `'relation' => 'AND'` as well, as that's no longer needed.

htdat (approved):
LGTM. 

Not related to this issue directly but from reading the code, it's pretty hard to trace down what `$this->filters`'s value is. I think we can move up this declaration into the constructor (but it should be another PR). 

https://github.com/Automattic/Co-Authors-Plus/blob/a5953a245caf2e26f38ac18ce99868f50923eaa3/php/class-coauthors-wp-list-table.php#L85-L89

nielslange:
> LGTM.
> 
> Not related to this issue directly but from reading the code, it's pretty hard to trace down what `$this->filters`'s value is. I think we can move up this declaration into the constructor (but it should be another PR).
> 
> https://github.com/Automattic/Co-Authors-Plus/blob/a5953a245caf2e26f38ac18ce99868f50923eaa3/php/class-coauthors-wp-list-table.php#L85-L89

I see what you mean. What about moving the filters into a separate function, e.g. `get_filters()`, so that we can also better unit test it?

rebeccahum (approved):

rebeccahum:
@nielslange A helper function and unit test is not necessary given it is just an assignment to a static variable.

htdat:
> I see what you mean. What about moving the filters into a separate function, e.g. get_filters(), so that we can also better unit test it?

Yeah, agree with @rebeccahum that is not necessary in this specific case.

PR #774, opened 2021-04-01
FIX: Correct unit tests when running WP 5.7

### LINKED ISSUE #458: Tests failing with PHP 7

nielslange:
Fixes #458 

**Changes**

Correct two unit tests that are causing an error when running under WP 5.7.

**Steps to test**

1. Check out this PR
2. Run `phpunit` in the plugin directory

**Notes**

Testing this PR requires a temporary WordPress installation. A temporary WordPress installation can be created with the following steps:

1. Install MariaDB via `brew install mariadb`
2. Start MariaDB as service via `brew services start mariadb`
3. Go to plugin directory
4. Run `./bin/install-wp-tests.sh co_authors_plus root root`

Please note that this PR only resolves the problem that appears when using WordPress 5.7, the latest available version. It does not resolve the theme-related failed test case when using WordPress 5.3.

Please also note that the previous featured image was an artificial file and not a real image file. This caused a problem in the Twenty Twenty-One theme as this artificial file does not have any metadata such as a height or a width. Therefore, I've replaced that artificial file with a real image dummy file.


tests/test-coauthors-guest-authors.php
```diff
@@ -124,22 +124,15 @@ public function test_get_guest_author_thumbnail() {
...
 
-		// Checks when guest author has thumbnail.
-		$filename = rand_str() . '.jpg';
-		$contents = rand_str();
-		$upload   = wp_upload_bits( $filename, null, $contents );
-
-		$this->assertTrue( empty( $upload['error'] ) );
-
-		$attachment_id = $this->_make_attachment( $upload );
+		$attachment_id = $this->factory->attachment->create_upload_object( __DIR__ . '/dummy-attachment.png' );
```
htdat, line 127: 👍 Agreed that this is a better approach to generated an attachment_id.

tests/test-coauthors-guest-authors.php
```diff
@@ -124,22 +124,15 @@ public function test_get_guest_author_thumbnail() {
...
 
 		set_post_thumbnail( $guest_author->ID, $attachment_id );
 
 		$thumbnail = $guest_author_obj->get_guest_author_thumbnail( $guest_author, 0 );
 
 		$this->assertContains( 'avatar-0', $thumbnail );
-		$this->assertContains( $filename, $thumbnail );
-		$this->assertContains( 'src="' . wp_get_attachment_url( $attachment_id ) . '"', $thumbnail );
+		$this->assertContains( 'dummy-attachment', $thumbnail );
```
htdat, line 134: I wondered why this should not be `dummy-attachment.png` but after running a few tests and `var_dump`'ing `$thumnail`, I can understand your intention. If we keep testing, `$thumnail` will be added with a suffix number like `dummy-attachment-7.png`. 

Nitpick: It would be nice to have a comment here.

tests/test-coauthors-guest-authors.php
nielslange: Glad to hear that, @htdat.

htdat (approved):
Great job here! 

If possible, it would be helpful if there was an explanation why this PR changes the way of creating attachments. 

As I understand, the old way of creating attachments without a real file will not create metadata for this attachment. And then the test (with WP 5.7) failed in line 136 of TwentyOne theme https://github.com/WordPress/WordPress/blob/5.7/wp-content/themes/twentytwentyone/inc/template-functions.php#L435-L436 as `$meta` here will be `false`:

```php
		$meta = wp_get_attachment_metadata( $attachment->ID );
		if ( $meta['width'] && $meta['height'] ) {
```

htdat:
Also, it would be nice to create a separate PR to add PHP 7.3 and PHP 7.4 to `.travis.yml` as currently the plugin is not running tests for these PHP versions. 

https://github.com/Automattic/Co-Authors-Plus/blob/371d7570e1585cfd8fe93beee6a3d7ae6c2f6783/.travis.yml#L22-L34

GaryJones:
Could also add TravisCI jobs for PHP 8.0 and nightly (allow_failure) as well

tests/test-coauthors-guest-authors.php
GaryJones, line 127: I would suggest putting the image into `tests/fixtures/` directory, so that any files that are needed for tests, _but are not test files_ are kept separate.

tests/test-coauthors-guest-authors.php
GaryJones: This could also be changed to use `assertStringContainsString()`.

tests/test-coauthors-guest-authors.php
nielslange: Thanks for pointing this out. The dummy image has been moved accordingly.

tests/test-coauthors-guest-authors.php
nielslange: > If possible, it would be helpful if there was an explanation why this PR changes the way of creating attachments.

Thanks for your feedback, @htdat. Regarding adding an explanation about this change, I wonder what the benefit of such an explanation would be. It's my understanding that comments should be added to parts that do not explain themselves. In this case, once the PR got merged, a developer will only see that a dummy image will be imported, but will not see that an artificial file had been created in an earlier version. I might be wrong in this case, but I'd not necessarily add an explanation here as the code looks self explaining to me.

> I wondered why this should not be dummy-attachment.png but after running a few tests and var_dump'ing $thumnail, I can understand your intention. If we keep testing, $thumnail will be added with a suffix number like dummy-attachment-7.png.

> Nitpick: It would be nice to have a comment here.

`dummy-attachment-n.png` is indeed the reason why I'm only checking for the filename in this case.

> This could also be changed to use `assertStringContainsString()`.

After your comment, I found the following info regarding `assertContains()` on https://github.com/sebastianbergmann/phpunit/issues/3422#issuecomment-442585223:

> [Soft Deprecation](https://github.com/sebastianbergmann/phpunit/wiki/Release-Process#stage-1-soft-deprecation) in PHPUnit 7.5
[Hard Deprecation](https://github.com/sebastianbergmann/phpunit/wiki/Release-Process#stage-2-deprecation) in PHPUnit 8
[Removal](https://github.com/sebastianbergmann/phpunit/wiki/Release-Process#stage-3-removal) in PHPUnit 9

Therefore, I went ahead and replaced all occurrences of `assertContains()` with `assertStringContainsString()`.

nielslange:
> Also, it would be nice to create a separate PR to add PHP 7.3 and PHP 7.4 to `.travis.yml` as currently the plugin is not running tests for these PHP versions.

> Could also add TravisCI jobs for PHP 8.0 and nightly (allow_failure) as well

@htdat & @GaryJones I created https://github.com/Automattic/Co-Authors-Plus/issues/777 to switch from Travis CI to Circle CI.

tests/test-coauthors-guest-authors.php
htdat: > Thanks for your feedback, @htdat. Regarding adding an explanation about this change, I wonder what the benefit of such an explanation would be. It's my understanding that comments should be added to parts that do not explain themselves.

@nielslange I mean an explanation in this PR description for changing the way of creating attachments. Agree that you do not need to add code comments for this. 

> dummy-attachment-n.png is indeed the reason why I'm only checking for the filename in this case.

The code comment is what I meant for this part.

nielslange: Ah, that's what you meant. Adding an explanation about the change to the PR makes sense. I'll add one now.

nielslange:
@rebeccahum and @GaryJones Can I have your thoughts regarding the failed Travis CI tests as seen on https://travis-ci.org/github/Automattic/Co-Authors-Plus/builds/766742529? Earlier today, I replaced `assertContains()` with `assertStringContainsString()`, as `assertContains()` is depricated and will be removed with PHPUnit 9. However, this repo was still using PHPUnit 5, which does not support `assertStringContainsString()` yet. I bumped the PHPUnit version from 5 to 7, to notice that PHPUnit 7 requires PHP 7.1, while the [minimum required version to run WordPress is PHP 5.6](https://wordpress.org/about/requirements/). 

Given that PHP 7.4 or greater is the official recommended version based on https://wordpress.org/about/requirements/, I tend run the PHPUnit tests only against PHP 7.1 or greater. I'm keen to hear your thoughts on this matter.

GaryJones:
If the plugin is still meant to run on PHP 5.6, then you can use PHPUnit 7.5, and ignore the platform requirements when installing Composer dependencies in the Travis config.

See https://github.com/Yoast/wordpress-seo/blob/ba38278c1d94c09d11f92ed6646930c72e3e4810/.travis.yml#L136-L154 as an example.

nielslange:
Thanks for your thoughts and pointing out Yoast's example, @GaryJones.

nielslange:
@rebeccahum Could you look up this change, to check if it's in a shape to get merged, to reduce the noise in the unit tests?

rebeccahum (approved):

PR #775, opened 2021-04-02
FIX: Handle display name with special chars

### LINKED ISSUE #370: Pipe character in user meta prevents adding as coauthor
If I, for example, have a guest author with `Adam Heckler | One Cool Dude` as the display name, I cannot add that user as a coauthor from within the post editing screen.

I think the bug is here:

https://github.com/Automattic/Co-Authors-Plus/blob/3.2/js/co-authors-plus.js#L195

... where we get the author info using `.split( '|' )`.

We should probably sanitize it or reject pipes in user meta altogether.

From VIP ticket 55563. cc @ryanmarkel

nielslange:
Fixes #370 

**Changes**

Allow users and guest users to use special characters in the display name.

**Steps to test**

1. Create a guest author with a special character in the display name, e.g. `Jane | Doe`
2. Open an existing post and add that guest author to it
3. Verify that the avatar/Gravatar is broken
4. Check out this PR
5. Refresh the post page
6. Add the guest author again
7. Verify that the avatar/Gravatar is visible

**Screenshots**

<table>
<tr>
<td>Before:
<br><br>

Screenshot 2021-04-02 at 17 30 01
</td>
<td>After:
<br><br>

Screenshot 2021-04-02 at 17 27 45
</td>
</tr>
</table>

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1209,6 +1209,8 @@ class CoAuthors_Plus {
 				"%s ∣ %s ∣ %s ∣ %s ∣ %s ∣ %s \n",
 				esc_html( $author->ID ),
 				esc_html( $author->user_login ),
+				// Ensure that author names can contain a pipe character by replacing the pipe character with the
+				// divides character, which will now serve as a delimiter of the author parameters. (#370)
 				esc_html( str_replace( '∣', '|', $author->display_name ) ),
 				esc_html( $author->user_email ),
 				esc_html( rawurldecode( $author->user_nicename ) ),
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1205,8 +1205,15 @@ class CoAuthors_Plus {
 		}
 
 		foreach ( $authors as $author ) {
-			$avatar_url = get_avatar_url( $author->ID );
-			echo esc_html( $author->ID . ' | ' . $author->user_login . ' | ' . $author->display_name . ' | ' . $author->user_email . ' | ' . rawurldecode( $author->user_nicename ) ) . ' | ' . esc_url( $avatar_url ) . "\n";
+			printf(
+				"%s ∣ %s ∣ %s ∣ %s ∣ %s ∣ %s \n",
+				esc_html( $author->ID ),
+				esc_html( $author->user_login ),
+				esc_html( str_replace( '∣', '|', $author->display_name ) ),
+				esc_html( $author->user_email ),
+				esc_html( rawurldecode( $author->user_nicename ) ),
+				esc_url( get_avatar_url( $author->ID ) )
+			);
 		}
 
 		die();
@@ -1376,7 +1383,7 @@ class CoAuthors_Plus {
 		?>
 			<script type="text/javascript">
 				// AJAX link used for the autosuggest
-				var coAuthorsPlus_ajax_suggest_link = 
+				var coAuthorsPlus_ajax_suggest_link =
 				<?php
 				echo wp_json_encode(
 					add_query_arg(
--- a/js/co-authors-plus.js
+++ b/js/co-authors-plus.js
@@ -193,7 +193,7 @@ jQuery( document ).ready(function () {
 	// Callback for when a user selects a co-author
 	function coauthors_autosuggest_select() {
 		$this = jQuery( this );
-		var vals = this.value.split( '|' );
+		var vals = this.value.split( '∣' );
 
 		var author = {}
 		author.id = jQuery.trim( vals[0] );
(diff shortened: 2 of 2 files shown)


co-authors-plus.php
```diff
@@ -1376,7 +1383,7 @@ public function js_vars() {
 		?>
 			<script type="text/javascript">
 				// AJAX link used for the autosuggest
-				var coAuthorsPlus_ajax_suggest_link = 
+				var coAuthorsPlus_ajax_suggest_link =
```
nielslange, line 1386: I noticed, that this line was somehow skipped with https://github.com/Automattic/Co-Authors-Plus/pull/771. To avoid that I need to create a separate PR, I'll keep it in this PR, even though it's not related to the problem this PR fixes.

co-authors-plus.php
```diff
@@ -1205,8 +1205,15 @@ public function ajax_suggest() {
 		}
 
 		foreach ( $authors as $author ) {
-			$avatar_url = get_avatar_url( $author->ID );
-			echo esc_html( $author->ID . ' | ' . $author->user_login . ' | ' . $author->display_name . ' | ' . $author->user_email . ' | ' . rawurldecode( $author->user_nicename ) ) . ' | ' . esc_url( $avatar_url ) . "\n";
+			printf(
+				"%s ∣ %s ∣ %s ∣ %s ∣ %s ∣ %s \n",
+				esc_html( $author->ID ),
+				esc_html( $author->user_login ),
+				esc_html( str_replace( '∣', '|', $author->display_name ) ),
```
htdat, line 1212: Hmm... I would explain the differences between the 2 characters in the first and second parameters here. 

It seems tricky and it does not work if the display name is using both characters like this`Jane | Doe ∣ do`. 

A workaround here is that using HTML entity for the 2nd parameter like this `&#x2223;`.

htdat (changes requested):
I commented a trick to workaround below. However, I think if possible, we may consider switching to JSON like what's recommended in #370

co-authors-plus.php
nielslange: I tried both using a JSON file and converting the pipe to its HTML entity. Both approaches were failing in this case, as the HTML entity will be seen as pipe and therefore also be handled as a delimiter and the JSON file will be visible in plain text to the user. Given that this is an edge case, I solved it by replacing the pipe sign with the divisor sign, which is not accessible on the keyboard by default.

That said, feel free to tinker with this PR. Maybe you'll find a way to improve the current PR. I simply don't want to spend too much time on this edge case.

htdat: > Both approaches were failing in this case, as the HTML entity will be seen as pipe and therefore also be handled as a delimiter and the JSON file will be visible in plain text to the user

Yeah, that makes sense now. 
If using JSON, actually, the current code will need to be changed and handled with the JSON object rather than our own format. However, I've looked at the code and this may include so many changes, which is not necessary for this edge case.

htdat (approved):
LGTM given @nielslange's explanation here https://github.com/Automattic/Co-Authors-Plus/pull/775#discussion_r606754872

rebeccahum:
During testing, it seems to break for me:

Screen Shot 2021-04-06 at 10 33 57 AM

My testing steps:
1) Create guest author with display name "Test | Pipe"
2) Go to a post
3) Assign "Test | Pipe" author.

nielslange:
@rebeccahum I'm afraid, I'm unable to reproduce what you;'re seeing on your end. On my end, when using a pipe, the guest author is visible as expected. I tested that both with my existing testing site and with a brand-new testing site.

Screenshot 2021-04-07 at 08 07 30

rebeccahum:
On a clean WP install, I'm still seeing the same issue:

Screen Shot 2021-04-06 at 7 27 36 PM

WP 5.7
PHP 7.4
Chrome

nielslange:
@rebeccahum Do you see any errors in the console? Strange enough, I'm using the exact environment and on my end, the fix works as expected. Without seeing what you are seeing, it's a bit tricky to apply changes.

rebeccahum:
No console errors that stand out for me. cc @htdat do you run into the same issue as me in testing?

htdat:
> @htdat do you run into the same issue as me in testing?
> WP 5.7
> PHP 7.4
> Chrome

Works well for me. I am using the same env. 
Maybe providing the content of the relevant AJAX response is helpful? Here is mine: 

```
115 ∣ test-pipe ∣ Test | Pipe ∣  ∣ test-pipe ∣ http://0.gravatar.com/avatar/?s=96&#038;d=mm&#038;r=g 
```

rebeccahum:
Ah, looks like the issue was browser cache enabled — it goes away when I disable it.

rebeccahum (approved):
LGTM 👍

nielslange:
> Ah, looks like the issue was browser cache enabled — it goes away when I disable it.

I'm glad to hear that it;s working on your end now.

PR #776, opened 2021-04-09
Fix incorrect post count for linked accounts

### LINKED ISSUE #460: Inaccurate Linked Guest Author post count if Guest Author made post prior to Linking

nielslange:
Fixes #460 

**Changes**

Fix incorrect post count for linked accounts

**Steps to test**

1. Check out this PR
2. Create the following three test cases.

---

**Test case**

1. Create an editor account (editor-460)
2. Create two posts (#460-1 & #460-2) as editor
3. Create a guest account (guest-460)
4. Create one post (#460-3) as guest
5. Check post count for editor and guest
6. Link guest account (guest-460) with editor account (editor-460)
7. Check post count for editor and guest

Expected count before linking:

- ✅ Editor: 2 posts
- ✅ Guest: 1 post 

Expected count after linking:

- ✅ Editor: 3 posts
- ✅ Guest: 3 posts

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1048,25 +1048,23 @@ class CoAuthors_Plus {
 	 * @return int Post count
 	 */
 	function filter_count_user_posts( $count, $user_id ) {
-		$user = get_userdata( $user_id );
+		$user     = get_userdata( $user_id );
+		$coauthor = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
+
+		// Return $count, if no coauthor exists.
+		if ( ! is_object( $coauthor ) ) {
+			return $count;
+		}
 
 		// Return combined post count, if account is linked.
-		$coauthor = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
-		if ( is_object( $coauthor ) && strlen( $coauthor->linked_account ) > 2 ) {
+		if ( strlen( $coauthor->linked_account ) > 2 ) {
 			return $this->get_combined_post_count( $coauthor );
 		}
 
-		// Return term count, if account is not linked and term count is available.
-		if ( $this->has_author_terms( $user ) ) {
-			$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
-
-			if ( ! is_object( $user ) ) {
-				return $count;
-			}
-
-			if ( $this->get_author_term( $user ) ) {
-				return $term->count;
-			}
+		// Return term count, if term count is available.
+		$term = $this->get_author_term( $coauthor );
+		if ( is_int( $term->count ) ) {
+			return $term->count;
 		}
 
 		// Return $count as fallback.
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1054,6 +1054,10 @@ class CoAuthors_Plus {
 			$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
 			$term = $this->get_author_term( $user );
 
+			if ( false === $term ) {
+				return $count;
+			}
+
 			if ( ! $user->linked_account ) {
 				$count = $term->count;
 			} else {
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1124,7 +1124,7 @@ class CoAuthors_Plus {
 				{$wpdb->prefix}posts.ID
 			",
 			$user->linked_account,
-			$user->user_nicename,
+			$user->user_nicename
 		);
 
 		return count( $wpdb->get_results( $query ) );
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1058,13 +1058,13 @@ class CoAuthors_Plus {
 
 		$term = $this->get_author_term( $coauthor );
 
-		// Return combined post count, if account is linked.
-		if ( is_object( $term ) && strlen( $coauthor->linked_account ) > 2 ) {
-			return $count + $term->count;
-		}
-
-		// Return term count, if term count is available.
 		if ( is_object( $term ) ) {
+			// Return combined post count, if account is linked.
+			if ( strlen( $coauthor->linked_account ) > 2 ) {
+				return $count + $term->count;
+			}
+
+			// Otherwise, return the term count.
 			return $term->count;
 		}
 
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1051,8 +1051,9 @@ class CoAuthors_Plus {
 		$user = get_userdata( $user_id );
 
 		// Return combined post count, if account is linked.
-		if ( $this->has_linked_account( $user ) ) {
-			return $this->get_combined_post_count( $this->get_coauthor_by( 'user_nicename', $user->user_nicename ) );
+		$coauthor = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
+		if ( is_object( $coauthor ) && strlen( $coauthor->linked_account ) > 2 ) {
+			return $this->get_combined_post_count( $coauthor );
 		}
 
 		// Return term count, if account is not linked and term count is available.
@@ -1070,18 +1071,6 @@ class CoAuthors_Plus {
 
 		// Return $count as fallback.
 		return $count;
-	}
-
-	/**
-	 * Check if user is linked to another account.
-	 *
-	 * @param WP_User $user The WP_User object.
-	 *
-	 * @return bool True on success, false on failure.
-	 */
-	function has_linked_account( $user ) {
-		$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
-		return (bool) $user->linked_account;
 	}
 
 	/**
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1059,12 +1059,12 @@ class CoAuthors_Plus {
 		$term = $this->get_author_term( $coauthor );
 
 		// Return combined post count, if account is linked.
-		if ( is_int( $term->count ) && strlen( $coauthor->linked_account ) > 2 ) {
+		if ( is_object( $term->count ) && strlen( $coauthor->linked_account ) > 2 ) {
 			return $count + $term->count;
 		}
 
 		// Return term count, if term count is available.
-		if ( is_int( $term->count ) ) {
+		if ( is_object( $term->count ) ) {
 			return $term->count;
 		}
 
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1058,7 +1058,7 @@ class CoAuthors_Plus {
 				return $this->get_linked_post_count( $user->user_nicename, $user->linked_account );
 			}
 
-			if ( null !== $term->count ) {
+			if ( isset( $term->count ) && null !== $term->count ) {
 				return $term->count;
 			}
 		}
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1131,32 +1131,6 @@ class CoAuthors_Plus {
 	}
 
 	/**
-	 * Get the post IDs by author name.
-	 *
-	 * @param string $author_name The name of the corresponding user.
-	 *
-	 * @return array The IDs of all matching posts.
-	 */
-	function get_post_ids_by_author_name( $author_name ) {
-		$results = array();
-		$args    = array(
-			'author_name' => $author_name,
-			'post_type'   => 'post',
-		);
-		$query   = new WP_Query( $args );
-
-		if ( ! $query->post_count ) {
-			return $results;
-		}
-
-		foreach ( $query->posts as $post ) {
-			$results[] = $post->ID;
-		}
-
-		return $results;
-	}
-
-	/**
 	 * Checks to see if the current user can set co-authors or not
 	 */
 	function current_user_can_set_authors( $post = null ) {
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1094,7 +1094,8 @@ class CoAuthors_Plus {
 	function get_combined_post_count( $user ) {
 		global $wpdb;
 
-		$query = $wpdb->prepare( "
+		$query = $wpdb->prepare(
+			"
 			SELECT
 				COUNT( {$wpdb->prefix}posts.ID ) as post_count
 			FROM
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1052,20 +1052,57 @@ class CoAuthors_Plus {
 
 		if ( is_object( $user ) ) {
 			$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
-
 			$term = $this->get_author_term( $user );
 
-			if ( $term && ! is_wp_error( $term ) ) {
-				if ( 'guest-author' === $user->type ) {
-					// If using guest author term count, add on linked user count.
-					$count = (int) $count + $term->count;
-				} else {
-					$count = $term->count;
-				}
+			if ( ! $user->linked_account ) {
+				$count = $term->count;
+			} else {
+				$count = $this->get_linked_post_count( $user->user_nicename, $user->linked_account );
 			}
 		}
 
 		return $count;
+	}
+
+	/**
+	 * Get the post count of linked accounts.
+	 *
+	 * @param string $user_account The name of the user account.
+	 * @param string $linked_account The name of the linked account.
+	 *
+	 * @return int The post count of linked accounts.
+	 */
+	function get_linked_post_count( $user_account, $linked_account ) {
+		$user_account_ids   = $this->get_post_ids_by_author_name( $user_account );
+		$linked_account_ids = $this->get_post_ids_by_author_name( $linked_account );
+
+		return count( array_unique( array_merge( $user_account_ids, $linked_account_ids ) ) );
+	}
+
+	/**
+	 * Get the post IDs by author name.
+	 *
+	 * @param string $author_name The name of the corresponding user.
+	 *
+	 * @return array The IDs of all matching posts.
+	 */
+	function get_post_ids_by_author_name( $author_name ) {
+		$results = array();
+		$args    = array(
+			'author_name' => $author_name,
+			'post_type'   => 'post',
+		);
+		$query   = new WP_Query( $args );
+
+		if ( ! $query->post_count ) {
+			return $results;
+		}
+
+		foreach ( $query->posts as $post ) {
+			$results[] = $post->ID;
+		}
+
+		return $results;
 	}
 
 	/**
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1054,14 +1054,12 @@ class CoAuthors_Plus {
 			$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
 			$term = $this->get_author_term( $user );
 
-			if ( false === $term ) {
-				return $count;
+			if ( strlen( $user->linked_account ) > 1 ) {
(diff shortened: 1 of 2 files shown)


htdat (changes requested):
I tested 3 cases and all is good 👍 

However, there are two things: 
 
- I think it's necessary to have unit tests for the 3 mentioned cases here. 
- In the inline comment I have a concern regarding the performance issue for so many posts and accuracy for VIP Go.

co-authors-plus.php
```diff
@@ -1052,22 +1052,59 @@ function filter_count_user_posts( $count, $user_id ) {
...
+	}
+
+	/**
+	 * Get the post IDs by author name.
+	 *
+	 * @param string $author_name The name of the corresponding user.
+	 *
+	 * @return array The IDs of all matching posts.
+	 */
+	function get_post_ids_by_author_name( $author_name ) {
```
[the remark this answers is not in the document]
mikeyarce: Good catch, @htdat !  

From my looking at this issue - is it a problem we can solve with the current code without needing to add more queries or functions?

rebeccahum:
I was testing Case 1 with the PR and I ran into these PHP errors:

```
Notice: Trying to get property 'linked_account' of non-object in /srv/www/example/public_html/wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 1057

Notice: Trying to get property 'count' of non-object in /srv/www/example/public_html/wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 1058

Notice: Trying to get property 'linked_account' of non-object in /srv/www/example/public_html/wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 1057

Notice: Trying to get property 'count' of non-object in /srv/www/example/public_html/wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 1058
```

co-authors-plus.php
```diff
@@ -1052,22 +1052,59 @@ function filter_count_user_posts( $count, $user_id ) {
...
 			$term = $this->get_author_term( $user );
 
-			if ( $term && ! is_wp_error( $term ) ) {
-				if ( 'guest-author' === $user->type ) {
-					// If using guest author term count, add on linked user count.
-					$count = (int) $count + $term->count;
-				} else {
-					$count = $term->count;
-				}
+			if ( ! $user->linked_account ) {
```
rebeccahum, line 1057: Sometimes users created before CAP is installed will not have a proper term assigned and thus, may return `false`. We need to practice defensive coding for these situations. See https://github.com/Automattic/Co-Authors-Plus/pull/776#issuecomment-816812351 for further details.

rebeccahum (changes requested):

co-authors-plus.php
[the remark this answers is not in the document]
nielslange: > I am concerned that this may be a performance and accuracy issue if either accounts or both have so many posts. Let's think of a few thousand posts at least.

> There is a limit of how many rows can be returned for each SQL command on VIP Go. If I remember correctly, it's 1000.

You're right, @htdat. I haven't thought about that while developing this fix. I think it's best to process the query in batches.

> From my looking at this issue - is it a problem we can solve with the current code without needing to add more queries or functions?

I'm afraid that's not possible, @mikeyarce. The default WordPress functionality does not consider having multiple authors for one post. Let's say you and I are working on the same post and our accounts get linked. Running `count_user_posts()` would result in 2 instead of in 1, as both of us are listed as author in the corresponding post.

That said, if you have an idea on how to solve this problem without adding extra code, I'm all ears.

co-authors-plus.php
nielslange: Thanks for pointing this out, @rebeccahum. I just addressed this problem by checking if `$term` exists for the current user.

rebeccahum: @nielslange Is there anywhere else you could see that could use some defensive coding as well?

co-authors-plus.php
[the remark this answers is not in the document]
mikeyarce: > The default WordPress functionality does not consider having multiple authors for one post

Ahhh interesting, is this another bug?  or is this something caught by the current unit tests?

One thought I had was - in this case, is the bug the counting or the fact that you have 1 author added to a post twice?   Maybe the count is right but the plugins shouldn't keep both of those authors on the post if they are linked?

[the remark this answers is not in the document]
nielslange: > One thought I had was - in this case, is the bug the counting or the fact that you have 1 author added to a post twice?

I wouldn't say that this is a bug as by default a page or post can only have one authors. The `count_user_posts()` function does simply not consider that a page of post can have multiple authors. 

> Maybe the count is right but the plugins shouldn't keep both of those authors on the post if they are linked?

Imagine you create the post `Performant WordPress Queries` with the guest account `VIP Developer` and your regular account `Mikey`. Running `count_user_posts()` shows a post count of 1 for both accounts. You're now linking the guest account with your regular account. `count_user_posts()` still returns 1 for both accounts. For linked accounts, the individual number of all pages and posts will be cumulated. This results in a 2, which is incorrect as both accounts were involved in the same post and not in two different posts.

I'd say, this is a very specific problem that is created by the CAP plugin and therefore needs custom code to calculate the proper post count.

co-authors-plus.php
nielslange: I've updated the logic in the following way:

- If a `$user->linked_account` exists, return the count of the linked account using `get_linked_post_count()`.
- If the `$term->count` is not null, return the `$term->count`.
- If none of the two criteria above are true, return the unfiltered `$count`.

co-authors-plus.php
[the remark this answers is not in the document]
nielslange: > I think it's necessary to have unit tests for the 3 mentioned cases here.

@htdat I just noticed that I haven't replied to that suggestion yet. I'll definitely create unit tests for this issue. Especially given the fact that there are several attempts to resolve this particular problem, unit tests should be added to ensure that future developments do not break this fix.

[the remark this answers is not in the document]
htdat: > That said, if you have an idea on how to solve this problem without adding extra code, I'm all ears.

@nielslange  - What's about if approaching this issue a bit differently? Instead of getting all post IDs from the WP user and the guest author and comparing them, we do this math: 

- (1) Total posts of the WP User
- (2) Total posts of the guest author.
- (3) Total shared posts between this WP User and this guest author. 

The final result is: (1) + (2) - (3). 

To calculate (3), I can think of a count query to get all posts from this specific real user ID having the guest account term. The only concern is that if there is any performant issue with this SQL query.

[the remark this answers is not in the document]
nielslange: That's a neat idea, @htdat. I'll try that right now.

co-authors-plus.php
```diff
@@ -1052,22 +1052,61 @@ function filter_count_user_posts( $count, $user_id ) {
...
 			$term = $this->get_author_term( $user );
 
-			if ( $term && ! is_wp_error( $term ) ) {
-				if ( 'guest-author' === $user->type ) {
-					// If using guest author term count, add on linked user count.
-					$count = (int) $count + $term->count;
-				} else {
-					$count = $term->count;
-				}
+			if ( strlen( $user->linked_account ) > 1 ) {
```
rebeccahum, line 1057: What if `$user` does not return an object from line 1054?

co-authors-plus.php
```diff
@@ -1052,22 +1052,61 @@ function filter_count_user_posts( $count, $user_id ) {
...
+				return $this->get_linked_post_count( $user->user_nicename, $user->linked_account );
+			}
+
+			if ( isset( $term->count ) && null !== $term->count ) {
```
rebeccahum, line 1061: Are you seeing any cases where `$term->count` is a null value?

rebeccahum (changes requested):

co-authors-plus.php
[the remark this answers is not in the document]
rebeccahum: If we can use counts over introducing additional queries, that would be ideal.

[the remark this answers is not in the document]
rebeccahum: If we can use counts over introducing additional queries, that would be ideal. The WP_Term class has the `count` property, have you had a chance to look into utilizing that? https://developer.wordpress.org/reference/classes/wp_term/.

co-authors-plus.php
nielslange: > What if $user does not return an object from line 1054?
In this case, my fix would probably trigger an error. That said, I have rewritten the that function from scratch and added inline comments to it, so that others understand how the posts are counted.

co-authors-plus.php
nielslange: See https://github.com/Automattic/Co-Authors-Plus/pull/776#discussion_r613196898.

co-authors-plus.php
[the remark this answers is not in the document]
nielslange: @rebeccahum After looking into @htdat's suggestion, I rewrote the function and I'm using `COUNT` now.

@htdat Based on your suggestion, I created a query that returns the combined post count directly. Thus, I'm doing (1), (2) and (3) directly in the query.

co-authors-plus.php
```diff
@@ -1050,24 +1050,112 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
 	function filter_count_user_posts( $count, $user_id ) {
 		$user = get_userdata( $user_id );
 
-		if ( is_object( $user ) ) {
+		// Return combined post count, if account is linked.
+		if ( $this->has_linked_account( $user ) ) {
```
rebeccahum, line 1054: Introducing a new helper function won't be necessary for this, as I'm seeing that all it does is return the boolean of `get_coauthor_by()` and on line 1060, `get_coauthor_by()` is also being called. For better readability, is there a better way we could just call the function once? See DRY principle: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

co-authors-plus.php
```diff
@@ -1050,24 +1050,112 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
+					{$wpdb->prefix}terms.name = %s
+				)
+			GROUP BY
+				{$wpdb->prefix}posts.ID
+			",
+			$user->linked_account,
+			$user->user_nicename
+		);
+
+		return count( $wpdb->get_results( $query ) );
```
rebeccahum, line 1130: Is there a way we could avoid direct DB queries? Something like this would negatively affect performance.

co-authors-plus.php
```diff
@@ -1050,24 +1050,112 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
 		$user = get_userdata( $user_id );
 
-		if ( is_object( $user ) ) {
+		// Return combined post count, if account is linked.
+		if ( $this->has_linked_account( $user ) ) {
+			return $this->get_combined_post_count( $this->get_coauthor_by( 'user_nicename', $user->user_nicename ) );
+		}
+
+		// Return term count, if account is not linked and term count is available.
+		if ( $this->has_author_terms( $user ) ) {
```
rebeccahum, line 1059: Instead of introducing an additional function call for a check to see if there is a term associated with an author, is there another way we can do so with the pre-existing code?

co-authors-plus.php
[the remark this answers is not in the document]
rebeccahum: @nielslange Is there a reason why you left this function since it's not being used anymore?

rebeccahum (changes requested):
During testing, I found these errors crop up:

```
Notice: Object of class WP_User could not be converted to int in /srv/www/example/public_html/wp-includes/taxonomy.php on line 2094

Notice: Trying to get property 'linked_account' of non-object in /srv/www/example/public_html/wp-content/plugins/Co-Authors-Plus/co-authors-plus.php on line 1084
```

co-authors-plus.php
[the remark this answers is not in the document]
nielslange: I missed removing that function initially. I just removed it and pushed an update.

co-authors-plus.php
nielslange: I see what you mean. Initially, I introduced `has_linked_account()` as it made that part easier to read for me. That said, I updated that part to avoid DRY.

co-authors-plus.php
nielslange: I removed the additional function call and restructured the code slightly.

co-authors-plus.php
nielslange: I created that additional SQL query for a slight different use case. At the moment, there are three different cases in which the post count is incorrect:

1. A user publishes a post on behalf of a guest author whose account is not linked to a user account. In this case, the post count of the editor will be increased as well, even though the user is not co-author. For this case, I created https://github.com/Automattic/Co-Authors-Plus/issues/780.
2. A user and a guest author, that is not linked to another user account, create a post together. Both see a post count of 1. If the guest account gets linked to the user account, both accounts show a post count of 2 then. For this case, I created https://github.com/Automattic/Co-Authors-Plus/issues/779.
3. A user creates two posts and a guest author creates one post. If the guest account gets linked to the user account, both accounts should show a post count of 3. However, only the user account shows a post count of 3 while the guest account shows a post count of 5. This PR takes care of this third case.

I hope that it's a bit more clear while I went the sidetrack with the additional SQL query. That said, I shouldn't have done that and should have created separate issues for the other two problem cases initially.

co-authors-plus.php
```diff
@@ -1048,23 +1048,27 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
-			if ( $term && ! is_wp_error( $term ) ) {
-				if ( 'guest-author' === $user->type ) {
-					// If using guest author term count, add on linked user count.
-					$count = (int) $count + $term->count;
-				} else {
-					$count = $term->count;
-				}
-			}
+		// Return combined post count, if account is linked.
+		if ( is_int( $term->count ) && strlen( $coauthor->linked_account ) > 2 ) {
```
rebeccahum, line 1062: What cases would `$term->count` not be an integer? In cases where it would not be, do you see a better check available?

co-authors-plus.php
```diff
@@ -1048,23 +1048,27 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
+			return $count + $term->count;
+		}
+
+		// Return term count, if term count is available.
+		if ( is_int( $term->count ) ) {
```
rebeccahum, line 1067: Same as above ^^.

co-authors-plus.php
```diff
@@ -1048,23 +1048,27 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
 
-		if ( is_object( $user ) ) {
-			$user = $this->get_coauthor_by( 'user_nicename', $user->user_nicename );
+		// Return $count, if no coauthor exists.
+		if ( ! is_object( $coauthor ) ) {
+			return $count;
+		}
 
-			$term = $this->get_author_term( $user );
+		$term = $this->get_author_term( $coauthor );
```
rebeccahum, line 1059: Do you see an opportunity from this line to practice defensive coding below?

rebeccahum (changes requested):

co-authors-plus.php
nielslange: I was under the impression that `is_int( $term->count )` in https://github.com/Automattic/Co-Authors-Plus/blob/c9e3f660c2629dfb288f2ae8b3109876c2890564/co-authors-plus.php#L1062 and https://github.com/Automattic/Co-Authors-Plus/blob/c9e3f660c2629dfb288f2ae8b3109876c2890564/co-authors-plus.php#L1067, the sections you mentioned before, are defensive coding for `$term`.

co-authors-plus.php
nielslange: Technically, `$term->count` should always be an integer, if that variable is available. By using `is_int( $term->count )`, I wanted to practice defensive coding, thus, only enter the statement when the variable is an integer.

co-authors-plus.php
nielslange: Same explanation as above.

co-authors-plus.php
nielslange: Ah, I see what you mean. I changed `is_int( $term->count )` to `is_object( $term->count )` to ensure that $term is an object. I kept `strlen( $coauthor->linked_account ) > 2 )` for now, as this check is needed to distinguish if an account is linked.

co-authors-plus.php
```diff
@@ -1048,23 +1048,27 @@ function filter_wp_get_object_terms( $terms, $object_ids, $taxonomies, $args ) {
...
-			if ( $term && ! is_wp_error( $term ) ) {
-				if ( 'guest-author' === $user->type ) {
-					// If using guest author term count, add on linked user count.
-					$count = (int) $count + $term->count;
-				} else {
-					$count = $term->count;
-				}
-			}
+		// Return combined post count, if account is linked.
+		if ( is_object( $term ) && strlen( $coauthor->linked_account ) > 2 ) {
```
rebeccahum, line 1062: Getting closer 🙂. I see that we're doing the `is_object()` check twice - once here and once below. Is there a better way to follow the DRY principle?

nielslange: Sure. I could remove one `is_object()` check by using a nested if-else statement. I decided against it for readability reasons. That said, if you prefer DRY over readability, I'm happy to remove the duplicate.

rebeccahum: Do you see a way to do it without using a nested if-else statement?

nielslange: I could change the order of the checks as follows:

```
if ( ! is_object( $term ) ) {
  return $count;
}

if ( strlen( $coauthor->linked_account ) > 2 ) {
  return $count + $term->count;
}

return $term->count;
```

This way, I would keep the code DRY. The only downside I see, if that `return $term->count;` becomes the fallback value instead of `return $count;`. However, functionality-wise nothing changes.

On the other hand, I could also avoid the else-case by changing the code as follows:

```
if ( is_object( $term ) {
  if ( strlen( $coauthor->linked_account ) > 2 ) ) {
    return $count + $term->count;
  }

  return $term->count;
}
```

While this avoids the else-case, it's still a nested if-statement. While this might be a personal preference, I prefer using [Guard Clauses](https://blog.webdevsimplified.com/2020-01/guard-clauses/) over if-statements.

rebeccahum: Thanks for sharing different approaches! The guard clause may not be ideal in this case, as we are hooking onto a filter and thus, should return back the original value if no conditions are met. 

I find the second example to be more readable while following the DRY principle.

nielslange: I've just updated the corresponding section according to your feedback. 

The reason why I marked the PR as draft is because I want to add a unit test to this PR, so that we can see immediately if a future change is breaking this PR.

rebeccahum:
@nielslange Why don't you work on a new PR for the unit test so we can get this fix merged in?

nielslange:
> @nielslange Why don't you work on a new PR for the unit test so we can get this fix merged in?

Sounds good, @rebeccahum. I'll move the unit test into a separate issue.

rebeccahum (approved):
Approving. Unit test will come later.

PR #778, opened 2021-04-12
get_co_author_by(): Only replace cap- prefix if no results found first

rebeccahum:
Since CAP prefixes terms with `cap-`, we usually replace it to ensure a result is found. However, this breaks for WP_Users that have "Cap" in their name. For testing:

1) Create WP_User with first name "Cap" and last name "America". You should see that the new user has a nicename of `cap-america` now.
2) Using the below snippet to illustrate:
```
global $coauthors_plus;
$coauthor = $coauthors_plus->get_coauthor_by( 'user_nicename', 'cap-america' );
var_dump( $coauthor );
```
One would see that `$coauthor` erroneously returns `false` due to it pre-emptively replacing the `cap-` prefix and therefore, not finding the user. 

We should only do the replacement if no result has been found first and then, try without the `cap-` prefix or else we risk missing nicenames with `cap-`.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -272,11 +272,12 @@ class CoAuthors_Plus {
 				if ( 'user_nicename' == $key ) {
 					$key = 'slug';
 				}
-				// Ensure we aren't doing the lookup by the prefixed value
-				if ( 'login' == $key || 'slug' == $key ) {
-					$value = preg_replace( '#^cap\-#', '', $value );
-				}
 				$user = get_user_by( $key, $value );
+				if ( ! $user && ( 'login' == $key || 'slug' == $key ) ) {
+					// Re-try lookup without prefixed value if no results found.
+					$value = preg_replace( '#^cap\-#', '', $value );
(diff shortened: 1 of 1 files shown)


mikeyarce:
This looks good!  I tested the PR locally and it works as expected.

My only question is what if you have two authors:

`cap-america` and `america`?  

Are there any instances where CAP might return the data from `cap-america` for the `america` user?

rebeccahum:
@mikeyarce Good question! I only kept the stripping of `cap-` for backwards compat purposes but how CAP typically uses the  `get_coauthor_by()` function, one should be passing the non-prefixed CAP term.

E.g. if you want to get the `cap-america` author you _should_ be using:

```
$coauthor = $coauthors_plus->get_coauthor_by( 'user_nicename', 'cap-america' );
```

If you want to get the `america` author you'd use:

```
$coauthor = $coauthors_plus->get_coauthor_by( 'user_nicename', 'america' );
```

mikeyarce (approved):
Awesome, the explanation makes sense!

PR #790, opened 2021-05-14
Sidebar Plugin for Block Editor Integration

laras126:
## Overview

This PR includes the addition of a panel in the Document sidebar in Gutenberg that provides the Co-Authors Plus functionality. There are a few things still to do, but I wanted to open this up now to start getting feedback. 

* A WP API endpoints to retrieve, save, and search coauthors using existing methods
* Use of the `Combobox` block editor component for the UI, including the autocomplete
* A data store to manage the Co-Authors state so we can access the data locally and globally
* A filter to enable the block editor integration, that is set to false by default
* Hide core author meta box by updating the _links entry in the response (To Do)

## Screenshots

Video of adding, removing, and re-ordering Co-Authors using the Sidebar panel.

https://user-images.githubusercontent.com/445861/118292172-59da9480-b4a6-11eb-8d45-c38387efcb34.mp4

## How to Test This PR

1. Pull down this branch for the coauthors plugin in a local WordPress development environment.
2. Create a few Guest Authors.
3. Add a new post. 
4. Observe that the Co-Authors meta box does not display in the editor. 
6. Observe the default Author selection in the "Status & Visibility" sidebar panel does not display. 
5. Observe the "Authors" panel added to the sidebar.
7. Observe the current user is assigned as the default author in the "Authors".
8. Search for the name of another author in the auto-complete section in the Authors panel and add it.
9. Do the same for a few other authors. Using the side buttons, remove and move the authors up and down. 
10. Make sure the Update button is activated when only the authors are changed and no other changes have been made.
7. Save the post and confirm the authors and their order has been saved by reloading the screen.


laras126:
Regarding these to-dos mentioned in the comment:
> There is a post meta value that is updated when the coauthors state changes. This is to trigger the Publish button to become active, see #13774. This approach does not seem to work when the Gutenberg plugin is active, so its needs some additional verification.
> Make sure the Save Draft button is activated when authors are updated on a new post.
Current functionality: With the Gutenberg plugin, there is a disabled Save Draft button on the initial new post editor, and adding a new author does not enable it. Without the Gutenberg plugin, there is no disabled Save Draft button until post content is added. Adding a new Co-Author does not prompt the auto-save.

I feel this is acceptable functionality for an opt-in release - other thoughts welcome!

rebeccahum:
@lschuyler Can you take a look when you get a chance? thx

lschuyler:
Hi @laras126 👋 

Thanks so much for contributing this PR. We appreciate all the work that is represented here.

A few questions/comments I have:

 - The test is failing with an error:
`1) Test_Endpoints::test_authors_get_coauthors`
Undefined index: user_nicename
I see some changes to `user_nicename` / `userNicename` in recent commits. Could that have caused this error?

- The new endpoints restrict access to users with permission to edit posts and authors. This looks good to keep the authors' and coauthors' private details (email, gravatar, etc) unaccessible to unauthorized users.

- In php/class-coauthors-endpoint.php L229 the email is the only element of the returned array that isn't escaped. Is this intentional?

- Rather than make this addition an opt-in feature (by adding the filter as described in the README.md change), could it be enabled automatically if the block editor is in use? This would be preferable.  I'm wondering if [WP_Screen::is_block_editor](https://developer.wordpress.org/reference/classes/wp_screen/is_block_editor/) is an option. What do you think?

lschuyler:
Hi @laras126 - Just checking in to see if you're open to addressing the feedback above, or if you are wanting to hand it off at this point. 
Thanks!

laras126:
Hi @lschuyler! Thank you for the feedback - I should be able to address the feedback in the coming weeks. I will let you know if that's not the case.

laras126:
Hi @lschuyler – I updated the PR with the following changes:

- Made feature the default and removed opt-in logic (much better!)
- Fixed failing tests (though I am only running these tests locally and not seeing the result from when they are run in the CI)
- Added `sanitize_email` to the returned email in the endpoint (lmk if another escaping function is preferred)

Happy to address anything else you see. Thank you!

co-authors-plus.php
```diff
@@ -164,6 +174,59 @@ public function action_init() {
...
+		// }
+
+		// if ( ! $editor_active ) {
+		// 	return false;
+		// }
+
+		// return (bool) apply_filters( 'coauthors_block_editor_integration', self::SIDEBAR_PLUGIN_ENABLED );
+		$screen = get_current_screen();
+
+		return $screen->is_block_editor();
```
lschuyler, line 200: Since `is_block_editor()` was introduced somewhat recently, [in WP 5.0](https://developer.wordpress.org/reference/classes/wp_screen/is_block_editor/), what do you think about checking if it exists `method_exists( $screen, 'is_block_editor' )` and defaulting to false if not?

laras126: Thanks, @lschuyler! Updated (and removed that commented code!). The tests appear to be passing now too. Let me know if there's anything else.

lschuyler: Thanks @laras126! I'll take a look this week.

lschuyler: This looks great @laras126! The tests are all passing for me now too.

lschuyler (dismissed):
This looks great @laras126. The tests are all passing for me now too.
Thanks for all your work here! 
We'll get this merged and released.

co-authors-plus.php
```diff
@@ -66,6 +67,12 @@ class CoAuthors_Plus {
 
 	var $to_be_filtered_caps = array();
 
+	/**
+	 * Block editor functionality in the sidebar is
+	 * disabled by default.
+	 */
+	const SIDEBAR_PLUGIN_ENABLED = false;
```
rebeccahum, line 74: What's the purpose of this constant?

php/class-coauthors-endpoint.php
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	/**
+	 * Validate input arguments.
+	 *
+	 * @param mixed $param Value to validate.
+	 * @return bool
+	 */
+	public function validate_numeric( $param ) {
+		return is_numeric( $param );
```
rebeccahum, line 188: Do we need a wrapper for `is_numeric()` or can we directly just use that as the validation?

rebeccahum:
Thank you so much for your work on this, @laras126! I've added just two minor comments and then we should be good to go.

php/class-coauthors-endpoint.php
ethitter: The wrapper is required because WP passes more arguments to the validation callback than `is_numeric()` accepts, which produces a warning starting in one of the 7.x releases.

co-authors-plus.php
laras126: It is outdated! Good catch – I'll update it momentarily.

php/class-coauthors-endpoint.php
rebeccahum: Ah, okay. Thanks for that, @ethitter.

rebeccahum (approved):

GaryJones:
@lschuyler @rebeccahum @laras126 Was there any consideration of the escaping that seems to be done [here](https://github.com/thedanielcho/Co-Authors-Plus/blob/fd94f7d4cec9346d60b3e7209c286eea4c355113/php/class-coauthors-endpoint.php#L210-L215)? How could the REST Response know what context the escaping would need to be done with? (Possibly causing #937)

PR #792, opened 2021-05-24
Delete user term correctly

wpf500:
The `delete_user_action` doesn't delete the author term as it passes the `user_login` to `wp_delete_term` rather than the term ID.

I found this bug when trying to convert users to guest authors. I wanted the new guest author to have the same slug as the old user, so I used a script to do the following:

1. Create new guest author with user_login "tmpuser"
2. `wp coauthors-plus swap-coauthor --from=user --to=tmpuser`
3. `wp_delete_user($user)`
4. `wp coauthors-plus rename-coauthor --from=tmpuser --to=user`

The final step (rename-coauthor) assigns the old user's term_taxonomy_id to the post rather than the new guest author's one. I'm guessing this is because when it looks up `cap-user` it finds the old term instead of the new one.

This bug has been raised before here:
https://github.com/Automattic/Co-Authors-Plus/issues/268, https://github.com/Automattic/Co-Authors-Plus/pull/313

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -981,7 +981,8 @@ class CoAuthors_Plus {
 		$delete_user = get_user_by( 'id', $delete_id );
 		if ( is_object( $delete_user ) ) {
 			// Delete term
-			wp_delete_term( $delete_user->user_login, $this->coauthor_taxonomy );
+			$term = $this->get_author_term( $delete_user );
+			wp_delete_term( $term->term_id, $this->coauthor_taxonomy );
 		}
 
 		if ( $this->is_guest_authors_enabled() ) {


rebeccahum:
@wpf500 On Step 4, are you getting the error `Error: New user_login value conflicts with existing co-author`?

rebeccahum:
@wpf500 Thanks for your contribution! These are the steps I took to reproduce:

1) Create user 'user1' and guest author 'guestauthor1'
2) Create post and assign 'user1' to it. Post has the current `author` terms to it:
```
  object(WP_Term)#2086 (10) {
    ["term_id"]=>
    int(10)
    ["name"]=>
    string(5) "User1"
    ["slug"]=>
    string(9) "cap-user1"
    ["term_group"]=>
    int(0)
    ["term_taxonomy_id"]=>
    int(10)
    ["taxonomy"]=>
    string(6) "author"
    ["description"]=>
    string(26) "User1   User1 4 user@1.com"
    ["parent"]=>
    int(0)
    ["count"]=>
    int(1)
    ["filter"]=>
    string(3) "raw"
  }
}
```
3) `wp co-authors-plus swap-coauthors --from=user1 --to=guestauthor1`
4) Delete user1
5) `wp co-authors-plus rename-coauthor --from=guestauthor1 --to=user1`
Now the terms attached to the post are:

```
  [0]=>
  object(WP_Term)#2094 (10) {
    ["term_id"]=>
    int(10)
    ["name"]=>
    string(5) "User1"
    ["slug"]=>
    string(9) "cap-user1"
    ["term_group"]=>
    int(0)
    ["term_taxonomy_id"]=>
    int(10)
    ["taxonomy"]=>
    string(6) "author"
    ["description"]=>
    string(24) "guestauthor1   user1 21 "
    ["parent"]=>
    int(0)
    ["count"]=>
    int(2)
    ["filter"]=>
    string(3) "raw"
  }
  [1]=>
  object(WP_Term)#2095 (10) {
    ["term_id"]=>
    int(9)
    ["name"]=>
    string(12) "guestauthor1"
    ["slug"]=>
    string(16) "cap-guestauthor1"
    ["term_group"]=>
    int(0)
    ["term_taxonomy_id"]=>
    int(9)
    ["taxonomy"]=>
    string(6) "author"
    ["description"]=>
    string(31) "guestauthor1   guestauthor1 21 "
    ["parent"]=>
    int(0)
    ["count"]=>
    int(1)
    ["filter"]=>
    string(3) "raw"
  }
```

It just appears that there is an extra term which is the old `cap-guestauthor1` one.

rebeccahum (approved):
Tested and works as expected, thank you!

PR #801, opened 2021-06-07
Bail out early if $wp_query does not exist

### LINKED ISSUE #615: Message "is_author was called <strong>incorrectly</strong>" with plugin restrict-user-access
Message "PHP Notice:  is_author was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run..."

Stack: 
```
#1  is_author() called at [/var/www/html/wordpress/wp-content/plugins/co-authors-plus/co-authors-plus.php:1075]
#2  CoAuthors_Plus-&gt;fix_author_page( AND wp_3_posts.post_type = 'restriction' AND ((wp_3_posts.post_status = 'publish' OR wp_3_posts.post_status = 'future' OR wp_3_posts.post_status = 'draft'))wp_3_posts.post_date DESC) called at [/var/www/html/wordpress/wp-includes/class-wp-hook.php:286]
#3  WP_Hook-&gt;apply_filters(, Array ([0] =&gt;  AND wp_3_posts.post_type = 'restriction' AND ((wp_3_posts.post_status = 'publish' OR wp_3_posts.post_status = 'future' OR wp_3_posts.post_status = 'draft'))wp_3_posts.post_date DESC)) called at [/var/www/html/wordpress/wp-includes/class-wp-hook.php:310]
#4  WP_Hook-&gt;do_action(Array ([0] =&gt;  AND wp_3_posts.post_type = 'restriction' AND ((wp_3_posts.post_status = 'publish' OR wp_3_posts.post_status = 'future' OR wp_3_posts.post_status = 'draft'))wp_3_posts.post_date DESC)) called at [/var/www/html/wordpress/wp-includes/plugin.php:453]
#5  do_action(posts_selection,  AND wp_3_posts.post_type = 'restriction' AND ((wp_3_posts.post_status = 'publish' OR wp_3_posts.post_status = 'future' OR wp_3_posts.post_status = 'draft'))wp_3_posts.post_date DESC) called at [/var/www/html/wordpress/wp-includes/class-wp-query.php:2602]
#6  WP_Query-&gt;get_posts() called at [/var/www/html/wordpress/wp-includes/class-wp-query.php:3230]
#7  WP_Query-&gt;query(Array ([numberposts] =&gt; -1,[category] =&gt; 0,[orderby] =&gt; date,[order] =&gt; DESC,[include] =&gt; Array (),[exclude] =&gt; Array (),[meta_key] =&gt; ,[meta_value] =&gt; ,[post_type] =&gt; restriction,[suppress_filters] =&gt; 1,[post_status] =&gt; Array ([0] =&gt; publish,[1] =&gt; draft,[2] =&gt; future),[posts_per_page] =&gt; -1,[ignore_sticky_posts] =&gt; 1,[no_found_rows] =&gt; 1)) called at [/var/www/html/wordpress/wp-includes/post.php:1763]
#8  get_posts(Array ([numberposts] =&gt; -1,[post_type] =&gt; restriction,[post_status] =&gt; Array ([0] =&gt; publish,[1] =&gt; draft,[2] =&gt; future))) called at [/var/www/html/wordpress/wp-content/plugins/restrict-user-access/app.php:315]...
```

I took the last version of co-authors-plus.php, the issue is the same.

nielslange:
Fixes #615 

**Note**

Add defensive programming to avoid that functions are called to early.

**Steps to test**

1. Install and activate [WooCommerce](https://wordpress.org/plugins/woocommerce/) and [WooCommerce Memberships](https://woocommerce.com/products/woocommerce-memberships/).
2. Create a simple product (a membership is not needed)
3. Look up the product page to see the warning (wp_debug must be active)
4. Check out this PR
5. Look up the product page again and see that warning is gone

**Screenshots**

<table>
<tr>
<td>Before:
<br><br>

#615-before
</td>
<td valign="top">After:
<br><br>

#615-after
</td>
</tr>
</table>

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1135,7 +1135,7 @@ class CoAuthors_Plus {
 
 		global $wp_query, $authordata;
  
-		if ( ! did_action( 'wp' ) ) {
+		if ( ! isset( $wp_query ) ) {
 			return;
 		}
 
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1147,7 +1147,7 @@ class CoAuthors_Plus {
 		}
 
 		$author = $this->get_coauthor_by( 'user_nicename', $author_name );
-		
+
 		global $wp_query, $authordata;
 
 		if ( is_object( $author ) ) {
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1133,6 +1133,12 @@ class CoAuthors_Plus {
 	 */
 	public function fix_author_page( $selection ) {
 
+		global $wp_query, $authordata;
+ 
+		if ( ! isset( $wp_query ) ) {
+			return;
+		}
+
 		if ( ! is_author() ) {
 			return;
 		}
@@ -1143,8 +1149,6 @@ class CoAuthors_Plus {
 		}
 
 		$author = $this->get_coauthor_by( 'user_nicename', $author_name );
-
-		global $wp_query, $authordata;
 
 		if ( is_object( $author ) ) {
 			$authordata = $author; //phpcs:ignore
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1133,6 +1133,8 @@ class CoAuthors_Plus {
 	 */
 	public function fix_author_page( $selection ) {
 
+		global $wp_query, $authordata;
+
 		if ( ! isset( $wp_query ) ) {
 			return;
 		}
@@ -1147,9 +1149,6 @@ class CoAuthors_Plus {
 		}
 
 		$author = $this->get_coauthor_by( 'user_nicename', $author_name );
-
-		global $wp_query, $authordata;
-
 		if ( is_object( $author ) ) {
 			$authordata = $author; //phpcs:ignore
 			$term       = $this->get_author_term( $authordata );
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1133,8 +1133,6 @@ class CoAuthors_Plus {
 	 */
 	public function fix_author_page( $selection ) {
 
-		global $wp_query, $authordata;
- 
 		if ( ! isset( $wp_query ) ) {
 			return;
 		}
@@ -1149,6 +1147,8 @@ class CoAuthors_Plus {
 		}
 
 		$author = $this->get_coauthor_by( 'user_nicename', $author_name );
+		
+		global $wp_query, $authordata;
 
 		if ( is_object( $author ) ) {
 			$authordata = $author; //phpcs:ignore
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1135,7 +1135,7 @@ class CoAuthors_Plus {
 
 		global $wp_query, $authordata;
  
-		if ( ! isset( $wp_query ) ) {
+		if ( ! did_action( 'wp' ) ) {
 			return;
 		}
 


co-authors-plus.php
```diff
@@ -1133,6 +1133,12 @@ function current_user_can_set_authors( $post = null ) {
 	 */
 	public function fix_author_page( $selection ) {
 
+		global $wp_query, $authordata;
+ 
+		if ( ! isset( $wp_query ) ) {
+			return;
```
lkraav, line 1139: I monkey-patched my instance with this, and found it doesn't seem to help.

At least in our case, `$wp_query` is always set here, so this check does nothing to avoid executing further inappropriate code :thinking:

nielslange: @lkraav Can you tell me for which plugin you're seeing the error message, so that I can reproduce your case on my end?

In addition, I'll also try the solution that you suggested on https://github.com/Automattic/Co-Authors-Plus/issues/615#issuecomment-855820522.

lkraav: WCM === WooCommerce Memberships

Do you have a access to Woo commercial plugin copies?

nielslange: I do have access to the commercial WooCommerce plugins and strange enough, I do not see the problem appearing when using [WooCommerce Memberships](https://woocommerce.com/products/woocommerce-memberships/). That said, I've updated the PR using your code. That fix works fine on my end and I assume it works fine on your end as well. Can you confirm that? And can you confirm that we're talking about the same plugin?

lkraav: > I do have access to the commercial WooCommerce plugins and strange enough, I do not see the problem appearing when using WooCommerce Memberships.

Sentry traceback I posted appears like a generic bootstrap process, but traceback public view doesn't fails to show this particular event sample was triggered by a POST request, which may have some different code paths from GET requests.

WCM also caches membership post query results, so only initial request for that user may be triggering it, etc.

> Can you confirm that?

Everything confirmed. Proof from live server - middle box is 24h event bar chart, and you can see it die down immediately after monkey-patching with `did_action wp`.

image

One remaining thing perhaps: I have not personally verified anything about what this filter is originally meant to achieve in the first place - whether that functionality still works as intended. I would assume you've been able to take a look?

lkraav: PS I've been using `did_action()` timing strategy successfully elsewhere as well, and also seen it done in the ecosystem. It's not a random idea.

nielslange: > One remaining thing perhaps: I have not personally verified anything about what this filter is originally meant to achieve in the first place - whether that functionality still works as intended. I would assume you've been able to take a look?

Pippin posted a great explanation of what `did_action` is doing on https://pippinsplugins.com/detect-when-an-action-has-run-with-did_action/. That filter was new to me, and it simply counts how often a certain function was called. This is useful when creating CPT's to ensure that they're only created during the first run. In this particular case, we do not execute the `is_author` function if the `wp` action hasn't been executed yet. 

Regarding the functionality, all PHPUnit tests are running as expected, and I did not spot any problems after adding `did_action`.

lkraav: Great to hear, but I should've been more clear in my comment: concern was about original `posts_selection` action, whose logic we're modifying here.

Unit tests passing sounds good to me, didn't think to check for their existence in this repo.

lkraav:
> are we able to reproduce the issue that @lkraav is having? In my tests, using if ( ! isset( $wp_query ) fixes the errors.

Have you eyeballed our traceback sample @mikeyarce?

Here's how I see it: problem triggers very early in the request, `plugins_loaded` action.

image

WCM bootstraps itself, their My Account integration makes a posts query really early, and that's how we get `$wp_query` instantiated that early.

Why @nielslange is not able to easily repro this, not sure, there may be x+1 additional configuration things at play.

(WCM team's @unfulvio maybe able to provide additional insight whether this actually makes sense from WCM end to have a posts query happen this early? Maybe this is a weakness in their current My Account endpoint integration strategy.)

Regardless, `did_action wp` is still a more solid strategy and a delivers a specific intent execution timing signal. Even outside WCM, why should I not be able to do `$wp_query` stuff early if I really needed / wanted to?

nielslange:
> Why @nielslange is not able to easily repro this, not sure, there may be x+1 additional configuration things at play, or maybe he missed something in setup.

@lkraav Can you tell me how I can reproduce the error that you are seeing when using `! isset( $wp_query )` instead of `! did_action( 'wp' )`? Do you also see the error when using a fresh WordPress installation with the WooCommerce, WooCommerce Memberships and Co-Author Plugin plugins only?

co-authors-plus.php
rebeccahum: > At least in our case, $wp_query is always set here, so this check does nothing to avoid executing further inappropriate code 🤔

@lkraav Can you please tell me what the value of `$wp_query `is when it is set?

I feel like the `isset()` check for `$wp_query` is headed in the right direction. However, instead of checking if it's set, perhaps we can more explicitly check what the function expects `$wp_query` to be.

rebeccahum: > At least in our case, $wp_query is always set here, so this check does nothing to avoid executing further inappropriate code 🤔

@lkraav Can you please tell me what the value of `$wp_query `is when it is set? Is it a NULL or empty value?

rebeccahum:
I'm not able to reproduce this as well with WCM + CAP. Agreed with @mikeyarce that it's worth investigating to see what other factors are at play since I feel that the checking `$wp_query` approach makes more sense.

nielslange:
Hello @lkraav 👋 You might have missed my comment in https://github.com/Automattic/Co-Authors-Plus/pull/801#issuecomment-858575798. Could you let me know how to reproduce what you initially saw on your end? And do you see the same behaviour when testing the plugin isolated, as mentioned in my comment?

nielslange:
I did further testing, and I'm unable to reproduce @lkraav's error when using the CAP, WOO and WCM plugins on a fresh WP install.

Personally. I tend to use `! isset( $wp_query )` over `! did_action( 'wp' )` as I'm not entirely sure how the second function call is working. The check `! isset( $wp_query )` is [used within the `is_author()` function call](https://developer.wordpress.org/reference/functions/is_author/#source). 

In comparison, the `did_action()` function call also works with a non-existing tag, e.g. `did_action( 'foobar' )`. I also could not find a single `! did_action( 'wp' )` call within the core, while `! isset( $wp_query )` occurs 28 times in [query.php](https://github.com/WordPress/WordPress/blob/master/wp-includes/query.php).

lkraav:
I'm backlogged a bit, but haven't given up on providing more details on this with some xdebugging.

> In comparison, the `did_action()` function call also works with a non-existing tag, e.g. `did_action( 'foobar' )`. I also could not find a single `! did_action( 'wp' )` call within the core, while `! isset( $wp_query )` occurs 28 times in [query.php](https://github.com/WordPress/WordPress/blob/master/wp-includes/query.php).

Not sure if this finding is relevant much. Core is pre-optimized to time its `$wp_query` usage correctly.

nielslange:
> I'm backlogged a bit, but haven't given up on providing more details on this with some xdebugging.

@lkraav I assume you haven't had the time to xdebug and/or test the `! isset( $wp_query )` isolated in a fresh WP environment, have you?

nielslange:
@rebeccahum & @mikeyarce I reverted this PR to the original version in which I checked for the existence of `$wp_query`. Would you mind doing another review, if you have time to do so?

mikeyarce:
@nielslange, can you tell me your reasoning for moving `global $wp_query, $authordata;` up to above the new check?

nielslange:
@mikeyarce I thought I need `global $wp_query` before the check `! isset( $wp_query )`. To avoid defining global variables on various places within the function, I decided to move both global variables up. Having the global variables at the beginning of the function is also consistent with the global variables of the other functions in this file.

That said, I just run a quick test and noticed that I don't need to move the variables. Therefore, I moved them back to their original position to change only what's needed to solve this particular problem.

co-authors-plus.php
```diff
@@ -1143,7 +1147,7 @@ public function fix_author_page( $selection ) {
 		}
 
 		$author = $this->get_coauthor_by( 'user_nicename', $author_name );
-
+
```
rebeccahum, line 1150: Looks like unnecessary extra whitespace was added here.

rebeccahum (changes requested):

rebeccahum:
> That said, I just run a quick test and noticed that I don't need to move the variables. Therefore, I moved them back to their original position to change only what's needed to solve this particular problem.

How would `fix_author_page()` know to use the global `$wp_query` on the `isset()` check?

co-authors-plus.php
nielslange: My bad. I just removed the accidentally added spacing.

nielslange:
> How would `fix_author_page()` know to use the global `$wp_query` on the `isset()` check?

Ouch! 🤦‍♂️ Of course, I need to declare a global variable before I can use it.

For maintenance reasons, I decided not to use `global $wp_query` at the beginning of the function and `global $authordata` further down the function, but to keep `global $wp_query, $authordata` together. That way, it's easy to see which global variables are used within this function. I also see this pattern, of keeping the global variables together, when looking at the WP core files.

rebeccahum (approved):

PR #805, opened 2021-06-30
Prevent posts from being calculated twice for linked accounts

### LINKED ISSUE #779: Incorrect post count when linked accounts are authors of the same post prior to linking
**Steps to reproduce**

1. Create an editor account (editor-n-1)
2. Create one post (#n-1)
3. Create a guest account (guest-n-1)
4. Add guest author to post #n-1
5. Check post count for editor and guest
6. Link guest account (guest-n-1) with editor account (editor-n-1)
7. Check post count for editor and guest

**Expected behaviour**

✅ Editor: 1 post
✅ Guest: 1 post

**Current behaviour**

❌ Editor: 2 posts
❌ Guest: 2 posts

nielslange:
Fixes #779 

**Notes**

Prevent posts from being calculated twice for linked accounts.

**Steps to test**

1. Create an editor account (editor-779)
2. Create one post (test-post-779) using the editor account (editor-779)
3. Create a guest author account (guest-779)
4. Add guest author to post (test-post-779)
5. Check post count for editor and guest author
6. Link guest author account (guest-779) with editor account (editor-779)
7. Check that post count for both editor-779 and guest-779 shows 2, which is incorrect
8. Check out this PR
9. Check that post count for both editor-779 and guest-779 shows 1, which is expected

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -600,22 +600,23 @@ class CoAuthors_Plus {
 		$query = "SELECT COUNT({$wpdb->posts}.ID) FROM {$wpdb->posts}";
 
 		$query .= " LEFT JOIN {$wpdb->term_relationships} ON ({$wpdb->posts}.ID = {$wpdb->term_relationships}.object_id)";
-		$query .= " LEFT JOIN {$wpdb->term_taxonomy} ON ( {$wpdb->term_relationships}.term_taxonomy_id = {$wpdb->term_taxonomy}.term_taxonomy_id )";
-
-		$having_terms_and_authors = $having_terms = $wpdb->prepare( "{$wpdb->term_taxonomy}.term_id = %d", $term->term_id );
-		if ( 'wpuser' == $coauthor->type ) {
-			$having_terms_and_authors .= $wpdb->prepare( " OR {$wpdb->posts}.post_author = %d", $coauthor->ID );
-		}
-
+		$query .= " LEFT JOIN {$wpdb->terms} ON ({$wpdb->term_relationships}.term_taxonomy_id = {$wpdb->terms}.term_id)";
+		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+		$query .= " LEFT JOIN {$wpdb->users} ON ({$wpdb->posts}.post_author = {$wpdb->users}.ID)";
+	
 		$post_types = apply_filters( 'coauthors_count_published_post_types', array( 'post' ) );
 		$post_types = array_map( 'sanitize_key', $post_types );
 		$post_types = "'" . implode( "','", $post_types ) . "'";
 
-		$query .= " WHERE ({$having_terms_and_authors}) AND {$wpdb->posts}.post_type IN ({$post_types}) AND {$wpdb->posts}.post_status = 'publish'";
+		$query .= " WHERE {$wpdb->posts}.post_type IN ({$post_types})";
+		$query .= " AND {$wpdb->posts}.post_status = 'publish'";
+		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+		$query .= " AND ( {$wpdb->users}.user_login = %s OR {$wpdb->terms}.name = %s )";
+		
+		$query .= " GROUP BY {$wpdb->posts}.ID";
 
-		$query .= $wpdb->prepare( " GROUP BY {$wpdb->posts}.ID HAVING MAX( IF ( {$wpdb->term_taxonomy}.taxonomy = '%s', IF ( {$having_terms},2,1 ),0 ) ) <> 1 ", $this->coauthor_taxonomy ); //phpcs:ignore
-
-		$count = $wpdb->query( $query ); // phpcs:ignore
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		$count = $wpdb->query( $wpdb->prepare( $query, $coauthor->linked_account, $coauthor->user_login ) ); 
 		$wpdb->update( $wpdb->term_taxonomy, array( 'count' => $count ), array( 'term_taxonomy_id' => $term->term_taxonomy_id ) );
 
 		wp_cache_delete( 'author-term-' . $coauthor->user_nicename, 'co-authors-plus' );
@@ -1060,13 +1061,8 @@ class CoAuthors_Plus {
 
 		$term = $this->get_author_term( $coauthor );
 
+		// Return $term->count, if $term exists.
 		if ( is_object( $term ) ) {
-			// Return combined post count, if account is linked.
-			if ( strlen( $coauthor->linked_account ) > 2 ) {
-				return $count + $term->count;
-			}
-
-			// Otherwise, return the term count.
 			return $term->count;
 		}
 


co-authors-plus.php
```diff
@@ -600,22 +600,23 @@ public function update_author_term_post_count( $term ) {
...
+		$query .= " WHERE {$wpdb->posts}.post_type IN ({$post_types})";
+		$query .= " AND {$wpdb->posts}.post_status = 'publish'";
+		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+		$query .= " AND ( {$wpdb->users}.user_login = %s OR {$wpdb->terms}.name = %s )";
+		
+		$query .= " GROUP BY {$wpdb->posts}.ID";
 
-		$count = $wpdb->query( $query ); // phpcs:ignore
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		$count = $wpdb->query( $wpdb->prepare( $query, $coauthor->linked_account, $coauthor->user_login ) );
```
rebeccahum, line 619: This assumes that there is a linked account, but what if there's no linked account?

rebeccahum (changes requested):

rebeccahum:
👋 In my testing using the latest CAP version, I'm not able to reproduce this below step:

> Check that post count for both editor-779 and guest-779 shows 2, which is incorrect

I see that the post count is 1 for both.

co-authors-plus.php
nielslange: I should have mentioned that I already checked what happens if there's no linked account. 🤭

When there's no account linked, the variable `$coauthor->linked_account` returns an empty string. Testing this with a post that had been created as `admin` with no linked account, the query looks as follows:

```sql
SELECT COUNT(wp_posts.ID)
FROM wp_posts
LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id)
LEFT JOIN wp_terms ON (wp_term_relationships.term_taxonomy_id = wp_terms.term_id)
LEFT JOIN wp_users ON (wp_posts.post_author = wp_users.ID)
WHERE wp_posts.post_type IN ('post')
AND wp_posts.post_status = 'publish'
AND ( wp_users.user_login = '' OR wp_terms.name = 'admin' )
GROUP BY wp_posts.ID
```

I thought about checking if `$coauthor->linked_account` actually has a value, but given that it returns an empty string which then leads to the correct post count, I decided against it. 

Would you advise practising defensive coding and checking if `$coauthor->linked_account` has a value other than an empty string?

co-authors-plus.php
```diff
@@ -600,22 +600,23 @@ public function update_author_term_post_count( $term ) {
...
+		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+		$query .= " LEFT JOIN {$wpdb->users} ON ({$wpdb->posts}.post_author = {$wpdb->users}.ID)";
+	
 		$post_types = apply_filters( 'coauthors_count_published_post_types', array( 'post' ) );
 		$post_types = array_map( 'sanitize_key', $post_types );
 		$post_types = "'" . implode( "','", $post_types ) . "'";
 
-		$query .= " WHERE ({$having_terms_and_authors}) AND {$wpdb->posts}.post_type IN ({$post_types}) AND {$wpdb->posts}.post_status = 'publish'";
-
-		$query .= $wpdb->prepare( " GROUP BY {$wpdb->posts}.ID HAVING MAX( IF ( {$wpdb->term_taxonomy}.taxonomy = '%s', IF ( {$having_terms},2,1 ),0 ) ) <> 1 ", $this->coauthor_taxonomy ); //phpcs:ignore
```
mikeyarce, line 616: Can you tell me your reasoning here behind removing the `HAVING MAX( IF( IF (` section of the query?

nielslange: My understanding of the `HAVING` statement is to further filter results of a previous `GROUP BY` statement. 

For example, I could group the salaries of all employees with a `GROUP BY` statement and then using a `HAVING` statement to only get the salaries that fulfil additional criteria.

In this particular case here, I already group by post ID. Looking at https://codex.wordpress.org/Database_Description#Indexes_6, I see that the post ID is the primary key of the wp_posts table. Therefore, the ID is already unique. Adding a `HAVING` statement after the `GROUP BY` statement is therefore obsolete in my opinion, unless I missed something in this case.

mikeyarce:
@nielslange - I'm getting varied results with just the steps listed above.  When I run a manual recount using `wp co-authors-plus update-author-terms` then the numbers are accurate, but without running that the numbers vary.  Is the count expected to be right immediately or does a specific action have to happen for the count to be updated?

I'm also still concerned about dropping the `HAVING` statement - can you think of why that might have been added?  What if the ID we're looking at is no longer an `author` term?  Can that happen?

Along with what @rebeccahum said - is there something we could do to prevent doing a query for an empty value?

nielslange:
> I'm getting varied results with just the steps listed above. When I run a manual recount using wp co-authors-plus update-author-terms then the numbers are accurate, but without running that the numbers vary. Is the count expected to be right immediately or does a specific action have to happen for the count to be updated?

On my end, I'm seeing the correct result immediately after linking a guest account to a regular account. That said, I noticed that I phrased one step poorly. I wrote _"2. Create one post (test-post-779)"_ which should have been _"2. Create one post (test-post-779) using the editor account (editor-779)"_. I just updated that to ensure that the test post will be created using the editor account rather than the default admin account.

> I'm also still concerned about dropping the HAVING statement - can you think of why that might have been added?

I've added `define( 'SAVEQUERIES', true );` to display the initial SQL query:

```sql
SELECT COUNT(wp_posts.ID)
FROM wp_posts
LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id)
LEFT JOIN wp_term_taxonomy ON ( wp_term_relationships.term_taxonomy_id = wp_term_taxonomy.term_taxonomy_id )
WHERE (wp_term_taxonomy.term_id = 4)
AND wp_posts.post_type IN ('post')
AND wp_posts.post_status = 'publish'
GROUP BY wp_posts.ID
HAVING MAX( IF ( wp_term_taxonomy.taxonomy = 'author', IF ( wp_term_taxonomy.term_id = 4,2,1 ),0 ) ) <> 1 
```

I'm not entirely sure about the purpose of the `HAVING` statement. I see that if `wp_term_taxonomy.taxonomy` equals `author`, the second `IF` statement is used. Otherwise, 0 gets returned. The second `IF` statement checks if the `wp_term_taxonomy.term_id` equals the value of `$this->coauthor_taxonomy`. If it does, the statement returns a 2, otherwise a 1. This leads to the following logic:

- If the `wp_term_taxonomy.taxonomy` is `author` and the `wp_term_taxonomy.term_id` matches, return 2.
- If the `wp_term_taxonomy.taxonomy` is `author` and the `wp_term_taxonomy.term_id` does not match, return 1.
- If the `wp_term_taxonomy.taxonomy` is not `author`, return 0.

However, I'm then struggling to understand the `MAX` statement. If I'm not mistaken, the following `HAVING` queries can occur:

- `HAVING MAX( 2 ) <> 1`
- `HAVING MAX( 1 ) <> 1`
- `HAVING MAX( 0 ) <> 1`

According to https://www.w3schools.com/sql/sql_operators.asp, the `<>` operator stands for `not equal to`. Could you point me in the right direction to understand the purpose of the `HAVING` statement?

> What if the ID we're looking at is no longer an `author` term? Can that happen?

When debugging this function call both with a linked account and an unlinked account, I receive `author` terms in both cases. 

**Author terms for linked account:**

```
/Users/nielslange/Development/plugins/Co-Authors-Plus/co-authors-plus.php:600:
object(stdClass)[1034]
  public 'ID' => string '9' (length=1)
  public 'display_name' => string 'guest-779' (length=9)
  public 'first_name' => string '' (length=0)
  public 'last_name' => string '' (length=0)
  public 'user_login' => string 'guest-779' (length=9)
  public 'user_email' => string '' (length=0)
  public 'linked_account' => string 'editor-779' (length=10)
  public 'website' => string '' (length=0)
  public 'aim' => string '' (length=0)
  public 'yahooim' => string '' (length=0)
  public 'jabber' => string '' (length=0)
  public 'description' => string '' (length=0)
  public 'user_nicename' => string 'guest-779' (length=9)
  public 'type' => string 'guest-author' (length=12)
```

**Author terms for unlinked account:**

```
/Users/nielslange/Development/plugins/Co-Authors-Plus/co-authors-plus.php:600:
object(stdClass)[1017]
  public 'ID' => string '9' (length=1)
  public 'display_name' => string 'guest-779' (length=9)
  public 'first_name' => string '' (length=0)
  public 'last_name' => string '' (length=0)
  public 'user_login' => string 'guest-779' (length=9)
  public 'user_email' => string '' (length=0)
  public 'linked_account' => string '' (length=0)
  public 'website' => string '' (length=0)
  public 'aim' => string '' (length=0)
  public 'yahooim' => string '' (length=0)
  public 'jabber' => string '' (length=0)
  public 'description' => string '' (length=0)
  public 'user_nicename' => string 'guest-779' (length=9)
  public 'type' => string 'guest-author' (length=12)
```

Is this what you meant, or have I misunderstood you? 

> Along with what @rebeccahum said - is there something we could do to prevent doing a query for an empty value?

I could practice defensive coding and alter the query depending on if the value is empty or not, e.g.

```
if ( '' !== $coauthor->linked_account ) {
  $query .= " AND ( {$wpdb->users}.user_login = %s OR {$wpdb->terms}.name = %s )";
} else {
  $query .= " AND ( {$wpdb->users}.user_login = %s";
}

if ( '' !== $coauthor->linked_account ) {
  $count = $wpdb->query( $wpdb->prepare( $query, $coauthor->linked_account, $coauthor->user_login ) ); 
} else {
  $count = $wpdb->query( $wpdb->prepare( $query, $coauthor->user_login ) ); 
}
```

What do you think about that suggestion?

mikeyarce:
@nielslange good questions!
> HAVING MAX( IF ( wp_term_taxonomy.taxonomy = 'author', IF ( wp_term_taxonomy.term_id = 4,2,1 ),0 ) ) <> 1 

If I take your logic statements and put them together I read: `If the ID we're checking does have the taxonomy of author and if that author id does match the id we have above - return this`  - taking that part out might mean that we could return things that might not be `author`'s in the taxonomy?   When could that happen?  I'm not really sure, but if we do want to remove I'd like to have a good case for it and as to why we can't keep it if it provides some extra checks.

> if ( '' !== $coauthor->linked_account ) {
  $query .= " AND ( {$wpdb->users}.user_login = %s OR {$wpdb->terms}.name = %s )";
} else {
  $query .= " AND ( {$wpdb->users}.user_login = %s";
}

For this part - I agree it would make sense to alter the query if no accounts are linked, but I'm not sure if this logic above is the right one ^

nielslange:
Hi @mikeyarce 👋

I'm a bit lost with this PR and think it's best if someone else is taking care of this. I'll close this PR for now.

Issue #687, opened 2019-07-08
Author page SQL error (front-end and back-end)


rebeccahum:
Reproduction steps on a fresh WP local install:

1) Login with user ID 1 (in my case, this was the default admin account, but for illustration, we'll call it user A)
2) Create post directly from account as User A
3) Create user B with user ID of 12 (or anything that prepends with "1") and user role of "Contributor"
4) Login as user B and view author archive for User A (example.com/wp-admin/edit.php?author_name=admin) and you'll see the below error:

```
WordPress database error: [You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '1 AND wp_posts.post_status = 'private') GROUP BY wp_posts.ID HAVING MAX( IF ( wp' at line 1]
SELECT SQL_CALC_FOUND_ROWS wp_posts.ID FROM wp_posts LEFT JOIN wp_term_relationships AS tr1 ON (wp_posts.ID = tr1.object_id) LEFT JOIN wp_term_taxonomy ON ( tr1.term_taxonomy_id = wp_term_taxonomy.term_taxonomy_id ) WHERE 1=1 AND ((wp_posts.post_author = 1 OR (wp_term_taxonomy.taxonomy = 'author' AND wp_term_taxonomy.term_id = '12'))) AND wp_posts.post_type = 'post' AND (wp_posts.post_status = 'publish' OR wp_posts.post_status = 'future' OR wp_posts.post_status = 'draft' OR wp_posts.post_status = 'pending' OR (wp_posts.post_author = 1 OR (wp_term_taxonomy.taxonomy = 'author' AND wp_term_taxonomy.term_id = '12'))1 AND wp_posts.post_status = 'private') GROUP BY wp_posts.ID HAVING MAX( IF ( wp_term_taxonomy.taxonomy = 'author', IF ( wp_term_taxonomy.term_id = '12' OR wp_term_taxonomy.term_id = '15',2,1 ),0 ) ) <> 1 ORDER BY wp_posts.post_date DESC LIMIT 0, 20
```

PR #818, opened 2021-08-13
Fix author page SQL error

lschuyler:
Change the `preg_replace` limit to 1, rather than unlimited (-1), to prevent additional pattern matches that resulted in a broken WHERE statement for the author page query.

Fixes issue #687 .

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -740,7 +740,7 @@ class CoAuthors_Plus {
 
 					$maybe_both_query = $maybe_both ? '$0 OR' : '';
 
-					$where = preg_replace( '/\s\b(?:' . $wpdb->posts . '\.)?post_author\s*IN\s*\(' . $id . '\)/', ' (' . $maybe_both_query . ' ' . $terms_implode . ')', $where, 1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
+					$where = preg_replace( '/\s\b(?:' . $wpdb->posts . '\.)?post_author\s*IN\s*\(' . $id . '\)/', ' (' . $maybe_both_query . ' ' . $terms_implode . ')', $where, -1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
 
 				} else {
 					$where = preg_replace( '/(\b(?:' . $wpdb->posts . '\.)?post_author\s*=\s*(' . $id . '))/', '(' . $maybe_both_query . ' ' . $terms_implode . ')', $where, 1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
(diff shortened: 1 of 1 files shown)


co-authors-plus.php
```diff
@@ -759,7 +759,7 @@ function posts_where_filter( $where, $query ) {
 						$current_user_query  = $wpdb->term_taxonomy . '.taxonomy = \'' . $this->coauthor_taxonomy . '\' AND ' . $wpdb->term_taxonomy . '.term_id = \'' . $current_coauthor_term->term_id . '\'';
 						$this->having_terms .= ' ' . $wpdb->term_taxonomy . '.term_id = \'' . $current_coauthor_term->term_id . '\' OR ';
 
-						$where = preg_replace( '/(\b(?:' . $wpdb->posts . '\.)?post_author\s*=\s*(' . get_current_user_id() . ') )/', $current_user_query . ' ', $where, -1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND}
+						$where = preg_replace( '/(\b(?:' . $wpdb->posts . '\.)?post_author\s*=\s*(' . get_current_user_id() . ') )/', $current_user_query . ' ', $where, 1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND}
```
rebeccahum, line 762: Do you see any other use cases or places in the code where the replacement would only need to be done once as well to prevent the SQL error from being triggered?

rebeccahum (changes requested):

co-authors-plus.php
lschuyler: Actually, yes! There are 3 places that would benefit from this fix. I've changed the additional two in https://github.com/Automattic/Co-Authors-Plus/commit/3a31a5698dfff889af4f9d7666f448880702cbc2 .

co-authors-plus.php
```diff
@@ -740,10 +740,10 @@ function posts_where_filter( $where, $query ) {
 
 					$maybe_both_query = $maybe_both ? '$0 OR' : '';
 
-					$where = preg_replace( '/\s\b(?:' . $wpdb->posts . '\.)?post_author\s*IN\s*\(' . $id . '\)/', ' (' . $maybe_both_query . ' ' . $terms_implode . ')', $where, -1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
+					$where = preg_replace( '/\s\b(?:' . $wpdb->posts . '\.)?post_author\s*IN\s*\(' . $id . '\)/', ' (' . $maybe_both_query . ' ' . $terms_implode . ')', $where, 1 ); // ' . $wpdb->postmeta . '.meta_id IS NOT NULL AND
```
rebeccahum, line 743: Were you able to trigger in your testing this particular section of the code? Curious because it slightly differs from the other `$where` statements on the pattern matching.

lschuyler: Yes, you can trigger this by logging into a Contributor account and looking at `/wp-admin/edit.php?author_name=admin` however with additional testing, this change does not fix or improve anything, so I'm going to revert this proposed changed just for this line.

rebeccahum (approved):
🚢 LGTM!

Issue #825, opened 2021-10-22
CSS hack for hiding Gutenberg post_author no longer works redux


gamebits:
This is a recurrence of https://github.com/Automattic/Co-Authors-Plus/issues/791 that was addressed in https://github.com/Automattic/Co-Authors-Plus/pull/793. The "Author" dropdown is again being displayed, despite Co-Authors Plus being active and inserting its own "Authors" field.

![wp-author](https://user-images.githubusercontent.com/12199101/138466406-ea621d4f-8a07-4d5b-9183-9b19be8cfb13.jpg)

![cap-authors](https://user-images.githubusercontent.com/12199101/138466412-275d0618-3f5a-4b9c-a6c0-18057ef38edd.jpg)

I've thus far been unsuccessful in determining what change to WordPress or Gutenberg broke the prior fix, or what new CSS should be targeted.

![wp-authors-css](https://user-images.githubusercontent.com/12199101/138466782-0ae8854f-85c4-4723-8d61-03e02d8e201f.jpg)

lschuyler:
With the changes to the author combobox in the editor from #790, I don't think this will be needed anymore, and the CSS hack introduced in #791 can likely be removed too. I'll test!

lschuyler:
Confirmed. We aren't using the author box from core anymore, so this is no longer a relevant issue.

CSS hack removed in #847.

PR #826, opened 2021-10-26
Fix CSS hack for hiding Gutenberg post author box

lschuyler:
Expand the CSS hack to hide the Gutenberg post author selector combobox when over 25 authors exist on the site, as reported in [issue 825](https://github.com/Automattic/Co-Authors-Plus/issues/825).

Gutenberg in WordPress core changes the type of author select combobox when there are [25 or more authors](https://github.com/WordPress/gutenberg/blob/d8fa5000cb00c02b455124b0e34b8ac382fe31e4/packages/editor/src/components/post-author/index.js#L14), and this isn't being hidden by the previous CSS hide hack. 

So where the current CSS hack works, as soon as there are over 25 authors, the type of select box changes, and the hack no longer targets the CSS class.

To fix, this hack can be expanded to also hide this >25 author select combobox class that Gutenberg uses, however there isn't a precise ID or class that can be targeted, so there is a possibility that this will hide any similar combobox added to the same area of the edit page by any other plugin.

### DIFF
--- a/css/co-authors-plus.css
+++ b/css/co-authors-plus.css
@@ -113,6 +113,7 @@
 /** Block Editor Hack for 5.0: Hide the core author input **/
 .block-editor label[for^="post-author-selector-"],
 .block-editor select[id^="post-author-selector-"],
+.block-editor .edit-post-sidebar .components-combobox-control,
 .block-editor .post-author-selector {
 	display: none;
 }


lschuyler:
I've submitted a PR to Gutenberg to add a class name to the post author component of the sidebar that would make this super easy and not hacky. 

https://github.com/WordPress/gutenberg/pull/36269

I'll update this PR when it has been merged.

rebeccahum:
@lshuyer You ready to update this PR? I see it's been merged now. 😉

lschuyler:
This is no longer necessary. 

As of https://github.com/Automattic/Co-Authors-Plus/pull/790, the authors box in the core block editor is prevented from showing [here](https://github.com/Automattic/Co-Authors-Plus/blob/cf8bc5ab87cd463cb985301078af0be5fd75b320/php/class-coauthors-endpoint.php#L284) and a
Co-Authors Plus specific author [combobox](https://github.com/Automattic/Co-Authors-Plus/blob/master/src/components/AuthorsSelection.js) has been added.

PR #831, opened 2021-11-11
Fix broken SQL query caused by wpdb->prepare() escaping quotes

josejaureguiberry:
The single quotes in the comma separate array passed to the IN clause was escaped by wpdb->prepare(), retrieving something like this:
`SELECT ID FROM wp_posts WHERE post_author=0 AND post_type IN ('posts\',\'pages\',\'review')`

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -233,7 +233,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 		}
 
 		$post_types = implode( "','", $coauthors_plus->supported_post_types );
-		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN (%s)", $user->ID, $post_types ) );
+		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('$post_types')", $user->ID ) );
 		$affected   = 0;
 		foreach ( $posts as $post_id ) {
 			$coauthors = cap_get_coauthor_terms_for_post( $post_id );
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -233,7 +233,7 @@ public function assign_user_to_coauthor( $args, $assoc_args ) {
 		}
 
 		$post_types = implode( "','", $coauthors_plus->supported_post_types );
-		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN (%s)", $user->ID, $post_types ) );
+		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('" . $post_types . "')", $user->ID ) );
```
rebeccahum, line 236: Do you mind throwing a `// phpcs:ignore` comment for the ruleset this does not need? Also, minor, but let's use interpolation over concatenation for better readability.

rebeccahum (changes requested):

php/class-wp-cli.php
josejaureguiberry: @rebeccahum changes ready for review

php/class-wp-cli.php
```diff
@@ -233,7 +233,8 @@ public function assign_user_to_coauthor( $args, $assoc_args ) {
 		}
 
 		$post_types = implode( "','", $coauthors_plus->supported_post_types );
-		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('" . $post_types . "')", $user->ID ) );
+		// phpcs:ignore
```
rebeccahum, line 236: Do you mind using the comment to target the specific rule? Thanks!

php/class-wp-cli.php
```diff
@@ -233,7 +233,8 @@ public function assign_user_to_coauthor( $args, $assoc_args ) {
...
+		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('$post_types')", $user->ID ) );
```
rebeccahum, line 237:
```suggestion
		$posts      = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ({$post_types})", $user->ID ) );
```

rebeccahum (changes requested):

rebeccahum (approved):

PR #832, opened 2021-12-02
Issue with cache key `id` not being deleted.

mauteri:
`get_guest_author_by` method in `CoAuthors_Guest_Authors` allows the key `id`. It caches the result using the key as park of the cache key. The only problem is that the `delete_guest_author_cache` method does not recognize `id` as a valid key, so that cache remains indefinitely. This PR fixes the issue with `id` cache not being deleted.


rebeccahum (approved):
Thanks!

PR #843, opened 2022-04-25
Fix error for undefined `use_block_editor_for_post`

### LINKED ISSUE #842: PHP Fatal error: Uncaught Error: Call to undefined function CoAuthors\API\use_block_editor_for_post() in /wp-content/plugins/co-authors-plus/php/class-coauthors-endpoint.php:292
Wordpress version 5.9.2
Co-Author-Plus version 3.5

When editing a page entry, located in the sidebar under "Page Attributes". Modifying the value of the pages parent we are receiving Error 500 with xhr request:
/wp-json/wp/v2/pages?context=edit&per_page=100&exclude=12345&parent_exclude=12345&orderby=menu_order&order=asc&_fields=id%2Ctitle%2Cparent&_locale=user

Error log
`PHP Fatal error: Uncaught Error: Call to undefined function CoAuthors\API\use_block_editor_for_post() in /wp-content/plugins/co-authors-plus/php/class-coauthors-endpoint.php:292`

Appears to be a name space issue

lschuyler:
Fix the fatal error caused by a call to an undefined function `use_block_editor_for_post`, by loading the WordPress Post Administration API (/wp-admin/includes/post.php) first, and then refers to the `use_block_editor_for_post`function in the global namespace.

Fixes #842.

### DIFF
--- a/php/class-coauthors-endpoint.php
+++ b/php/class-coauthors-endpoint.php
@@ -236,7 +236,7 @@ class Endpoints {
 	 * Get authors' data and add it to the response.
 	 *
 	 * @param array The response array.
-	 * @param int   Thet post ID from the request.
+	 * @param int   The post ID from the request.
 	 */
 	public function _build_authors_response( &$response, $request ) {
 		$authors = get_coauthors( $request->get_param( 'post_id' ) );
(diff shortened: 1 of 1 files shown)


rebeccahum (approved):
Awesome!

PR #851, opened 2022-06-01
Remove the /wp/v2/coauthors endpoint

lschuyler:
Remove the `/wp/v2/coauthors` endpoint added in #790 that is not being used.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -226,9 +226,7 @@ class CoAuthors_Plus {
 			'public'       => false,
 			'sort'         => true,
 			'args'         => array( 'orderby' => 'term_order' ),
-			'show_in_rest' => true,
 			'show_ui'      => false,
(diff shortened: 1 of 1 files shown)


rebeccahum (approved):
Thanks!

Antonio-Laguna:
@rebeccahum and @lschuyler this seems to be like a breaking change. This prevents accessing this information for Gutenberg blocks. Can't query anymore for this data. Was that intentional?

sbcatania:
This endpoint was really useful! Could it please be added back?? There is no other way to access co-authored post data from the API now.

sbcatania:
@lschuyler and @rebeccahum Thank you so much for your contributions to this plugin!! By any chance, is it possible this change could be reversed? We depended on this endpoint to power some custom plugins we've created and our mobile app and I think many others use this endpoint too so I'm concerned this is a breaking change.

Issue #703, opened 2019-08-29
302 redirect to /404 for non-existent paged Author archive


kclarke10:
I'm talking about non-existent paged Author archives -> (like for example: https://domain.com/author/mike/page/15/, where mike has only 100 posts with wp set to 10 posts per page)

When Co Authors is enabled, instead of just doing a 404 on that URL, it sends a 302 redirect to domain.com/404 which then returns a 404 code.  

Only noticed this because Google has been trying to crawling domain.com/404 thousands of times when it decides to probe these paged archives that don't exist and have never existed.   This wastes crawl budget.

GaryJones:
Fixed with #874.

PR #874, opened 2022-10-19
Remove pre_handle_404 filter

maximomartinezsoria:
Problem: non-existent paged author archives are not throwing a 404. For example: /author/slug/page/100000. Will load and be indexed by crawlers.

Others are having this problem too: https://github.com/Automattic/Co-Authors-Plus/issues/703

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1209,7 +1209,6 @@ class CoAuthors_Plus {
 		if ( is_object( $authordata ) || ! empty( $term ) ) {
 			$wp_query->queried_object    = $authordata;
 			$wp_query->queried_object_id = $authordata->ID;
-			add_filter( 'pre_handle_404', '__return_true' );
 		} else {
 			$wp_query->queried_object = $wp_query->queried_object_id = null;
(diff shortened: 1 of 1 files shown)


leogermani:
This would add a regression. See https://github.com/Automattic/Co-Authors-Plus/issues/425 and https://github.com/Automattic/Co-Authors-Plus/commit/09a1c50f7f0e70a730cae86f3366ca140cdbdc0e

So what we need is:

We need to ensure the same behavior for regular users and Guest authors.

- When you visit the authors's page, never return 404, even if the author or guest author do not have any posts yet
- When you visit the author's page with `paged` in a range outside the range, return 404

I did a quick test here and if I think it's just a matter of surrounding that filter in a `is_paged()` check:

```PHP
if ( ! is_paged() ) {
	add_filter( 'pre_handle_404', '__return_true' );
}
```

Bonus points if you are able to update the tests on `test-author-queried-object.php` to check for all these cases

maximomartinezsoria:
Thanks for your answer!

Gusts authors behavior matching core behavior is already tested so I just added a new test for pagination.

leogermani (approved):
LGTM! Thanks!

PR #877, opened 2022-10-25
Adds the Yoast integration

leogermani:
This PR adds the Yoast integration.

This integration was initially developed as part of Yoast itself -> https://github.com/Yoast/wordpress-seo/pull/18331

I then added support to Guest Authors and did a couple of other improvements in -> https://github.com/Yoast/wordpress-seo/pull/19076

However, while chatting with @jdevalk, we agreed that it would be better if this integration lived in the CoAuthors plugin.

So this PR moves the whole integration, with my additions to the CoAuthors codebase.

Note: This only acts on Yoast filters so it should have no effect on sites not running Yoast.

## Test instructions

This PR can be acceptance tested by following these steps:

* Enable Yoast and CoAuthors Plus plugin
* Make sure you DO NOT have the `YOAST_SEO_COAUTHORS_PLUS` constant set
* Create a post

Now try different authorship states and check the resulting schema:

* Visit the post you created and look at the source code
* We want to look at the schema that is inserted in the `<script type="application/ld+json" class="yoast-schema-graph">` tag
* Make sure the "author" key has a list of IDs and they match the authors you have set up for the post:

```
"@graph": [
	        {
	            "@type": "Article",
	            "@id": "https://leogermani.jurassic.tube/2022/09/15/venenatis-dapibus-odio-sollicitudin-ultricies-magna-sem-hac-hendrerit-cursus-faucibus/#article",
	            "isPartOf": {
	                "@id": "https://leogermani.jurassic.tube/2022/09/15/venenatis-dapibus-odio-sollicitudin-ultricies-magna-sem-hac-hendrerit-cursus-faucibus/"
	            },
	            "author": [
	                {
	                    "@id": "https://leogermani.jurassic.tube/#/schema/person/743c49680386c3bcf517e1753cd46a25"
	                }
	            ],
 ...
```
* The example above shows a post with only one author
* At the end of the schema, look for the `Person` sections. There should be one section for each author you added to the post. Example:
```
	        {
	            "@type": "Person",
	            "@id": "https://leogermani.jurassic.tube/#/schema/person/743c49680386c3bcf517e1753cd46a25",
	            "name": "leo",
	            "image": {
	                "@type": "ImageObject",
	                "inLanguage": "en-US",
	                "@id": "https://leogermani.jurassic.tube/#/schema/person/image/00d99ee8603d99603d8c6be4dbc57016",
	                "url": "https://secure.gravatar.com/avatar/ea8b076b398ee48b71cfaecf898c582b?s=96&d=mm&r=g",
	                "contentUrl": "https://secure.gravatar.com/avatar/ea8b076b398ee48b71cfaecf898c582b?s=96&d=mm&r=g",
	                "caption": "leo"
	            },
	            "sameAs": [
	                "http://localhost"
	            ],
	            "mainEntityOfPage": {
	                "@id": "https://leogermani.jurassic.tube/author/leo/"
	            }
	        }
```
* The persons in the list must match the IDs in the author key we saw above. 
* Confirm that the `<meta name="author" ..` tag correctly displays the name of all authors
* Repeat the above steps changing the post authorship with different combinations:
  * only one regular author
  * 2 or more authors who are regular users
  * mix of regular users and guest authors as authors of the post
  * only one guest author as the post author
  * multiple guest authors as post authors
  * Also test guest authors with and without avatars, and with and without emails with gravatars. Check that the image tag is properly added in all cases.
 * Visit an author archive page for both a regular author and a guest author and make sure the schema for `profilePage` has the right links

### DIFF
--- a/php/integrations/yoast.php
+++ b/php/integrations/yoast.php
@@ -9,6 +9,7 @@ use CoAuthors\Integrations\Yoast\CoAuthor;
 use Yoast\WP\SEO\Config\Schema_Types;
 use Yoast\WP\SEO\Context\Meta_Tags_Context;
 use Yoast\WP\SEO\Generators\Schema\Abstract_Schema_Piece;
+use WP_User;
 
 /**
  * The main Yoast integration class
@@ -77,10 +78,10 @@ class Yoast {
 	 * @return void
 	 */
 	public static function register_hooks() {
-		\add_filter( 'wpseo_schema_graph', [ __CLASS__, 'filter_graph' ], 11, 2 );
-		\add_filter( 'wpseo_schema_author', [ __CLASS__, 'filter_author_graph' ], 11, 4 );
-		\add_filter( 'wpseo_schema_profilepage', [ __CLASS__, 'filter_schema_profilepage' ], 11, 4 );
-		\add_filter( 'wpseo_meta_author', [ __CLASS__, 'filter_author_meta' ], 11, 2 );
+		add_filter( 'wpseo_schema_graph', [ __CLASS__, 'filter_graph' ], 11, 2 );
+		add_filter( 'wpseo_schema_author', [ __CLASS__, 'filter_author_graph' ], 11, 4 );
+		add_filter( 'wpseo_schema_profilepage', [ __CLASS__, 'filter_schema_profilepage' ], 11, 4 );
+		add_filter( 'wpseo_meta_author', [ __CLASS__, 'filter_author_meta' ], 11, 2 );
 	}
 
 	/**
@@ -99,14 +100,14 @@ class Yoast {
 			return $data;
 		}
 
-		$user = \get_queried_object();
+		$user = get_queried_object();
 
 		if ( empty( $user->type ) || $user->type !== 'guest-author' ) {
 			return $data;
 		}
 
 		// Fix author URL.
-		$author_url                                     = \get_author_posts_url( $user->ID, $user->user_nicename );
+		$author_url                                     = get_author_posts_url( $user->ID, $user->user_nicename );
 		$graph_piece_generator->context->canonical      = $author_url;
 		$graph_piece_generator->context->main_schema_id = $author_url;
 
@@ -129,11 +130,11 @@ class Yoast {
 		}
 
 		if ( isset( $data['image']['@id'] ) ) {
-			$data['image']['@id'] .= \md5( $data['image']['url'] );
+			$data['image']['@id'] .= md5( $data['image']['url'] );
 		}
 
 		if ( isset( $data['logo']['@id'] ) ) {
-			$data['logo']['@id'] .= \md5( $data['image']['url'] );
+			$data['logo']['@id'] .= md5( $data['image']['url'] );
 		}
 
 		return $data;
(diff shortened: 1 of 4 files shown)


php/integrations/yoast.php
```diff
@@ -0,0 +1,243 @@
...
+			if ( \in_array( $piece['@type'], $article_types, true ) ) {
+				$data[ $key ]['author'] = $ids;
+				$add_to_graph           = true;
+				break;
+			}
+		}
+
+		if ( $add_to_graph ) {
+			// Clean all Persons from the schema, as the user stored as post owner might be incorrectly added if the post post has only guest authors as authors.
+			$data = \array_filter(
```
dkoo, line 201: Very much a nitpick, but our PHP convention in Newspack is to not globally namespace core PHP functions—only global functions like those from WP or other layers on top of PHP.

php/integrations/yoast.php
```diff
@@ -0,0 +1,243 @@
...
+		 */
+		$author_objects = \get_coauthors( $context->post->ID );
+
+		$ids     = [];
+		$authors = [];
+
+		// Add the authors to the schema.
+		foreach ( $author_objects as $author ) {
+			$author_generator          = new CoAuthor();
+			$author_generator->context = $context;
```
dkoo, line 172: If context and Yoast helpers are required for the `CoAuthor` class, why not add them on the class constructor? Just in case anyone tries to instantiate a `CoAuthors\Integrations\Yoast\CoAuthor` class without these items.

dkoo (dismissed):
Works exactly as described. Left a couple of very nitpicky style questions that you can feel free to ignore if they're not relevant.

php/integrations/yoast.php
leogermani: This comes by influence of how they do things in Yoast, and since I was writing it for Yoast I tried to follow their standards. I'm gonna check and see how we do this in CAP

php/integrations/yoast.php
leogermani: Makes sense. I'll have a look. Again something that comes from Yoast and this particular piece not written by me

leogermani: So, the context I only have in this context, so I can't add this into the constructor, unless I make it a required argument.

But this `CoAuthor` class extends Yoasts' `Author` which extends `Person` and `Abstract_Schema_Piece`. I don't want to change the classe structure.

Also, this piece of the code was written by one of the founders of Yoast, so I prefer leaving it as it is :)

php/integrations/yoast.php
leogermani: Cleaned up some namespacing of global functions in https://github.com/Automattic/Co-Authors-Plus/pull/877/commits/ff9db8206c398d7421362ff6c7e1a86e21e155d5

php/integrations/yoast.php
dkoo: Sounds good!

php/integrations/yoast.php
dkoo: Nice, I see that CAP does not use the global namespace, so this looks good to me.

dkoo (approved):

Issue #875, opened 2022-10-22
CLI Command: create_guest_author


oscarssanchez:
Hello,

With a client who didn't have their data in appropriate CSV format we could've used a command that creates a new user with CLI. In the end, we formatted the data but I see people being in the same scenario. It would be helpful if they could just loop through a text file while reading lines, assign values to variables and repeat the same command. Also it would be helpful for people who just want to create a guest author from CLI.

I'd be happy to craft a PR if this is something you would consider. Thanks!

rebeccahum:
@oscarssanchez Yes, a PR is always welcome!

oscarssanchez:
Thanks @rebeccahum I opened https://github.com/Automattic/Co-Authors-Plus/pull/879 in the meantime, solving that problem first would be great before doing this one!

PR #879, opened 2022-10-28
Handle WP_Error on failed guest author creation for method create_guest_author()

### LINKED ISSUE #878: BUG: create_guest_author() private method does not display anything when WP_Error is returned
It seems the intended behavior of: https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-wp-cli.php#L886 is failing as of now. If I try these steps I'm not able to see what was the problem with the CSV, or if there was a problem at all:

1.- Use this sample file 
[test.csv](https://github.com/Automattic/Co-Authors-Plus/files/9883818/test.csv)
2.- Run command: wp co-authors-plus create-guest-authors-from-csv --file=test.csv
3.- See the following output: 

Screen Shot 2022-10-27 at 19 20 13

Since `create()` only returns `WP_Error` or the guest author id, I think we can instead check if the returned value is a `WP_Error`

oscarssanchez:
Hi there,

This PR checks if `create()` returns a WP_Error and displays a warning with the error message https://github.com/Automattic/Co-Authors-Plus/compare/master...oscarssanchez:fix/error-handling-create-guest-author-cli?expand=1#diff-a52979a3ff3a75e65a5166f5fc4a259ef7e243368b5953e587054bbfbada9865R887

Fixes #878 

It also changes a couple things I think could make the command more readable (both in code and output in the terminal)

* Changed output from a single line to display `success` when creating an author, or `warning` when there's an issue. This could help readability in the terminal in my opinion
* Adds i18n
* Adds documentation to create_guest_author() and create()
* A bit of refactoring to reduce if statements

This could also help with development for #875 

Thanks for your feedback!

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -889,14 +889,14 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 			return WP_CLI::warning( sprintf( esc_html__( '-- Failed to create guest author: %s', 'co-authors-plus' ), $guest_author_id->get_error_message() ) );
 		}
 
-		// translators: Guest Author ID.
-		WP_CLI::success( sprintf( esc_html__( '-- Created as guest author #%s', 'co-authors-plus' ), $guest_author_id ) );
-
 		if ( isset( $author['author_id'] ) ) {
 			update_post_meta( $guest_author_id, '_original_author_id', $author['ID'] );
 		}
 
 		update_post_meta( $guest_author_id, '_original_author_login', $author['user_login'] );
+
+		// translators: Guest Author ID.
+		WP_CLI::success( sprintf( esc_html__( '-- Created as guest author #%s', 'co-authors-plus' ), $guest_author_id ) );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


php/class-wp-cli.php
```diff
@@ -858,36 +864,39 @@ private function create_guest_author( $author ) {
...
+		}
 
-				update_post_meta( $guest_author_id, '_original_author_login', $author['user_login'] );
-			} else {
-				WP_CLI::warning( '-- Failed to create guest author.' );
-			}
-		} else {
-			WP_CLI::line( sprintf( '-- Author already exists (ID #%s); skipping.', $guest_author->ID ) );
+		// translators: Guest Author ID.
+		WP_CLI::success( sprintf( esc_html__( '-- Created as guest author #%s', 'co-authors-plus' ), $guest_author_id ) );
```
leogermani, line 893: Nitpick, but I think it's better if we have the success output message as the last thing, after the update_post_meta calls, so we are sure we only output a success after everything was done

oscarssanchez: That's a good point @leogermani I updated the code here: https://github.com/Automattic/Co-Authors-Plus/pull/879/commits/105d38b92079662ed5fce28d744c616dc345a7d0#diff-a52979a3ff3a75e65a5166f5fc4a259ef7e243368b5953e587054bbfbada9865R899

leogermani (approved):

PR #880, opened 2022-11-02
Add command documentation

oscarssanchez:
This PR implements #875 .

This is just a wrapper for `create_guest_author()` we can use in the terminal. It should handle all of the scenarios correctly now that we return back the corresponding `WP_Error` telling the user what is missing from the required fields. Given `create_guest_author()` is used in other commands I didn't make any other changes or method rename. 

Happy to discuss anything if needed!

Implements #875

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -770,6 +770,25 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	}
 
 	/**
+	 * Create a single guest author.
+	 *
+	 * self::create_guest_author() wrapper.
+	 *
+	 * @subcommand create-single-author
+	 * @synopsis
+	 * [--display_name=<display_name>]
+	 * [--user_login=<user_login>]
+	 * [--first_name=<first_name>]
+	 * [--last_name=<last_name>]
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -769,6 +769,25 @@ public function create_guest_authors_from_wxr( $args, $assoc_args ) {
 		WP_CLI::line( 'All done!' );
 	}
 
+	/**
+	 * Create a single guest author.
+	 *
+	 * self::create_guest_author() wrapper.
+	 *
+	 * @subcommand create-single-author
```
rebeccahum, line 777: Can we do `create-author`?

rebeccahum (changes requested):

php/class-wp-cli.php
oscarssanchez: Thanks for reviewing @rebeccahum , for sure I just updated my commit here https://github.com/Automattic/Co-Authors-Plus/pull/880/commits/c9d26853a3f19363c90c324fe6210b9be6b5350a

rebeccahum (approved):

PR #883, opened 2022-11-03
Change permission callback for coauthors/v1/authors and coauthors/v1/search endpoint

rebeccahum:
This PR does two things:
- Simplifies `current_user_can_set_authors()` to user caps only. It shouldn't matter what the `post_type` is...especially when the endpoint at the end of the day falls back to user caps anyways, so we can save on processing for that
- Changes the permission callback for `coauthors/v1/authors` and `coauthors/v1/search` endpoint

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1134,7 +1134,16 @@ class CoAuthors_Plus {
 	 * Checks to see if the current user can set co-authors or not
 	 */
 	function current_user_can_set_authors() {
-		$can_set_authors = current_user_can( 'edit_others_posts' );
+		$current_user = wp_get_current_user();
+		if ( ! $current_user ) {
+			return false;
+		}
+		// Super admins can do anything
+		if ( function_exists( 'is_super_admin' ) && is_super_admin() ) {
+			return true;
+		}
+
+		$can_set_authors = isset( $current_user->allcaps['edit_others_posts'] ) ? $current_user->allcaps['edit_others_posts'] : false;
 
 		return apply_filters( 'coauthors_plus_edit_authors', $can_set_authors );
 	}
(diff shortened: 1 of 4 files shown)


leogermani (approved):
The change looks good to me. Both the simplification and the change on the permission_callback.

The code you are removing is most certainly some left over... All that code to get the `$post_type_object` and that's not used anywhere else! Just the current user...

Maybe in the past they were trying to have different set of capabilitites for each post type. This is something we discussed in another PR. The plugin is not ready to have a different set of capabilities for each post type - it will always use the `post` capability_type.

If you wanted to simplify this even further, it looks to me that we could simply replace

```PHP
$current_user = wp_get_current_user();
if ( ! $current_user ) {
	return false;
}
// Super admins can do anything
if ( function_exists( 'is_super_admin' ) && is_super_admin() ) {
	return true;
}

$can_set_authors = isset( $current_user->allcaps['edit_others_posts'] ) ? $current_user->allcaps['edit_others_posts'] : false;
```

with

```PHP
$can_set_authors = current_user_can( 'edit_others_posts' );
```

leogermani (approved):

mjangda:
> If you wanted to simplify this even further

@leogermani we had to revert this because of some interesting hacks we have to allow lower-privileged users (e.g. author) that are added as coauthors to edit those posts. Some of this legacy code could probably be removed since the original core issues forcing those hacks have been addressed.

mjangda (approved):
Tested and verified that editing posts continues to work as expected for different roles and the REST API endpoints correctly block access if you don't have the correct capabilities.

One minor note is that we could allow `GET /authors/:id` access for users who are coauthors of the requesting post. We may revisit that in the future.

leogermani:
@mjangda ok! Sorry for adding this extra step to the process.. I should have guessed there was a reason for that

PR #887, opened 2022-11-07
Fix failing tests on 6.1

rebeccahum:
Looks like ` decoding="async"` gets added to `get_avatar()` as a default `$args` now: https://github.com/WordPress/WordPress/blob/65cdec186bc58b25258625f9db8e4a1c90aeb985/wp-includes/pluggable.php#L2794. We need to adjust the regex to pick that up.


leogermani (approved):
Fix confirmed.

Also confirmed that the test still pass in older versions of WP

PR #890, opened 2022-11-16
CI: Release to WordPress.org on release

rebeccahum:
Secrets have been already added as well.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,5 +16,6 @@ jobs:
       - name: Push to WordPress.org
         uses: 10up/action-wordpress-plugin-deploy@stable
         env:
+          SLUG: co-authors-plus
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: New release to WordPress.org
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Build # Remove or modify this step as needed
+        run: composer install --no-dev
+      - name: Push to WordPress.org
+        uses: 10up/action-wordpress-plugin-deploy@stable
+        env:
```
GaryJones, line 18: You'll need to add the "slug" in this config, or rename the repo to use lowercase letters - the deploy action is case-sensitive.

GaryJones (changes requested):

PR #894, opened 2022-11-16
Add .distignore file

rebeccahum:
Specify which files should not be included in the push to WordPress.org. These are all development files and directories.

Note that `/vendor` is not included here - that means that it WILL be available in the plugin on WordPress.org, which is essential as we have some `require` dependencies.

The deploy Action will use rsync + `.distignore` if the `.distignore` exists, so it doesn't care what may or may not be ignored via `.gitignore`.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -1,0 +1,9 @@
+/.distignore
+/.editorconfig
+/.git
+/.gitignore
+/.github
+/bin
+/package.json
+/composer.json
+/tests


GaryJones (changes requested):
How about the Makefile, package-lock.json, phpcs and PHPUnit XML files, etc.

Go through the root of the repo and add files that only exist for developers, and aren't needed in the plugin zip.

GaryJones (commented):
The PHPCS config file is still missing.

GaryJones (approved):

PR #898, opened 2022-11-23
remove composer autoloader dependency

### LINKED ISSUE #897: Installing CAP via git will fatal error without a composer install (breaking change in 3.5.5)
3.5.5 carried a breaking change not included in the release notes, so I suspect that this was unintentional. The main plugin file now [requires a composer autoloader](https://github.com/Automattic/Co-Authors-Plus/blob/bc70909c8e22db21b9724494192719df8cb002b3/co-authors-plus.php#L37). Prior to 3.5.5, a `composer install` was only required for development on the plugin; now it is a prerequisite to run the plugin. This is not documented in the README's installation steps. Any site including CAP as a submodule or subtree would start to fatal if updated.

Since the autoloader is only being included for one class, which is then immediately called, it serves no benefit (in fact, it's instead ever-so-slightly _less_ performant). Given the cost:benefit ratio in this instance, I would recommend removing the autoloader in favor of requiring the class directly.

A site installing CAP via the dot-org plugin repo would be unaffected, since the deployment to the dot-org repo runs a `composer install`.

leogermani:
## Description

Removes the composer autoloader dependency

Fixes #897

## Deploy Notes

We are removing composer autoloader dependecy, so we no longer need the vendor folder to be shipped.

## Steps to Test

1. Activate the plugin and make sure you see no errors or warnings
2. Test it with and without Yoast plugin activated alongside it

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -34,16 +34,13 @@ Author - user with the role of author
 
 define( 'COAUTHORS_PLUS_VERSION', '3.5.6' );
 
-require_once dirname( __FILE__ ) . '/vendor/autoload.php';
-
 require_once dirname( __FILE__ ) . '/template-tags.php';
 require_once dirname( __FILE__ ) . '/deprecated.php';
 
 require_once dirname( __FILE__ ) . '/php/class-coauthors-template-filters.php';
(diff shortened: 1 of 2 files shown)


mboynes (approved):
:+1: Thanks!

rebeccahum (approved):

PR #909, opened 2022-12-27
fix: Make CoAuthors_Plus::get_coauthor_by() compatible with unicode author usernames

malhusseiny:
Currently CoAuthors_Plus::get_coauthor_by() fails to retrieve guest authors whose "user_login" names contains non-ascii characters. Those user_logins are effectively slugs so they are stored in the database url-encoded, and thus should be compared against a url-encoded version of the user_login if they are to be found.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -929,7 +929,7 @@ class CoAuthors_Guest_Authors {
 				}
 				// Ensure we aren't doing the lookup by the prefixed value
 				if ( 'user_login' == $key ) {
-					$value = preg_replace( '#^cap\-#', '', $value );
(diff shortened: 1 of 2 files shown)


php/class-coauthors-guest-authors.php
```diff
@@ -929,7 +929,7 @@ function get_guest_author_by( $key, $value, $force = false ) {
 				}
 				// Ensure we aren't doing the lookup by the prefixed value
 				if ( 'user_login' == $key ) {
-					$value = preg_replace( '#^cap\-#', '', $value );
+					$value = preg_replace( '#^cap\-#', '', sanitize_title( $value ) );
```
rebeccahum, line 932: Do you think `sanitize_title_for_query()` would be a better function to use since it gets passed into `$query` later?

rebeccahum (approved):

PR #870, opened 2022-09-14
Refactors Gutenberg Feature

jomurgel:
Closes: #868

- Refactors the Gutenberg plugin to use modern WP JavaScript standards.
- Removes use of legacy functions `withState`, `withDispatch`, `withSelect`.
- Restructures src directory and breaks out and documents components.
- Improves fetch performance.
- Updates and includes necessary dependencies.
  - Updates the `@wordpress/scripts` package and updates the `package.json` script for formatting.

NOTE: The giant diff is a result of the `package-lock.json` file.

PR #914, opened 2023-01-17
only update authors on post save

leogermani:
## Description

Based on https://github.com/eupolitico/Co-Authors-Plus/commit/230b1f354b22f41a771e4751c1a507fcd1e251c9

This fixes a regression added in https://github.com/Automattic/Co-Authors-Plus/pull/870 that would immediately update the authors of a post once you selected them in the sidebar.

With this PR, authors will only be updated when saving the post.

Note: This does not fix a bug that was present that will trigger the author update also for auto-saves.

## Deploy Notes

Nothing to note.

## Steps to Test

1. Edit an existing post
2. Make changes to the author list (add or remove an author)
3. Confirm that the change is not applied until you save the post

cc @elliott-stocks @jomurgel


rebeccahum (approved):
Tested and works great!

PR #920, opened 2023-01-23
fix subscribing triggering multiple saves

leogermani:
## Description

After the last review, several sites are experiencing issues when saving posts.

They describe that when they remove an author from the list and save, the original author is added back to the post.

I noticed that the `subscribe` callback is triggering multiple ajax requests to save the post. I fixed it by putting the `subscribe` call outside the component, and thus running it only once. This is how things were done before. It seems to solve the issue.

Context: In https://github.com/Automattic/Co-Authors-Plus/pull/870 the `subscribe` was removed and the changes were saved instantly. https://github.com/Automattic/Co-Authors-Plus/pull/914 reintroduced the `subscribe` behavior.

There are other issues though, that were not introduced by the latest changes, which this PR does not fix.
* sometimes, when you change authors and save, the old authors are still present in cache and will only be refreshed after a cache flush
* sometimes the update authors ajax request is not triggered at all

These 2 bugs are not new and should be addressed in other PRs

## Deploy Notes

None

## Steps to Test

On `master`

1. Edit a post with multiple authors
2. Remove an author from the list and Update the post
3. Inspect the Network tab and notice there are at least two requests to `coauthors/v1/authors/{$POST_ID}?new_authors={$AUTHORS_NICENAMES}&_locale=user`

Apply this branch and do the same.

1. Verify that every time you save the post only one request is made with the correct list of authors.
2. Try to edit the authors in a way only one guest author is the author
3. Try to edit the authors in a way only one WP user is the author


src/index.js
```diff
@@ -4,11 +4,12 @@
 import { registerPlugin } from '@wordpress/plugins';
 import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
 import { __ } from '@wordpress/i18n';
+import { select, subscribe } from "@wordpress/data";
 
 /**
  * Components
  */
-import CoAuthors from './components/co-authors';
+import CoAuthors from "./components/co-authors";
```
dkoo, line 12: Nit: why the `"` double quotes? I see single quotes elsewhere in the code.

dkoo (dismissed):
Moving the `subscribe` outside of the `CoAuthors` component fixes the double-update issue for me. I notice some very minor, non-blocking code style changes which are probably due to IDE automations, but wanted to point out in case we want to keep the code style consistent.

src/index.js
leogermani: This was prettify working on its own... I'll manually revert those in another editor

Issue #905, opened 2022-12-15
CoAuthors Plus & Elementor Archive Pages conflict.


mattpramschufer:
We have a sidebar global template in Elementor, it has a Posts widget that has the Author meta data shown. This works great on ALL pages EXCEPT author archive templates. When viewing the author archive the post widget takes the author name of the author archive rather than then post author on the widget block.

https://s3.amazonaws.com/pramadillo/co-authors1.jpg
https://s3.amazonaws.com/pramadillo/coauthors-2.jpg
https://s3.amazonaws.com/pramadillo/co-authors3.jpg

I have created a support ticket with Elementor and was escalated to their tier 3 support which then indicated that the error resides in the CoAuthors Plus plugin and not within Elementor.  Their response was

> On Wed, Dec 14, 2022 at 1:46 PM [help@support.elementor.com](mailto:help@support.elementor.com) wrote:
> Hello there,
> 
> I appreciate your patience. I have received feedback from Tier 3.
> 
> They said that it is a plugin incompatibility with [Co-Authors Plus](https://wordpress.org/plugins/co-authors-plus/). We created an incompatibility report, but I recommend you report this issue to the [Co-Authors Plus](https://wordpress.org/plugins/co-authors-plus/) support as well. So they can inspect this further.
> 
> If you have any more questions, feel free to email me back.
> 
> Kind regards,

mattpramschufer:
Any insight on this at all?

mattpramschufer:
This is a direct result of the `public function fix_author_page( $selection ) {`

You are making the author of every post on the author archive page be whatever the query_var author_name equals.  Which that is always going to be whichever author of the archive you are looking at.  I have submitted PR #926 which corrects the issue.

GaryJones:
Closed in https://github.com/Automattic/Co-Authors-Plus/pull/926.

PR #926, opened 2023-02-07
Fixes CoAuthors Plus & Elementor Archive Pages conflict. #905

mattpramschufer:
## Description

Fixes issue where `fix_author_page()` makes every post show on the author archive page be the same author as the archive author.  This is an issue when you have a sidebar that lists out posts from site that do not belong to the archive author.

## Deploy Notes

## Steps to Test

Outline the steps to test and verify the PR here.

Example:

1. Check out PR.
1. Do stuff

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -1169,7 +1169,7 @@ class CoAuthors_Plus {
 	 */
 	public function fix_author_page( $selection ) {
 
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
Makes sense. Tests pass - thanks!

GaryJones:
Reverted with #970.

Issue #937, opened 2023-06-15
Issue with apostrophes in names


thedanielcho:
Hello, our site was having an issue with the co-authors plugin wherein if a user had a name with an apostrophe in it, on the edit-post page, it would display the apostrophe in the name as &$:! (ex. Jonathan O'Reily turned into Jonathan O&$:!Reily). For the time being we have resolved this issue by modifying the plugin on line 214 of class-coauthors-endpoint.php and switched the esc_html function call to strip_tags which worked on our end, but we would appreciate support from your end if this solution isn't ideal. Thank you.

PR #938, opened 2023-06-15
Change esc_html to strip_tags for displayName in _format_author_data function in class-coauthors-endpoint.php

thedanielcho:
## Description

This is a fix for https://github.com/Automattic/Co-Authors-Plus/issues/937#issue-1759165795 where I switched out the esc_html function call for strip_tags for the displayName to fix an apostrophe display issue on the post-edit page.

Should include any special considerations, decisions, and links to relevant GitHub issues.

Please don't include internal or private links

## Deploy Notes

## Steps to Test

Outline the steps to test and verify the PR here.

Example:

1. Check out PR.
1. Do stuff

### DIFF
--- a/php/class-coauthors-endpoint.php
+++ b/php/class-coauthors-endpoint.php
@@ -211,7 +211,7 @@ class Endpoints {
 			'userNicename' => esc_html( rawurldecode( $author->user_nicename ) ),
 			'login'        => esc_html( $author->user_login ),
 			'email'        => sanitize_email( $author->user_email ),
-			'displayName'  => esc_html( str_replace( '∣', '|', $author->display_name ) ),
+			'displayName'  => strip_tags( str_replace( '∣', '|', $author->display_name ) ),
 			'avatar'       => esc_url( get_avatar_url( $author->ID ) ),
 		);
 	}


GaryJones:
Thanks, Daniel. 

I'm not convinced that the escaping should be done before the REST response is sent back since there's no idea of what context that escaping should be for - will it be for HTML, an attribute, XML, or something else? Needs an investigation.

GaryJones:
Thanks for raising this and apologies for the very long delay, Daniel.

Your diagnosis is right: `esc_html` in the REST response is what's breaking the apostrophe display (the entity-encoded value reaches the editor and is rendered literally). However, swapping to `strip_tags` isn't the correct fix:

- `strip_tags` doesn't escape anything; it just removes tag pairs. It only "fixes" the symptom because it leaves the apostrophe alone.
- The underlying issue is that escaping is being done in the wrong place. A REST endpoint shouldn't escape for a specific output context (HTML body, attribute, JS string, XML, etc.) since it can't know how the consumer will render the value. That's the consumer's responsibility.
- The same flaw applies to the other fields in `_format_author_data` (`userNicename`, `login`, `userType`), so the proper fix is broader than the single line here.

Closing this PR as the right fix needs to address the whole method and verify the JS-side rendering escapes for HTML context. Issue #937 will remain open to track the underlying bug.

Thanks again for spotting it.

PR #939, opened 2023-06-20
Ensure 'nickname' is created to prevent is_author() PHP notice with Guest Author

srtfisher:
Ensure `nickname` is created on the object to prevent an issue with `nickname` not existing on the Guest Author's object: https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L4087

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -973,7 +973,7 @@ class CoAuthors_Guest_Authors {
 		$guest_author['type']          = 'guest-author';
 
 		if ( ! isset( $guest_author['nickname'] ) ) {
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks Sean!

Would it make more sense just to set it as an empty string anyway, regardless of the population of the display name?

srtfisher:
@GaryJones sounds good to me, made that change

GaryJones:
Thanks @srtfisher.

In what context (admin page? another request?, etc.) were you seeing the PHP Warning? (so I can reproduce it before applying the patch).

srtfisher:
I'm seeing this crop up a lot when using `is_author( 'name' )` on a guest author feed. the name argument is required to get the notice.

GaryJones (approved):
✅

PR #941, opened 2023-07-04
fix: only shows authors in the Link user dropdown

leogermani:
## Description

This is a PR / question.

We have had issues with the "Link user" dropdown, under the "Unique slug" metabox when editing a Guest Author.

When a site has thousands of subscribers, the dropdown gets too big and it causes the browser to crash.

We have worked around this issue via the provided filter, but I was wondering if this makes sense to be the default behavior, as chances are that we will want to link a Guest Author to a WP author, and not a subscriber or a user without rights to write a post.

Related/similiar PR -> https://github.com/Automattic/Co-Authors-Plus/pull/872

## Deploy Notes

No

## Steps to Test

* Create a site with thousands of subscribers (like 200,000).
* Edit a Guest Author
* You should see the dropdown causing the page to halt and maybe the browser to crash

Checkout this PR and see that the page works fine and only users with edit_posts cap are displayed.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -653,6 +653,7 @@ class CoAuthors_Guest_Authors {
 					'selected'         => $linked_account_id,
 					// Don't let user accounts to be linked to more than one guest author
 					'exclude'          => $linked_account_user_ids,
+					'capability'       => array( apply_filters( 'coauthors_edit_author_cap', 'edit_posts' ) ),
 				)
 			)
 		);
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -653,6 +653,11 @@ class CoAuthors_Guest_Authors {
 					'selected'         => $linked_account_id,
 					// Don't let user accounts to be linked to more than one guest author
 					'exclude'          => $linked_account_user_ids,
+					// Restrict candidates to users who can write posts. The
+					// coauthors_edit_author_cap filter mirrors the same check used
+					// by the AJAX co-author search and validation paths, so a site
+					// that already overrides who can be a co-author keeps its
+					// linked-account dropdown consistent.
 					'capability'       => array( apply_filters( 'coauthors_edit_author_cap', 'edit_posts' ) ),
 				)
 			)
--- a/tests/Integration/GuestAuthorLinkedAccountDropdownTest.php
+++ b/tests/Integration/GuestAuthorLinkedAccountDropdownTest.php
@@ -1,0 +1,148 @@
+<?php
+
+namespace Automattic\CoAuthorsPlus\Tests\Integration;
+
+/**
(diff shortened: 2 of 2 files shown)


GaryJones:
Not fully checked, but if the code does what I think it does, that looks sensible. This would be a bit of a breaking change (in behaviour), though, so I would suggest saving it for a minor (not point) release.

GaryJones:
> as chances are that we will want to link a Guest Author to a WP author, and not a subscriber or a user without rights to write a post

I guess there could be a case when creating a Guest Author linked to a Subscriber is exactly the use case for someone - that they don't want to give a Subscriber any higher role, but do want to be able to mark them as a co-author on a post.

GaryJones:
Thanks @leogermani for the original patch and for your patience while this sat open. Merged.

For posterity, a quick summary of what landed and the reasoning that got us there:

**What changed on the branch before merge.** The branch was 713 commits behind `develop`, so I rebuilt it: reset to current `develop`, cherry-picked your commit (`8184245`, authorship preserved), then added a follow-up test commit on top with an inline comment in the metabox explaining the rationale. The PR diff against `develop` ended up as one capability line plus a comment block in `php/class-coauthors-guest-authors.php`, plus a new `tests/Integration/GuestAuthorLinkedAccountDropdownTest.php`.

**Why we treated this as a fix, not a breaking behaviour change.** My initial 2023 reaction was that this should wait for a minor release because Subscribers might legitimately be linked as guest authors. Re-reading the code three years on, that concern doesn't really hold up. The `coauthors_edit_author_cap` filter is already used in two places in CAP:

- the AJAX co-author search (`class-coauthors-plus.php`) only returns users with `edit_posts` by default, and
- the co-author validation path uses the same cap check before accepting a found user.

So a site that wants to allow Subscribers to be co-authors already has to filter `coauthors_edit_author_cap` for the rest of CAP to work for them. This PR brings the Guest Author "Link user" dropdown into alignment with that existing convention rather than introducing a new restriction. Sites with a custom `coauthors_edit_author_cap` filter keep their existing behaviour for free.

**What the tests cover.** Three integration tests in `GuestAuthorLinkedAccountDropdownTest`:

1. The args passed through `coauthors_guest_author_linked_account_args` include `'capability' => ['edit_posts']` by default.
2. Overriding `coauthors_edit_author_cap` to return a different capability flows through to the dropdown args.
3. The rendered dropdown HTML excludes Subscriber accounts and includes Author accounts.

I verified all three pass with the change and fail without it, so they genuinely cover the behaviour rather than just the wiring.

**One thing left for a follow-up, not blocking this.** A `<select>` is the wrong control at the scale that prompted the original report (200,000 subscribers). The right long-term fix is to swap `wp_dropdown_users` here for an autocomplete, mirroring what CAP already does for the co-author search. This PR removes the immediate browser-crash failure mode; a future change can take the next step.

Issue #854, opened 2022-06-07
Post featured image size is being used for avatars


manudavidos:
I have a theme that crops featured images. Because this plugin uses the featured image feature for avatars, all the guest avatars are also cropped according to my featured image sizing.

manudavidos:
`/**
 * Post Thumbnails.
 */
if ( function_exists( 'add_theme_support' ) ) { 
    add_theme_support( 'post-thumbnails' );
    set_post_thumbnail_size( 400, 200, true ); // default Post Thumbnail dimensions (cropped)
    add_image_size( 'enjoyblog_featured_thumb', 880, 440, true );    
    add_image_size( 'enjoyblog_post_thumb', 480, 270, true );
}`

Here is part of the code from functions.php that is cropping thumbnails

DustinHartzler:
Thanks for including this code.

Can you also share with us the code to display the guest avatars in your theme? This will help us understand what's going on.

PR #959, opened 2023-07-27
Issue#958/ unable to change avatar images sizes

### LINKED ISSUE #958: Unable to change avatar image sizes
While working on #940 ( specifically while trying to add coauthor avatars to the block editor ), I noticed that `rest_get_avatar_urls()` returned the same URL for all requested sizes.

`rest_get_avatar_urls()` uses `get_avatar_url()` as shown here https://developer.wordpress.org/reference/functions/rest_get_avatar_urls/

- For a user, the requested size is ignored and the default 96px image is always returned.
- For a guest-author with a featured image, the requested size is ignored and the full resolution featured image is always returned.

If I deactivate CoAuthors Plus, obviously there's no guest authors any more, but the user avatars start producing the correct size.

I believe the issues stem from `filter_pre_get_avatar_data_url`

- It calls `get_the_post_thumbnail_url()` and passes a single integer for the size, when an array of integers is expected. So it is treated a request for the full resolution image.
  - https://github.com/Automattic/Co-Authors-Plus/blob/3.5.11/co-authors-plus.php#L1837
- When making its recursive calls to `get_avatar_url()` it does not pass the `$args` which changes the args back to default.
  - https://github.com/Automattic/Co-Authors-Plus/blob/3.5.11/co-authors-plus.php#L1839

Other uses of `get_avatar()` or `get_avatar_url()` in CoAuthors Plus that only return the default size for users, and the full resolution feature image for guest authors:

- https://github.com/Automattic/Co-Authors-Plus/blob/3.5.11/co-authors-plus.php#L452C71-L452C71
- https://github.com/Automattic/Co-Authors-Plus/blob/3.5.11/php/class-coauthors-endpoint.php#L216

douglas-johnson:
## Description

Resolves #958 
#854 may also be resolved by this

- Updated `filter_pre_get_avatar_data_url` so it passes the requested width and height arguments to `get_the_post_thumbnail_url`
  - Ensures the requested avatar size ( or the closest your particular combination of theme and hosting environment can produce ) is applied to a guest author's feature image.
- Updated `filter_pre_get_avatar_data_url` so it passes the same arguments it received when recursively calling `get_avatar_url`
  - Ensures the requested avatar size is applied to Gravatar images rather than reverting to the default.

## Steps to Test

- Add the Post Author block to a post or template.
  - Any changes to the Avatar Size should be reflected in the rendered content.
- Inspect the response to a request for `/wp-json/wp/v2/users/:user-id/` from the block editor.
  - The URLs in the `avatar_urls` object should match their associated keys. For example the default sizes: 24, 48, 96
- Inspect the response to a request for `/wp-json/coauthors/v1/authors/:post-id/` from the block editor.
  - For authors using gravatar, the avatar should use the default size of 96px.
  - For guest authors using a feature image, the value should use the closest available size to 96px.
- Inspect the classic editor coauthor meta box.
  - For authors using gravatar, the avatar should use the default size of 96px.
  - For guest authors using a feature image, the value should use the closest available size to 96px. 
- Use your browser's dev tools to turn off JavaScript and open a post in the classic editor.
  - The coauthor avatars in `#coauthors-readonly` should be constrained to 25px through CSS.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -56,8 +56,6 @@ class CoAuthors_Plus {
 	var $coauthors_meta_box_name   = 'coauthorsdiv';
 	var $force_guest_authors       = false;
 
-	var $gravatar_size = 25;
-
 	var $_pages_whitelist = array( 'post.php', 'post-new.php', 'edit.php' );
 
 	var $supported_post_types = array();
@@ -449,7 +447,7 @@ class CoAuthors_Plus {
 					$avatar_url = get_avatar_url( $coauthor->ID );
 					?>
 					<li>
-						<?php echo get_avatar( $coauthor->ID, $this->gravatar_size ); ?>
+						<?php echo get_avatar( $coauthor->ID ); ?>
 						<span id="<?php echo esc_attr( 'coauthor-readonly-' . $count ); ?>" class="coauthor-tag">
 							<input type="text" name="coauthorsinput[]" readonly="readonly" value="<?php echo esc_attr( $coauthor->display_name ); ?>" />
 							<input type="text" name="coauthors[]" value="<?php echo esc_attr( $coauthor->user_login ); ?>" />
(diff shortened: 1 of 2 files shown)


douglas-johnson:
@GaryJones I updated this pull request to remove `var $gravatar_size` as discussed in #958 . The desciption and testing steps have been updated accordingly.

GaryJones (approved):

Issue #907, opened 2022-12-22
Matching IDs can cause WP Users to pick up Guest Author avatars in WP Admin


laurelfulford:
When a WP User and Guest Author share the same ID, the WP User can pick up the wrong avatar in the backend.

## Steps to reproduce:

1. Start with a test site that has a Guest Author and WP User that share the same ID. If you don't already have one, the quickest way to recreate is:
    * Set up a new site. 
    * Before adding/importing any posts, install CoAuthors Plus and create a Guest Author. Usually they will have an ID around 5-6. This is necessary because Guest Authors share the same ID pool as other post types.
    * Create 5-6 regular WP Users, until you have one whose ID matches your Guest Author.
2. Give your "ID twin" Guest Author an avatar. 
3. You can also give your "ID twin" WP User an email address that's associated with a Gravatar, but it's not necessary. 
4. View the user list under WP Admin > Users, or view the individual profile of your "ID twin" WP User. Note that the avatar its using is the one assigned to your Guest Author. 

## Screenshots

Guest author settings: 

![image](https://user-images.githubusercontent.com/177561/209243771-3f6c4f97-6021-4977-bd57-4eb70f51a5e0.png)

"ID twin" WP User on the Users screen:

![image](https://user-images.githubusercontent.com/177561/209243830-6b66e896-23b9-461f-b8d3-b475d5f41d9a.png)

Actual avatar the WP User should be using: 

![image](https://user-images.githubusercontent.com/177561/209243863-58d7530c-f199-46a0-bbd0-f98749a72376.png)

davidvexel:
Having this same issue and I think it's because the filter in ```filter_pre_get_avatar_url``` is assuming that the `$id` parameter sent is an Author ID, which is incorrect, I'm seeing the user ID there.

https://github.com/Automattic/Co-Authors-Plus/blob/master/co-authors-plus.php#L1825-L1840

GaryJones:
Fixed in #907.

PR #960, opened 2023-07-28
Fix avatar id collision between user id and guest-author post type

katag9k:
## Description

Fixes filtering issues when a user has the same id as a guest-author post id.

## Steps to Test

Ensure a guest author shares the same id (`post_id`) with a current WordPress user id.

* Login as the WordPress user - the WordPress admin bar will display the proper user avatar instead of the wrong guest author avatar
* Go to `/wp-admin/profile.php` the correct avatar will be displayed under *Profile Picture*
* Go to `/wp-admin/users.php` correct avatars will be displayed for both WordPress users and guest authors
* Create a new article and add both guest authors and WordPress users to the Authors meta box, proper avatars will be displayed

Addresses: #907

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -446,7 +446,11 @@ class CoAuthors_Plus {
 				<?php
 				foreach ( $coauthors as $coauthor ) :
 					$count++;
-					$avatar_url = get_avatar_url( $coauthor->ID );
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
Tested and confirmed - admin bar, user list, guest author list all have the correct images. Thanks!

PR #961, opened 2023-07-29
Add i18n string for PluginDocumentSettingPanel title

antonlukin:
## Description

I've noticed that Setting Panel title was not translated.

## Steps to Test

Open "Edit Post" page with enabled non-English language. 
Screenshot 2023-07-29 at 14 25 20

## Note
I also want to note that for some reason the prepared strings are not displayed in the translate.wordpress.org interface:
https://translate.wordpress.org/projects/wp-plugins/co-authors-plus/

I still can't figure out what the problem is. When I execute the command `wp i18n make-pot . languages/co-authors-plus.pot` new strings appear in this pot file.


GaryJones (approved):
✅

GaryJones:
Thanks!

> When I execute the command wp i18n make-pot . languages/co-authors-plus.pot new strings appear in this pot file.

Thanks for highlighting this. https://github.com/Automattic/Co-Authors-Plus/pull/963 should improve things a lot here.

PR #973, opened 2023-08-01
check get_current_screen exists

katag9k:
The Avada theme (and possibly others) seem to try and retrieve avatars before the `get_current_screen()` method is available - see https://wordpress.org/support/topic/fatal-error-with-avada-theme/ and https://wordpress.org/support/topic/updated-recent-co-authors-plus-plugin-and-caused-error/ - so let's check we can use that function. It isn't easy to test without access to Avada itself.


GaryJones (approved):
✅

PR #976, opened 2023-08-02
975-verification with isset before using properties

cobianzo:
## Description

This is just one line of code to ensure that we are not trying to access to an non existing property in an instance of a coauthor.

## Deploy Notes

No dependencies.

## Steps to Test

Basic fix, just a regular use of the plugin to ensure it goes ok.


GaryJones (approved):
I can't see the fix causing any issues, though it would be good to understand why `$coauthor` isn't an object with the right properties.

PR #982, opened 2023-08-23
Check if $term->term_id is set before using it to delete the term

georgejipa:
## Description

Make sure to check if `$term->term_id` is set before using it to delete the term. This will fix the following PHP notice:

```
PHP Notice:  Trying to get property 'term_id' of non-object in [redacted]/wp-content/plugins/co-authors-plus/co-authors-plus.php on line 1064
```

## Steps to Test

1. Go to Users page.
2. Try to delete a user who is not an author.


GaryJones:
Closing as this was already fixed in commit 9ecae09 (merged via develop). Thanks for the contribution! The fix now includes a null check before accessing `$term->term_id`.

PR #974, opened 2023-08-02
Fix ajax_suggest user and guest author id collision

katag9k:
## Description

Related to #960. When a WordPress user shares the same id as a guest-author post id, the CAP meta box will display the wrong avatar after searching for user or guest author.

## Steps to Test

Ensure a guest author shares the same id (`post_id`) with a current WordPress user id.

* Login as the WordPress user 
* Create a new article and search for both the guest author and the WordPress user that have colliding ids in the Authors Metabox. Correct avatars will be displayed

PR #996, opened 2023-09-06
Fix avatar collision on profile, post and edit screens

katag9k:
## Description

Fixes filtering issues when a user has the same id as a guest-author post id on the post list screen, profile screen and the post edit takeover pop-up.

Related: #960 #973 #974

## Steps to Test

Create a guest user that has the same `post_id` as another WP `user_id`.

In one browser, log in as the WordPress user and edit a post. In another browser login as the guest user. While logged in as a guest user, go to the post list at `/wp-admin/edit.php` the post will be locked as it's being edited by the WordPress user, however, the incorrect guest user's gravatar will be displayed.

Screenshot 2023-09-06 at 2 35 42 PM

Correct avatar should be:
Screenshot 2023-09-06 at 2 58 47 PM

Click edit on the post that is being edited, the warning popup will also have the incorrect gravatar from the guest user.
Screenshot 2023-09-06 at 2 37 52 PM

Correct avatar should be:

Screenshot 2023-09-06 at 1 22 37 PM

Take over the post editing. Go back to the browser where the Guest User was editing the post and wait for the "Somone else has taken over this post" pop-up. The incorrect avatar will be displayed:

 
Screenshot on 2023-09-15 at 14-10-16

The correct avatar should be:

Screenshot on 2023-09-15 at 14-05-54

Now log in as the WordPress user and go to the profile page at `/wp-admin/profile.php` the incorrect gravatar is displayed.

Screenshot 2023-09-06 at 2 40 37 PM

The correct avatar for the WordPress user: 
Screenshot 2023-09-06 at 2 59 38 PM

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1960,6 +1960,16 @@ class CoAuthors_Plus {
 			return $args;
 		}
 
+		// Do not filter on the post list screen, profile screen, and post takeover pop-up.
+		if ( isset( $current_screen->base ) && ( 'post' === $current_screen->base || 'profile' === $current_screen->base || 'edit' === $current_screen->base ) ) {
+			return $args;
+		}
+
+		// Do not filter the avatar if this is doing a heartbeat request on WP refresh lock.
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing -- Only checking action name, not processing data.
+		if ( wp_doing_ajax() && isset( $_POST['action'] ) && 'heartbeat' === sanitize_key( $_POST['action'] ) ) {
+			return $args;
+		}
 
 		$coauthor = $this->get_coauthor_by( 'id', $id );
 		if ( false !== $coauthor && isset( $coauthor->type ) && 'guest-author' === $coauthor->type ) {


BrookeDot:
@katag9k, @GaryJones  after some more testing it appears that this PR may be missing the condition when the post lock is removed via JS/Ajax. Adding the following appears to resolve it, but the two conditionals could likely be combined. 

Thoughts?

```
// Do not filter the avatar if this is doing a heartbeat request on WP refresh lock
		if ( wp_doing_ajax() && isset( $_POST['action'] ) && $_POST['action'] === 'heartbeat' ) {
			return $args;
		}
```

GaryJones:
@katag9k The merge conflict will need addressing please, as well as considering Brooke's comment.

katag9k:
@BrookeDot Thank you, PR updated with the suggested code. 

@GaryJones We are conflict free.

php/class-coauthors-plus.php
```diff
@@ -1826,6 +1826,15 @@ public function filter_pre_get_avatar_data_url( $args, $id ) {
 			return $args;
 		}
 
+		// do not filter on the post list screen, profile screen and post takeover pop-up
+		if ( isset( $current_screen->base ) && ( 'post' === $current_screen->base || 'profile' === $current_screen->base || 'edit' === $current_screen->base ) ) {
+			return $args;
+		}
+
+		// Do not filter the avatar if this is doing a heartbeat request on WP refresh lock
+		if ( wp_doing_ajax() && isset( $_POST['action'] ) &&  'heartbeat' === $_POST['action'] ) {
```
GaryJones, line 1835: Is there a nonce that can be verified here for the heartbeat?
(PHPCS failure)

php/class-coauthors-plus.php
```diff
@@ -1826,6 +1826,15 @@ public function filter_pre_get_avatar_data_url( $args, $id ) {
 			return $args;
 		}
 
+		// do not filter on the post list screen, profile screen and post takeover pop-up
+		if ( isset( $current_screen->base ) && ( 'post' === $current_screen->base || 'profile' === $current_screen->base || 'edit' === $current_screen->base ) ) {
+			return $args;
+		}
+
+		// Do not filter the avatar if this is doing a heartbeat request on WP refresh lock
```
GaryJones, line 1834:
```suggestion
		// Do not filter the avatar if this is doing a heartbeat request on WP refresh lock.
```

php/class-coauthors-plus.php
```diff
@@ -1826,6 +1826,15 @@ public function filter_pre_get_avatar_data_url( $args, $id ) {
 			return $args;
 		}
 
+		// do not filter on the post list screen, profile screen and post takeover pop-up
```
GaryJones, line 1829:
```suggestion
		// Do not filter on the post list screen, profile screen, and post takeover pop-up.
```

GaryJones (commented):
With this PR, we've now got a whole lot more exceptions to the rule than whatever circumstance the rule was meant to be.

Can you work out where the RIGHT places are that should be filtered, and just target those?

PR #998, opened 2023-09-17
Fix script translations problem

antonlukin:
It seems that js script translations (for blocks) still not working.
I determined that the problem is the absence of a required `wp_set_script_translations` function call.

Checked this edition on the last plugin version and it works.


GaryJones:
Thanks!

Since this 3.5.x branch still supports WP 4.1, then I think it will need the JS language files generating and adding to the `/languages` directory and the function call update to include the path argument. Could you add that in please?

I'm expecting this to be the last (or one of the last) 3.5.x releases and the 3.6.x branch will have a much higher version of minimum WordPress, so all of the local translation files and language path references can then be removed.

antonlukin:
I have no confidence that I did everything right. 

Added the latest translations from existing language packs:
https://translate.wordpress.org/projects/wp-plugins/co-authors-plus/language-packs/

GaryJones:
Regenerate .pot file:

```sh
wp i18n make-pot . languages/co-authors-plus.pot
```

Update .po files from .pot:

```sh
wp i18n update-po languages/co-authors-plus.pot
```

Regenerate `.mo` files from `.po` files (ideally the `.po` files would have been translated first - I did en-GB).

```sh
wp i18n make-mo languages/
```

Regenerate JSON files:

```sh
wp i18n make-json languages/
```

GaryJones:
@antonlukin I cherry-picked work and rejigged it into some new commits in #1007, which unfortunately didn't retain you as the original committer. I'd love if you could test the current `develop` branch and double-check that the JS translations are working for you now.

Thank you!

antonlukin:
I checked, everything works correctly. Thanks a lot!

PR #1013, opened 2023-10-28
Create co-authors-plus-da_DK.po

hhjuhl:
Added Danish translation

### DIFF
--- a/languages/co-authors-plus-da_DK.po
+++ b/languages/co-authors-plus-da_DK.po
@@ -1,0 +1,639 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Plugins - Co-Authors Plus - Stable (latest release)\n"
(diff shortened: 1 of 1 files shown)


GaryJones:
Thank you! I'll merge this for the 3.5.16 release but then the 3.6.0 release will have a higher WP minimum version which means we can use the translations from https://translate.wordpress.org/locale/da/default/wp-plugins/co-authors-plus/ (though they will need to be approved by one of https://make.wordpress.org/polyglots/teams/?locale=da_DK).

hhjuhl:
I gave up trying to get hold of the Danish translation editors. I also tried contacting the Automattic PTE on slack, but to no avail. Some of the suggested strings in the Danish glotpress for this plugin have been awaiting review for over 6 years.

I didn't have access to a real PO editor, so I haven't been able to compile the MO file. Hope that's ok

PR #1034, opened 2024-04-22
Fix PHP version 7.4 requirement

alecgeatches:
## Description

Changes in https://github.com/Automattic/Co-Authors-Plus/pull/1032 increased the required [PHP version to `7.4`](https://github.com/Automattic/Co-Authors-Plus/pull/1032/commits/8d73746c651b065d11907dcbfda627c6f02a2473), but missed a few steps, including:

- The PHP version [listed in the README](https://github.com/Automattic/Co-Authors-Plus/blob/586e803/README.md?plain=1#L6)
- [`composer.json` requirements](https://github.com/Automattic/Co-Authors-Plus/blob/586e803/composer.json#L26)
- Tested PHP versions [in integration tests](https://github.com/Automattic/Co-Authors-Plus/blob/586e803/.github/workflows/integrate.yml#L17)

PR #1038, opened 2024-04-25
Fix missing author column in wp-admin Posts table

### LINKED ISSUE #1033: Authors now missing from the Posts list table after 3.6.0 update
I've been working on an issue related to Co-Authors Plus with the WP List Table at `wp-admin/edit.php?post_status=publish&post_type=post` and I just saw there was an update. I went ahead and updated my local environment and noticed that the Authors column is completely missing. Weirdly, the regular Author column is also missing. I seemingly don't have any way to see the author for a post before I click to edit.

~I checked the code that I had previously been working with and noticed it's entirely missing from the updated plugin. Here's a link to the filters I was working with before, that appear to do at least part of the admin column handling: https://github.com/Automattic/Co-Authors-Plus/blob/33ae3b0c8adf54f0ad193caa5fa13f166fcc4ddc/co-authors-plus.php#L265C1-L268C100~

~It's not just the filters missing. The functions are also gone. Was this intentional? Am I missing something?~

Edit: I figured out that PHPStorm's index had not fully updated so the search function hadn't found the new class yet. That said, I am still missing the Author/Authors column in the Posts table and not sure why.

### LINKED ISSUE #1037: On posts list: "Uncaught ReferenceError: coAuthorsPlus_ajax_suggest_link is not defined"
I'll add more information here as I find any. After the 3.6.0 update, we're seeing this error on the posts list, preventing any other scripts from running:

```
Uncaught ReferenceError: coAuthorsPlus_ajax_suggest_link is not defined
    at HTMLDocument.<anonymous> (co-authors-plus.js?ver=3.6.0:356:30)
    at HTMLDocument.dispatch (jquery.min.js?ver=3.7.1:2:40035)
    at v.handle (jquery.min.js?ver=3.7.1:2:38006)
    at Object.trigger (jquery.min.js?ver=3.7.1:2:70124)
    at Function.ajax (jquery.min.js?ver=3.7.1:2:76037)
    at Function.<anonymous> (jquery-migrate.min.js?ver=3.4.1:2:4771)
    at e.<computed> [as ajax] (jquery-migrate.min.js?ver=3.4.1:2:1582)
    at HTMLTableSectionElement.update (cpt.js?ver=6.5.2:48:68)
    at x.<computed>.<computed>._trigger (core.min.js?ver=1.13.2:116:7687)
    at x.<computed>.<computed>._trigger (sortable.min.js?ver=1.13.2:9:25015)
```

Code in question is this:

```js
// Show loading cursor for autocomplete ajax requests
	jQuery( document ).ajaxSend(function( e, xhr, settings ) {
		if ( settings.url.indexOf( coAuthorsPlus_ajax_suggest_link ) != -1 ) {
			// Including existing authors on the AJAX suggest link
			// allows us to filter them out of the search request
			var existing_authors = jQuery( 'input[name="coauthors[]"]' ).map(function(){return jQuery( this ).val();}).get();
			settings.url = settings.url.split( '&existing_authors' )[0];
			settings.url += '&existing_authors=' + existing_authors.join( ',' );
			show_loading();
		}
	});
```

I can see the variable gets defined here https://github.com/Automattic/Co-Authors-Plus/blob/586e803589a5d3f86de6f04a8391385529b3fbbe/php/class-coauthors-plus.php#L1460 but this code doesn't seem to run for the posts list, at least not in our case.

GaryJones:
## Description

In CAP 3.6.0, there was [a bug](https://github.com/Automattic/Co-Authors-Plus/issues/1033) that resulted in the author's column missing in wp-admin Posts table. When removing the author dropdown from the quick edit box, the old logic removed post type support for the CAP-registered author taxonomy, which has the same name as the regular author feature. This hard-hitting removal was likely due to the "inconvenient truths" documented in #1036.

The new logic has a more targeted approach - it updates the query args for getting the quick-edit authors list population to only try and include a `0`, and this successfully bails the dropdown from being displayed.

Fixes #1033.
Fixed #1037.
Closes #1036.

## Steps to Test

Use `3.6.0` tag and see that the author column is missing. Switch to the PR branch, and see that the author column is present. In both cases, the quick edit does not show the author box.

### 3.6.0
Author column not present (incorrect):

Screenshot 2024-04-25 at 16 06 22

### This PR
Author column present (correct):

Screenshot 2024-04-25 at 16 06 35

Quick Edit view does not include author box (correct):

Screenshot 2024-04-25 at 16 05 50

(For comparison, here's what the view would look like with the author box - we don't want this):

Screenshot 2024-04-25 at 16 05 15

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -250,7 +250,7 @@ class CoAuthors_Plus {
 	 *
 	 * By default, this is the built-in and custom post types that have authors.
 	 *
-	 * @since 3.5.16
+	 * @since 3.6.0
 	 *
 	 * @return array Supported post types.
 	 */
@@ -487,13 +487,23 @@ class CoAuthors_Plus {
 	}
 
 	/**
-	 * Removes the default 'author' dropdown from quick edit
+	 * Removes the default 'author' dropdown from quick edit.
 	 */
 	public function remove_quick_edit_authors_box() {
 		global $pagenow;
 
 		if ( 'edit.php' === $pagenow && $this->is_post_type_enabled() ) {
-			remove_post_type_support( get_post_type(), $this->coauthor_taxonomy );
+			/*
+			 * The author dropdown isn't displayed if wp_dropdown_users( $args ) returns an empty string.
+			 * It will return an empty string if the user query returns an empty array.
+			 * We can force it return an empty array by changing $args to include only the user ID 0 which doesn't exist.
+			 * We can target the $args specific to Quick Edit using the filter quick_edit_dropdown_authors_args.
+			 * See https://github.com/Automattic/Co-Authors-Plus/issues/1033.
+			 */
+			add_filter(
+				'quick_edit_dropdown_authors_args',
+				static fn() => [ 'include' => [ 0 ] ]
+			);
 		}
 	}
 


GaryJones:
Tests failing for PHP 7.1 will be handled via #1034. They can be ignored for this PR.

douglas-johnson:
This is working for me locally with WP 6.5.2

I think this also resolves #1037

justinmaurerdotdev:
Love it. I've checked out this branch and it looks like everything works as expected.

alecgeatches:
This fix has been merged and included in the `3.6.1` release!

Issue #1024, opened 2024-02-18
Does co-authors plus fire its own HTTP request to store coauthors?


scottfennell-toptal:
I have written a plugin that synchs blog posts from one blog to another as they are published/updated/deleted.

It uses php to hook into `save_post`.  I noticed recently that it is no longer catching updates to co-author assignments, because it seems like co-authors sends its own HTTP request to store coauthors, long after `save_post` has come and gone.

This is a nuisance for me because now I have to hook in twice to capture all post changes, roughly doubling my network activity.

This behavior from co-authors is intuitive.  The post should be done saving ... when save_post is done.

Should I be doing something different or thinking about this differently somehow?

oekeur:
@scottfennell-toptal 
We have the same problem where we sync posts to a different database and need to have the data complete.
My PR doesn't solve the underlying problem you observed, but adding a new action at least helps to narrow down where to hook.
We have a patched version like this running for quite some time, which works a charm. 
I hope it also helps you if the PR gets merged!

carosioa:
I am bumping this because it seems like #1053 was closed but the underlying bug has not really been addressed. In that thread, they suggest hooking in to save_post later. But when you are saving a post in Gutenberg, there is no priority of save_post in php that will accurately capture the current state of coauthors in JS. 

From my reading, it looks like the author terms are being saved here, with a separate HTTP request like Scott mentioned, specifically after the post has finished saving: https://github.com/Automattic/Co-Authors-Plus/blob/14725a6f877fecd8853f40bf4cc5e9262f71e580/src/index.js#L45

It would be really nice if we could have the current coauthor state in PHP before the post is finished saving.

oekeur:
Hi @carosioa ,

Yeah, that PR didn't seem like it would get accepted, so I closed it.
That is indeed where the authors are saved.

Our current solution is that we've added _(with an exponential back-off)_ an event listener to the publish button, and we trigger a save of the authors on mousedown. This way we're really sure authors are saved before we start saving the post..

```js
domReady(() => {
	const { getCurrentPost } = select('core/editor');
	const { getAuthors, saveAuthors } = select('cap/authors');
	let publishButtonAttempts = 0;
	const maxPublishButtonAttempts = 25;

	function findAndAttachPublishButtonListener() {
		const publishButton = document.querySelector(
			'.editor-post-publish-button__button'
		);
		if (publishButton) {
			publishButton.addEventListener('mousedown', () => {
[...]

PR #1053, opened 2024-06-14
Add an action to signal processing co-author has finished

oekeur:
## Description

Adds a (partly) solution for #1024. Which we also encounter at our company.
We sync posts to an external database secondary database, so we need to know when all data has been updated. 

Because CAP saves co-authors with CPTs and metadata, the authors are updated only after save_post has finished.
It's counter-intuitive from a user perspective but logical from the architecture perspective.

This PR adds an action to explicitly signal when CAP has finished processing, allowing it to hook into that, for example, when syncing to external systems.

## Deploy Notes
❌

## Steps to Test
- Add a minimal plugin that hooks into the save_post action and var_dump (or use a tool like Xdebug to use step debugging)
- Observe that the save_post hook doesn't have up-to-date data when updating an author
- Check out this PR
- Observe the previous functionality is unchanged
- Observe that a new hook is now present that fires after the co-author has been updated, allowing for a direct hook into that.
<details>
<summary>Minimal reproduction</summary>

``` php

<?php  
  
add_action('save_post', 'savePost', 10, 2 );  
  
function savePost ($post_id, $post) {  
    var_dump($post);  
    return true;}  
  
add_action('coauthors_post_updated', 'coAuthorUpdate', 10, 2 );  
  
function coAuthorUpdate ($post_id, $coauthor_objects) {  
    var_dump($coauthor_objects);  
    return true;}  
  
?>

````
</details>

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1029,6 +1029,7 @@ class CoAuthors_Plus {
 			$wpdb->update( $wpdb->posts, array( 'post_author' => $new_author->ID ), array( 'ID' => $post_id ) );
 			clean_post_cache( $post_id );
 		}
+	        do_action('coauthors_post_updated', $post_id, $coauthor_objects);
 		return true;
 
 	}
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -956,6 +956,8 @@ class CoAuthors_Plus {
 				}
 			}
 		}
+		do_action('cap_finished', $post_id);
+
 	}
 
 	public function has_author_terms( $post_id ): bool {
@@ -1029,7 +1031,6 @@ class CoAuthors_Plus {
 			$wpdb->update( $wpdb->posts, array( 'post_author' => $new_author->ID ), array( 'ID' => $post_id ) );
 			clean_post_cache( $post_id );
 		}
-	        do_action('coauthors_post_updated', $post_id, $coauthor_objects);
 		return true;
 
 	}

### CODE REVIEW

**Comment by oekeur** [comment]
I am not sure who I should tag but @GaryJones, what would you think? :)

**Comment by GaryJones** [comment]
@laurelfulford @leogermani @alecgeatches

**Comment by leogermani** [comment]
This can easily be solved by adding your hook to a later priority:

```PHP
add_action('save_post', 'savePost', 20, 2 );  
```

I don't see a strong reason for a hook.

But I'm also not strongly against it. But in that case, the hook should live in the `coauthors_update_post` method, and not on the `add_coatuhros` method. `add_coauthors` is a generic command to add authors that can be used in many other contexts, like CLI for example.  `coauthors_update_post` is the method that is hooked to `save_post` and what would make sense to have a `finished` action to address the issue you describe

oekeur:
I'm sorry for not getting back to you sooner.

Well, adding at a lower priority is always an option. Still, you're then never really sure if cap has run already. Adding a specific hook would allow you to hook into the action itself.

Moving to `coauthors_update_post` seems logical as well.

How about df1c289a15b39205371e6189916a7c271efbe55a?

leogermani:
Hi @oekeur 

>  Still, you're then never really sure if cap has run already. 

Why not? You are. If the plugin is enabled you know that the CAP filter has already been fired. There's no doubt about that.

I'm struggling to see a good justification for this new hook.

All the methods used in that callback are public, like `is_post_type_enabled` and `current_user_can_set_authors`, so you are able to know exactly what happened there and act accordingly.

If there was something that could only checked in that context, and we wanted to run something conditionally, it would make sense to add a hook inside that condition.

But not only all the conditions are easily to check from outside of that context, you added the hook after the `if ` and `else` .. so it will just run anyways... it's the exact same thing as having another hook at a later priority

leogermani:
Maybe another approach, that would satisfy the inconviniences you're having on https://github.com/Automattic/Co-Authors-Plus/issues/1024 is to increase the priority of the `save_post` hook in CAP [here](https://github.com/Automattic/Co-Authors-Plus/blob/df1c289a15b39205371e6189916a7c271efbe55a/php/class-coauthors-plus.php#L49). We could reduce it to say 8.

Then whenever someone hooks into the default priority of `save_post`, CAP will already has done its thing

Issue #1059, opened 2024-09-11
allow_indexing_guest_author_archive for Yoast compatibility ignores `Show author archives in search results` setting in Yoast


mchanDev:
The code block here, ignores the Yoast settings that disable search indexing of Author archives.

https://github.com/Automattic/Co-Authors-Plus/blob/37065d289a9da07aa5228aa4969c8f96576a58c0/php/integrations/yoast.php#L281-L316

The relevant option in Yoast appears to be used here `$this->options->get( 'noindex-author-wpseo', false )`
https://github.com/Yoast/wordpress-seo/blob/0a1623546fed6266f466035dcf63d41209542060/src/presentations/indexable-author-archive-presentation.php#L121-L124

I believe adding the same options check should address this

PR #1062, opened 2024-09-24
Patch functionality to allow Yoast to control if author pages should be indexed or not

mchanDev:
## Description

This PR aims to address the issue noted in #1059 

Yoast has introduced functionality to allow control over the indexing of author pages.

This option is controlled with an option called `noindex-author-wpseo`

## Steps to Test
On a setup with Yoast and CoAuthors Plus installed
Set `noindex-author-wpseo` to true, by toggling `Show author archives in search results` to the off position
Observe that author pages are included in the sitemap
Checkout CoAuthors Plus to this branch
Observe that author archive pages are not included in the sitemap

https://yoast.com/help/how-to-exclude-author-pages-from-sitemap/#customize

### DIFF
--- a/php/integrations/yoast.php
+++ b/php/integrations/yoast.php
@@ -305,11 +305,14 @@ class Yoast {
 		}
 		
 		/*
-		 * Check if the `noindex-author-wpseo` option
-		 * 		which is set in Yoast, this setting drops author pages being indexed
-		 * if set to true do not change robots
+		 * Respect Yoast's "Show Authors archives in search results?" setting.
+		 *
+		 * This is stored in Yoast's serialized `wpseo_titles` option, not as a
+		 * standalone WordPress option, so it must be read via the Yoast options API.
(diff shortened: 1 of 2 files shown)


GaryJones:
Thanks for this, @mchanDev, and sorry for the long delay. The intent is good, but I think the patch is a no-op as written.

Yoast doesn't expose `noindex-author-wpseo` as a standalone option — it's stored as a key inside Yoast's aggregated `wpseo_titles` option group. So `get_option( 'noindex-author-wpseo', false )` returns the default `false` every time and the early return never fires, meaning the setting isn't actually honoured. To read it you'd want Yoast's own accessor, e.g. `WPSEO_Options::get( 'noindex-author-wpseo' )` (or `YoastSEO()->helpers->options->get( … )`). A small test mocking that option would also let us verify the behaviour.

We're deferring this past 4.1.0. If you're able to rework it to read the Yoast option correctly (with a test), we'd be glad to take it. Leaving it open. Thanks again.

PR #1065, opened 2024-10-18
fix: prevent the backfill from running forever.

eddiesshop:
## Description
There's an edge case where an author that no longer exists can still be assigned to a post. This throws the backfill script into an infinite loop, because the respective author-term is never found/created, and so the underlying problem of missing author-term records is never resolved. The infinite loop is started when at the end of the while loop, the script asks for "remaining posts which need author terms" and so it returns the same rows over and over.

This fix addresses this in 2 ways:
1. If an author is not found, we insert a postmeta record indicating that it should be skipped for processing. This prevents the loop continuously returning the same post for processing.
2. Checks have been added so that the script can't go beyond what should be the maximum number of rows needing to be addressed.

## Deploy Notes

No.

## Steps to Test

1. Create a handful of new posts. Make sure they're published.
2. Confirm what the maximum User ID is with `SELECT MAX(ID) FROM wp_users`.
3. Update the `post_author` column for those posts to IDs that *do not exist* in the `wp_users` table (any ID above MAX(ID)).
4. Run the command (`wp co-authors-plus create-author-terms-for-posts`). Be prepared to kill it. Notice how it goes beyond 100%.
5. Check out the PR.
6. Re-run the command from Step 4. This time, notice how the posts with non-existing authors have been skipped for processing, and the command finishes running on its own.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1260,7 +1260,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	 *
 	 * @return WP_User
 	 */
-	public function get_first_admin_user() {
+	private function get_first_admin_user() {
 		if ( ! wp_cache_get( 'co-authors-plus-most-prolific-author', 'co-authors-plus' ) ) {
 			$admin_user_query = new WP_User_Query(
 				[
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -192,6 +192,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 					$author = get_user_by( 'id', $record->post_author );
 
 					if ( false === $author ) {
+						// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users -- This is just trying to convey where the root problem should be resolved.
 						WP_CLI::warning( sprintf( 'Post Author ID %d does not exist in %s table, inserting skip postmeta (`%s`).', $record->post_author, $wpdb->users, self::SKIP_POST_FOR_BACKFILL_META_KEY ) );
 						$this->skip_backfill_for_post( $record->post_id, 'nonexistent_post_author_id' );
 						continue;
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -196,7 +196,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 					$author = get_user_by( 'id', $record->post_author );
 
 					if ( false === $author ) {
-						$author = $this->get_most_prolific_author();
+						$author = $this->get_first_admin_user();
 						WP_CLI::warning( sprintf( 'Must transfer posts from User ID: %d to Author ID: %d (%s)', $record->post_author, $author->ID, $author->user_nicename ) );
 						$author_trans[ $record->post_author ] = $author->ID;
 						$record->post_author                  = $author->ID;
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -1233,6 +1255,185 @@ private function get_posts_with_missing_terms( $author_taxonomy, $post_types = [
...
 	}
 
+	/**
+	 * This function handles obtaining an author account which should have the most posts assigned to it. If unable
+	 * to find an appropriate account, this function will attempt to create an author account for use.
+	 *
+	 * @return WP_User
+	 * @throws Exception If unable to successfully create an author user account.
+	 */
+	public function get_most_prolific_author() {
```
leogermani, line 1265: This is a very opinionated approach.. It makes sense.. but I think creating an author with a random display name doesn't.

What do you think of simply assigning them to the admin instead?

brookegs:
Unsubscribe me from this list. Thanks


Luanne Rife
540-467-2189

php/class-wp-cli.php
```diff
@@ -1233,6 +1255,185 @@ private function get_posts_with_missing_terms( $author_taxonomy, $post_types = [
...
+			global $wpdb;
+
+			// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+			$most_prolific_users = $wpdb->get_results(
+				'SELECT
+					u.ID,
+					u.user_email,
+					u.display_name,
+					COUNT(p.ID) as true_count
+				FROM wp_posts p
```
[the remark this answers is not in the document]
eddiesshop: Thanks @iuravic! I ended up removing the query entirely to go with obtaining the first available admin account instead.

php/class-wp-cli.php
eddiesshop: My initial thought with going with this approach was obfuscation. Going with an admin account as you're suggesting avoids creating random user accounts though, and seems like a cleaner solution, so I've updated the code to use that approach instead.

eddiesshop: My mind was changed once again, this time by @naxoc 's suggestion that we use postmeta markers to indicate whether a post should be skipped if the `post_author` doesn't exist. This is a better approach, as it would leave things as they are, allowing web admins to figure out which User ID no longer exists and investigate as to why.

php/class-wp-cli.php
```diff
@@ -135,7 +137,7 @@ public function create_terms_for_posts(): void {
 	 * @subcommand create-author-terms-for-posts
 	 * @synopsis [--post-types=<csv>] [--post-statuses=<csv>] [--unbatched] [--records-per-batch=<records-per-batch>] [--specific-post-ids=<csv>] [--above-post-id=<above-post-id>] [--below-post-id=<below-post-id>]
 	 * @return void
-	 * @throws Exception If above-post-id is greater than or equal to below-post-id.
+	 * @throws Exception If above-post-id is greater than or equal to below-post-id, or if unable to obtain a prolific author account.
```
naxoc, line 140: Does it still throw an exception? If it does we should maybe adjust the "prolific author" part.

php/class-wp-cli.php
```diff
@@ -160,6 +162,7 @@ public function create_author_terms_for_posts( $args, $assoc_args ) {
 		WP_CLI::line( sprintf( 'Found %d posts with missing author terms.', $count_of_posts_with_missing_author_terms ) );
 
 		$authors      = [];
+		$author_trans = [];
```
naxoc, line 165: This var is unused.

eddiesshop: Good catch

php/class-wp-cli.php
eddiesshop: Good catch

naxoc (approved):
This looks good to go!

iuravic (approved):

PR #1069, opened 2024-10-24
Fix the usage of the 'manage_posts_custom_column' hook

### LINKED ISSUE #1068: List table $post is null in exports
There are some (I believe, legitimate) cases where the Co-Authors column is failing to generate the correct values. One such example is the exports created in the [Admin Columns](https://wordpress.org/plugins/codepress-admin-columns/) plugin. I believe their plugin renders the list tables using WP_Query, but not actually doing a Loop. They, I suppose, rely on the  [`manage_{$post->post_type}_posts_custom_column`](https://developer.wordpress.org/reference/hooks/manage_post-post_type_posts_custom_column/) hook being used as advertised, where the post ID is taken from the callback's arguments. In the current Co-Authors code, the ID is taken from the `global $post;`, which, in these use cases, is not set.

The result is the "warning" below, which is accompanied by in an empty Co-Authors column in the export.

```
PHP Warning:  Attempt to read property "ID" on null in /var/www/html/wp-content/plugins/co-authors-plus/php/class-coauthors-plus.php on line 580
PHP Stack trace:
PHP   1. {main}() /var/www/html/wp-admin/edit.php:0
PHP   2. WP_Posts_List_Table->prepare_items() /var/www/html/wp-admin/edit.php:235
PHP   3. wp_edit_posts_query($q = *uninitialized*) /var/www/html/wp-admin/includes/class-wp-posts-list-table.php:165
PHP   4. wp($query_vars = ['post_type' => 'post', 'post_status' => '', 'perm' => '', 'order' => 'desc', 'orderby' => '15b2a971b5c28f', 'posts_per_page' => 200]) /var/www/html/wp-admin/includes/post.php:1306
PHP   5. WP->main($query_args = ['post_type' => 'post', 'post_status' => '', 'perm' => '', 'order' => 'desc', 'orderby' => '15b2a971b5c28f', 'posts_per_page' => 200]) /var/www/html/wp-includes/functions.php:1336
PHP   6. WP->query_posts() /var/www/html/wp-includes/class-wp.php:816
PHP   7. WP_Query->query($query = ['order' => 'desc', 'orderby' => '15b2a971b5c28f', 'author_name' => '[REDACTED]', 'post_type' => 'post', 'posts_per_page' => 200, 'post_status' => '', 'perm' => '']) /var/www/html/wp-includes/class-wp.php:696
PHP   8. WP_Query->get_posts() /var/www/html/wp-includes/class-wp-query.php:3852
PHP   9. apply_filters_ref_array($hook_name = 'the_posts', $args = [0 => [0 => class WP_Post { ... }, 1 => class WP_Post { ... }, 2 => class WP_Post { ... }, 3 => class WP_Post { ... },...
```

justinmaurerdotdev:
Fixes #1068.

## Description
Per the #1068 issue, there are some use cases where the 'manage_posts_custom_column' hook is used outside of The Loop. In current versions of the plugin, this scenario produces an empty column. This PR refactors the callback for that hook to use the passed $post_id instead of using the global $post variable.

## Testing
You can test that nothing has broken by viewing the Posts table and checking that the Authors column displays the same as before. I'm not 100% sure of a reliable test for the edge case, besides installing a plugin like [Admin Columns](https://wordpress.org/plugins/codepress-admin-columns/) and running the export from the Posts table.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -231,8 +231,8 @@ class CoAuthors_Plus {
 		// Hooks to add additional co-authors to 'authors' column to edit page
 		add_filter( 'manage_posts_columns', array( $this, '_filter_manage_posts_columns' ) );
 		add_filter( 'manage_pages_columns', array( $this, '_filter_manage_posts_columns' ) );
-		add_action( 'manage_posts_custom_column', array( $this, '_filter_manage_posts_custom_column' ) );
-		add_action( 'manage_pages_custom_column', array( $this, '_filter_manage_posts_custom_column' ) );
+		add_action( 'manage_posts_custom_column', array( $this, '_filter_manage_posts_custom_column' ), 10, 2 );
+		add_action( 'manage_pages_custom_column', array( $this, '_filter_manage_posts_custom_column' ), 10, 2 );
 
 		// Add quick-edit co-author select field
 		add_action( 'quick_edit_custom_box', array( $this, '_action_quick_edit_custom_box' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks for this, @justinmaurerdotdev — it's the right fix for a still-live bug (#1068), where the Authors column reads `global $post` and blows up when a list table is rendered outside The Loop (Admin Columns exports and similar).

I've taken it over in #1309: your commit is cherry-picked across with authorship preserved and rebased onto current `develop`, so it keeps the newer `data-*` byline markup that had been added to the same method since you opened this. I also added a regression test that renders the column from a post ID with no `global $post` set (it fails without your fix and passes with it).

Closing in favour of #1309. Credit for the fix is yours — thank you.

PR #1074, opened 2024-11-01
fix invalid sql in assign-user-to-coauthor subcommand

jkeasley:
## Description

ensure that the list of post-types is correctly wrapped in single quotes to ensure that the resulting sql is valid.

currently $post_types evaluates to `post','page` resulting in an invalid sql query, after the change $post_type will evaluate to `'post','page'` with is the expected syntax for this sql query

## Deploy Notes

N/A

## Steps to Test

Run `wp co-authors-plus assign-user-to-coauthor --user_login=<user_login> --coauthor=<cap-user_login> `  confirm that the command executes as expected without the error `WordPress database error You have an error in your SQL syntax;` being returned

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -426,7 +426,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 			WP_CLI::error( __( 'Please specify a valid co-author login', 'co-authors-plus' ) );
 		}
 
-		$post_types = implode( "','", $coauthors_plus->supported_post_types() );
+		$post_types = "'" . implode( "','", $coauthors_plus->supported_post_types() ) . "'";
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks for the patch, @jkeasley.

Closing this as already resolved — the same SQL quoting bug was fixed on `develop` by commit 5ab2946 ("Fix #1161: SQL syntax error in CLI command when migrating users to guest authors") in December 2025. The current line wraps the interpolation in outer single quotes in the SQL template itself:

```php
$posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('{$post_types}')", $user->ID ) );
```

So the underlying problem is resolved; your patch targets an earlier state of the file. Appreciate the contribution.

PR #1090, opened 2024-12-19
fix: verify Guest Authors when running CLI commands

ronchambers:
## Description

Co-authors-plus allows developers to use the filter `add_filter( 'coauthors_guest_authors_enabled', '__return_false' )`, but when developers implement this filter, many of the Co-Authors-Plus CLI commands will then throw FATAL PHP errors.

This PR will verify that when a developers uses the filter `add_filter( 'coauthors_guest_authors_enabled', '__return_false' )`, that the CLI commands that need Guest Authors will display an error and die gracefully as opposed to throwing a fatal error in the debug.log.

This PR may not be inclusive of all the Co-Authors-Plus CLI commands that need to verified.  I simply searched for `$coauthors_plus->guest_authors` in each command; and if found, I added `$this->verify_guest_authors_or_die();`.

I did not fully test all commands nor all edge cases as I'm not a Co-Authors-Plus expert.

Example fatal error (without this PR) when running CLI commands:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_guest_author_by() on null in wp-content/plugins/co-authors-plus/php/class-wp-cli.php:1087
Stack trace:
#0 wp-content/plugins/co-authors-plus/php/class-wp-cli.php(995): CoAuthorsPlus_Command->create_guest_author()
#1 [internal function]: CoAuthorsPlus_Command->create_author()
```

## Deploy Notes

No new dependencies were added.

## Steps to Test

1. Do not pull this PR yet.
2. Implement `add_filter( 'coauthors_guest_authors_enabled', '__return_false' )` in a separate plugin.
3. Run Co-Authors-Plus CLI commands and see that they throw FATAL errors.
4. Pull this PR now.
5. Re-run CLI commands and see that FATALs are no longer thrown and instead WP_CLI will show an error message to the user (and die gracefully).

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -24,6 +24,8 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	public function create_guest_authors( $args, $assoc_args ): void {
 		global $coauthors_plus;
 
+		$this->verify_guest_authors_or_die();
+
 		$defaults = array(
 				// There are no arguments at this time
 		);
@@ -576,6 +578,8 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	public function rename_coauthor( $args, $assoc_args ): void {
 		global $coauthors_plus, $wpdb;
 
+		$this->verify_guest_authors_or_die();
+
 		$defaults   = array(
 			'from' => null,
 			'to'   => null,
@@ -830,6 +834,9 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	 */
 	public function update_author_terms(): void {
 		global $coauthors_plus;
+
+		$this->verify_guest_authors_or_die();
+
 		$author_terms = get_terms( $coauthors_plus->coauthor_taxonomy, array( 'hide_empty' => false ) );
 		WP_CLI::log( 'Now updating ' . count( $author_terms ) . ' terms' );
 		foreach ( $author_terms as $author_term ) {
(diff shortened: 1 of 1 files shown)


leogermani:
This avoids the Fatals, but ideally these commands should work if guest authors are disabled.. :/ But that's I guess would require a bigger refactor

GaryJones:
Thanks for this, @ronchambers, and apologies for the slow turnaround. We're going to defer this past 4.1.0 and into the more CLI-focused 4.2.0, but it needs a little rework first:

1. **Rebase** — it's currently conflicting with `develop` (`class-wp-cli.php` has moved on since this was opened).
2. **Narrow the guard** — `verify_guest_authors_or_die()` is right for the guest-only commands (`create_guest_authors`, the CSV importer), but `update_author_terms()` and `rename_coauthor()` also operate on WP-user co-authors and work fine with Guest Authors disabled. Dying there would break legitimate non-guest use, so the guard should apply only to the commands that genuinely require Guest Authors.
3. **Tests** — a couple covering the die / no-die paths would let us merge with confidence.

Leaving it open and milestoned for 4.2.0. Thanks again for the contribution.

ronchambers:
@GaryJones - thanks for the response, but I'll probably never have the time to work on this.  I should have just submitted my concerns above as a github issue 🙂  but back then I was testing my assumptions in the code so I just submitted the PR instead.  Anyway, moving forward though, this is no longer an issue - we found a solution around it.  Best of luck!

PR #1098, opened 2025-03-03
Use WordPress 6.8 filter to skip heavy, unnecessary query

ethitter:
## Description

When the plugin is active, it replaces the post-count column in the Users list table with its own, which pulls the counts from the author terms. By default, Core uses `count_many_users_posts()`, which is uncached and can be slow when a site has many posts associated with a single user.

Introduced in WordPress 6.8, the `pre_count_many_users_posts` filter can be used to skip the unnecessary query: https://core.trac.wordpress.org/ticket/63004.

## Deploy Notes

N/A

## Steps to Test

1. Install WordPress 6.8
2. Install and activate Query Monitor
3. Load the Users list table, open Query Monitor's Database Queries panel, and select `count_many_users_posts` from the "Caller" filter to confirm that the query is run
4. Activate Co-Authors Plus on this PR's branch, reload the Users list table, and confirm that Query Monitor shows no queries originating from `count_many_users_posts()`

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -243,6 +243,7 @@ class CoAuthors_Plus {
 
 		// Apply some targeted filters
 		add_action( 'load-edit.php', array( $this, 'load_edit' ) );
+		add_action( 'load-users.php', array( $this, 'load_users_screen' ) );
 	}
 
 	/**
@@ -1507,6 +1508,16 @@ class CoAuthors_Plus {
 		$views        = array_reverse( $views );
 
 		return $views;
+	}
+
+	/**
+	 * Prevent WordPress from counting users' posts for Users table column that
+	 * is removed by the `_filter_manage_users_columns` method.
+	 *
+	 * @return void
+	 */
+	public function load_users_screen(): void {
+		add_filter( 'pre_count_many_users_posts', '__return_false' );
 	}
 
 	/**
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1517,7 +1517,22 @@ class CoAuthors_Plus {
 	 * @return void
 	 */
 	public function load_users_screen(): void {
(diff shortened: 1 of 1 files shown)


GaryJones (changes requested):
Thanks @ethitter .

While the query does indeed stop on WP 6.8-beta1, there seems to be a PHP Warning introduced on this PR branch that I don't get with `develop`:

```
Trying to access array offset on value of type bool  
wp-admin/includes/class-wp-users-list-table.php:415

`WP_Users_List_Table->display_rows()`  
wp-admin/includes/class-wp-list-table.php:1712

`WP_List_Table->display_rows_or_placeholder()`  
wp-admin/includes/class-wp-list-table.php:1639

`WP_List_Table->display()`  
wp-admin/users.php:814
```

I quickly tried `__return_empty_array` but that then gave "Undefined array key 1", "Undefined array key 2", etc. at `wp-admin/includes/class-wp-users-list-table.php:415`.

Looking at [that line](https://github.com/WordPress/WordPress/blob/239f2e0f0c9b96ae030823e2169c997ecc839ca8/wp-admin/includes/class-wp-users-list-table.php#L415), I see that there's an `isset()` check on the results of the result of the call to `count_many_users_posts()`. So, we'd want it to be `null` to be skipped over, but, that's exactly the (non-value) we can't have else the new filter won't return. I suspect line 415 might need an update to be able to take advantage of this filter.

Alternatively, we'd want the new method you added here not to return an empty array, but an array filled with enough keys to match users, and values of `0`.

Can you take another look please?

ethitter:
Ah, sounds like the Core update needs a tweak, to return an array with the expected keys. I'll create a followup for that.

GaryJones (approved):
On WP 6.8, the extra query is stopped. No adverse impact is seen.

On WP 6.7, the extra query is not stopped (because the new core filter doesn't exist), again, with no adverse impact.

Approving.

PR #1099, opened 2025-03-06
Ensure term is valid before deleting term

srtfisher:
Fix an issue where this notice was being thrown:

```
1) /Users/srtfisher/broadway/www/mantle/wp-content/plugins/co-authors-plus/php/class-coauthors-plus.php:1121
Attempt to read property "term_id" on false
```


GaryJones:
Already handled by #1166.

PR #1100, opened 2025-03-07
Fix SQL Syntax Error in assign_user_to_coauthor

burhandodhy:
## Description

This PR resolves a SQL syntax error in the `assign_user_to_coauthor` method. The previous implementation was missing the required single quotes around values, causing a syntax error. 

## Deploy Notes

## Steps to Test

Run `wp co-authors-plus assign-user-to-coauthor` command and make sure it works find

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -429,8 +429,8 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 		$post_types = implode(
 			', ',
 			array_map(
-				function ( $type ) {
-					return "'{$type}'";
+				function ( $post_type ) {
+					return "'{$post_type}'";
 				},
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks for the patch, @burhandodhy.

Closing this as already resolved — the same SQL quoting bug was fixed on `develop` by commit 5ab2946 ("Fix #1161: SQL syntax error in CLI command when migrating users to guest authors") back in December 2025. The current line reads:

```php
$posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_author=%d AND post_type IN ('{$post_types}')", $user->ID ) );
```

The wrapping quotes around `{$post_types}` in the SQL template make the query valid; your patch targets an earlier state of the file. Appreciate the fix regardless.

PR #1104, opened 2025-03-21
Increase composer/installers version

NickGreen:
### Changes in this PR

A small maintenance update which changes composer/installers version to `^2` to make it compatible with later versions of PHP.


GaryJones (approved):

GaryJones:
Failing tests are unrelated.

PR #1108, opened 2025-03-27
Adjust import of PluginDocumentSettingPanel to resolve deprecation notice

kadamwhite:
## Description

Resolves a console warning encountered when using the current version of the plugin, encountered since this component was moved from the "editPost" to "editor" package within Gutenberg.

> wp.editPost.PluginDocumentSettingPanel is deprecated since version 6.6. Please use wp.editor.PluginDocumentSettingPanel instead.

Documentation for PluginDocumentSettingPanel: https://developer.wordpress.org/block-editor/reference-guides/slotfills/plugin-document-setting-panel/

## Deploy Notes

- None

## Steps to Test

- Activate this plugin on a WordPress 6.6.x or 6.7.x website
- Set SCRIPT_DEBUG to `true` in wp-config (otherwise these warnings will not be shown)
- Load the editor
- Verify no deprecation notice is shown relating to this plugin's code

### DIFF
--- a/src/__tests__/plugin-document-setting-panel.test.js
+++ b/src/__tests__/plugin-document-setting-panel.test.js
@@ -1,0 +1,66 @@
+/**
+ * Tests for PluginDocumentSettingPanel backwards compatibility.
+ *
+ * The component moved from @wordpress/edit-post to @wordpress/editor in WP 6.6.
+ * We use a runtime fallback to support both WP 6.4-6.5 (edit-post) and WP 6.6+ (editor).
+ */
+
+describe( 'PluginDocumentSettingPanel compatibility', () => {
+	const MockComponent = () => 'MockPluginDocumentSettingPanel';
+
+	beforeEach( () => {
+		// Reset the wp global before each test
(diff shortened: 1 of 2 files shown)


GaryJones:
Thanks for the original fix, @kadamwhite! 🎉

Just to explain the additional commit: your change to import from `@wordpress/editor` was the right fix for WordPress 6.6+, but we discovered it would break compatibility with WordPress 6.4-6.5 where `PluginDocumentSettingPanel` only exists in `@wordpress/edit-post`.

Since the plugin supports WP 6.4+, we implemented a runtime fallback pattern that:
- Uses `wp.editor.PluginDocumentSettingPanel` when available (WP 6.6+) — no deprecation warning
- Falls back to `wp.editPost.PluginDocumentSettingPanel` (WP 6.4-6.5) — maintains compatibility

This ensures the deprecation warning is resolved for users on newer WordPress versions whilst keeping the plugin working for those still on 6.4 or 6.5.

Added a Jest test to document this requirement and prevent future regressions.

Issue #1113, opened 2025-05-08
Warnings when using the filter ‘coauthors_supported_post_types’ to restrict the plugin functionalities to a unique custom post type


garubi:
### The context

- WordPress 6.8.1 with one (or more) Custom Post Type added
- Co-Authors-Plus 3.6.5 installed and activated
- the filter `coauthors_supported_post_types` used to restrict the plugin functionalities to the Custom Post Type only:
```php
add_filter( 'coauthors_supported_post_types', 'pandora_coauthors_supported_post_types' );
function pandora_coauthors_supported_post_types($post_types_with_authors){
	return array( 'sedi' );
}
``` 

### The issue
In the error log we have warnings like this on every page view
```
[08-May-2025 14:44:58 UTC] PHP Warning:  Attempt to read property "cap" on null in C:\Users\Stefano Garuti\Local Sites\co-author-debug\app\public\wp-content\plugins\co-authors-plus\php\class-coauthors-plus.php on line 1594
[08-May-2025 14:44:58 UTC] PHP Warning:  Attempt to read property "edit_post" on null in C:\Users\Stefano Garuti\Local Sites\co-author-debug\app\public\wp-content\plugins\co-authors-plus\php\class-coauthors-plus.php on line 1594
[08-May-2025 14:44:58 UTC] PHP Warning:  Attempt to read property "cap" on null in C:\Users\Stefano Garuti\Local Sites\co-author-debug\app\public\wp-content\plugins\co-authors-plus\php\class-coauthors-plus.php on line 1595
[08-May-2025 14:44:58 UTC] PHP Warning:  Attempt to read property "edit_others_posts" on null in C:\Users\Stefano Garuti\Local Sites\co-author-debug\app\public\wp-content\plugins\co-authors-plus\php\class-coauthors-plus.php on line 1595
[08-May-2025 14:44:58 UTC] PHP Warning:  Attempt to read property "cap" on null in C:\Users\Stefano Garuti\Local Sites\co-author-debug\app\public\wp-content\plugins\co-authors-plus\php\class-coauthors-plus.php on line 1596
[08-May-2025 14:44:58 UTC] PHP Warning:  Attempt to read property "read_post" on null in C:\Users\Stefano Garuti\Local Sites\co-author-debug\app\public\wp-content\plugins\co-authors-plus\php\class-coauthors-plus.php on line 1596
``` 

### Look at the issue live on WP Playground
[...]

PR #1114, opened 2025-05-09
Avoids Warnings when using the filter ‘coauthors_supported_post_types’ with a unique custom post type. Fix issue #1113

garubi:
## Description

Adds a check to avoid warnings in logfile when using the `coauthors_supported_post_types` filter to restrict the plugin functionalities to a unique custom post type.
For details see issue #1113

The code assumes that the function `get_post_type_object()` returns an object, but it can also return NULL (see [WordPress Developer Resources](https://developer.wordpress.org/reference/functions/get_post_type_object) ).
So before using the returned value we need to check for it.

This is what the added `if` statemen of this PR does.


garubi:
Please accept this Pull Request: it's really a easy fix (it should be a best practice adding check like the one implemented here) and will avoid lot of unuseful warning in the logs...

Issue #551, opened 2018-07-04
CAP removes the Author field in bulk edit actions


TheCrowned:
Without CAP, it's possible to bulk edit post author:

![wo cap 2](https://user-images.githubusercontent.com/7880569/42264647-8cf2e222-7f71-11e8-9097-7f27ac080750.png)

With CAP active, the field vanishes and no coauthor field is added in replacement:

![w cap](https://user-images.githubusercontent.com/7880569/42264667-9a555e36-7f71-11e8-9f95-f4d68d70a8e8.png)

psaikali:
I think we should wait to fix this issue, here's why.
I started working on a fix for that in a separate forked repo: https://github.com/psaikali/Co-Authors-Plus/commit/2ac1afbacaec965bbd35bbdd569b8eeb3c160c03

![cap-bulk-edit-coauthors](https://user-images.githubusercontent.com/1606655/43582595-42d9b8de-965d-11e8-957f-42788fdcac1c.jpg)
As you can see, the UI is ready. 
I just need to process the bulk action when editing multiple posts, and that's where things get complicated.

The main core function taking care of this bulk editing action is `bulk_edit_posts()`. Unfortunately, this function does not offer any hook to let us save additional data when bulk editing.
[There is a ticket for that here](https://core.trac.wordpress.org/ticket/28112) and it looks like the `bulk_edit_posts` hook will show up in 5.0 (very soon).

IMHO, it's better to wait for this official hook and use it to easily process the bulk-edit action and assign co-authors to multiple posts at the same time. I started to find another way to do that, by hooking to another action (`wp_insert_post` is triggered by the `bulk_edit_posts()` function) and checking for $_POST data, but I feel that this is a hacky way of doing things knowing that a proper hook might be available in a couple of months.

Any feedback is appreciated!

DailenG:
Just ran into this myself. Looks like we'll have to wait for 5.0.2 for the bulk_edit_posts to be completed.

jadlimcaco:
Is this fix still in progress?

colorcrate:
This is still an issue as of May, 2020, any update?

jxxe:
This would be so useful! Any word?

melmo:
Also looking for this update

RubenatorX:
Bump

TheCrowned:
@psaikali are you still working on this? Otherwise I can take over and hopefully fix it for good :)

[...]

PR #747, opened 2020-10-03
Allow bulk edit of coauthors

TheCrowned:
Fixes #551.

Bulk edit of coauthors has been disabled for a long time. This PR finishes work initiated by https://github.com/psaikali/Co-Authors-Plus/commit/2ac1afbacaec965bbd35bbdd569b8eeb3c160c03 and makes it fully working, by actually updating the coauthors on bulk edit actions. 

Screenshot from 2020-10-03 21-29-06

PR #1120, opened 2025-06-02
Add bulk editing support for Co-Authors in the admin interface

mattradford-sage:
## Description

This PR adds Bulk Editing support to Co-Authors.

It finishes work started in https://github.com/psaikali/Co-Authors-Plus/commit/2ac1afbacaec965bbd35bbdd569b8eeb3c160c03 and https://github.com/Automattic/Co-Authors-Plus/pull/747.

It fixes https://github.com/Automattic/Co-Authors-Plus/issues/551.

Props to: @psaikali and @TheCrowned

## Deploy Notes

None.

## Steps to Test

Bulk editing needs to be tested on standard Posts and CPTs with WordPress versions >=6.3.
 - `bulk_edit_posts` was [introduced in 6.3](https://developer.wordpress.org/reference/functions/bulk_edit_posts/).

1. Check out the PR.
2. Ensure you have multiple users with roles capable of [editing other users' posts](https://wordpress.org/documentation/article/roles-and-capabilities/#editor).
3. Bulk edit Posts and assign new users.
4. No errors or warnings should be present either in the admin or in the error log.
5. Repeat these tests using: 
* a CPT.
* setting one Author.
* not changing the Author(s).

### Notes

* The fatal error [noted in a previous PR](https://github.com/Automattic/Co-Authors-Plus/pull/747#issuecomment-721961415) could not be replicated.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -108,11 +108,15 @@ class CoAuthors_Plus {
 		// REST API: Depending on user capabilities, hide author term description.
 		add_action( 'rest_prepare_author', array( $this, 'conditionally_hide_author_term_description' ) );
 
-		// Add Co-Author select field to the Bulk Edit actions form.
-		add_action( 'bulk_edit_custom_box', array( $this, '_action_bulk_edit_custom_box' ), 10, 2 );
+		// Add Bulk Edit support on supported versions of WordPress.
+		global $wp_version;
+		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
+			// Add Co-Author select field to the Bulk Edit actions form.
+			add_action( 'bulk_edit_custom_box', array( $this, '_action_bulk_edit_custom_box' ), 10, 2 );
 
-		// Update Co-Authors when bulk editing posts.
-		add_action( 'bulk_edit_posts', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
+			// Update Co-Authors when bulk editing posts.
+			add_action( 'bulk_edit_posts', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
+		}
 	}
 
 	/**
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -112,13 +112,7 @@ class CoAuthors_Plus {
 		add_action( 'bulk_edit_custom_box', array( $this, '_action_bulk_edit_custom_box' ), 10, 2 );
 
 		// Update Co-Authors when bulk editing posts.
-		// bulk_edit_posts was introduced in WordPress 6.3, so we need to check the version.
-		global $wp_version;
-		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
-			add_action( 'bulk_edit_posts', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
-		} else {
-			add_action( 'wp_insert_post_data', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
-		}
+		add_action( 'bulk_edit_posts', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
 	}
 
 	/**
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -2026,13 +2026,13 @@ class CoAuthors_Plus {
 			return;
 		}
 		?>
-		<label class="bulk-edit-group bulk-edit-coauthors">
+		<label class="bulk-edit-group bulk-edit-coauthors" aria-labelledby="coauthors-edit-desc">
 			<span class="title"><?php esc_html_e( 'Authors', 'co-authors-plus' ) ?></span>
-			<div id="coauthors-edit" class="hide-if-no-js">
-				<p><?php echo wp_kses( __( 'Leave the field below blank to keep the Authors unchanged. Any change here will overwrite all previously assigned Authors.', 'co-authors-plus' ), array( 'strong' => array() ) ); ?></p>
-			</div>
-			<?php wp_nonce_field( 'coauthors-edit', 'coauthors-nonce' ); ?>
 		</label>
+		<div id="coauthors-edit" class="inline-edit-group wp-clearfix hide-if-no-js">
+			<p id="coauthors-edit-desc"><?php echo wp_kses( __( 'Leave the field below blank to keep the Authors unchanged. Any change here will overwrite all previously assigned Authors.', 'co-authors-plus' ), array( 'strong' => array() ) ); ?></p>
+			<?php wp_nonce_field( 'coauthors-edit', 'coauthors-nonce' ); ?>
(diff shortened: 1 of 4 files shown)


php/class-coauthors-plus.php
```diff
@@ -107,6 +107,18 @@ public function __construct() {
...
+		// Add Co-Author select field to the Bulk Edit actions form.
+		add_action( 'bulk_edit_custom_box', array( $this, '_action_bulk_edit_custom_box' ), 10, 2 );
+
+		// Update Co-Authors when bulk editing posts.
+		// bulk_edit_posts was introduced in WordPress 6.3, so we need to check the version.
+		global $wp_version;
+		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
+			add_action( 'bulk_edit_posts', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
+		} else {
+			add_action( 'wp_insert_post_data', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
```
leogermani, line 120: I don't think we need or should keep backward compatibility with WP < 6.3... imo it's better if we limit this to WP > 6.3

php/class-coauthors-plus.php
```diff
@@ -1999,4 +2011,66 @@ public function conditionally_hide_author_term_description( WP_REST_Response $re
...
+		}
+
+		foreach( $postarr['post'] as $post_id ) {
+			$post = get_post( $post_id );
+
+			if ( $this->current_user_can_set_authors( $post ) && isset( $postarr['coauthors'] ) ) {
+					$coauthors = array_map( 'sanitize_title', (array) $postarr['coauthors'] );
+					$this->add_coauthors( $post_id, $coauthors );
+			} else {
+				// If a Co-Author isn't currently set, explicitly set one.
```
leogermani, line 2064: Can you share the rationale here?

Looks like you are just backfilling posts here in case they don't have a term. Looks like an arbitrary place to have such backfill. I don't think we should do that. We have a CLI for those situations like this...

mattradford-sage: This logic was unchanged from the previous PR. Now I've reviewed it I agree it is not required. Testing doesn't show any issues.
I've also corrected and tightened array checks.

php/class-coauthors-plus.php
leogermani: Thanks! I don't think we need to bum the `Requires at least` flag to 6.3. It's not that the whole plugin requires it. Instead we can just do a check for the version to decide whether to enable this particular feature or not.

php/class-coauthors-plus.php
```diff
@@ -107,6 +107,16 @@ public function __construct() {
...
+
+		// Add Bulk Edit support on supported versions of WordPress.
+		global $wp_version;
+		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
+			// Add Co-Author select field to the Bulk Edit actions form.
+			add_action( 'bulk_edit_custom_box', array( $this, '_action_bulk_edit_custom_box' ), 10, 2 );
+
+			// Update Co-Authors when bulk editing posts.
+			add_action( 'bulk_edit_posts', array( $this, 'action_bulk_edit_update_coauthors' ), 10, 2 );
+		}
```
GaryJones, line 119: Is the condition still needed here?

If the current WP is earlier than 6.3, then the `add_action()` won't do anything (other than update a global hooks array).

A version compare that will otherwise last until the minimum version is bumped isn't the worst thing in the world.

mattradford-sage: `bulk_edit_custom_box` was [introduced in 2.7.0](https://developer.wordpress.org/reference/hooks/bulk_edit_custom_box/), so we will need a version check somewhere.

Would you prefer to:
1) Keep the version compare as-is?
2) Revert to the previous compare, falling back to `wp_insert_post_data` for <6.3?

php/class-coauthors-plus.php
```diff
@@ -1999,4 +2009,57 @@ public function conditionally_hide_author_term_description( WP_REST_Response $re
...
+	 * This is used in the Bulk Edit screen to allow users to set Co-Authors
+	 * for multiple posts at once.
+	 *
+	 * @link https://github.com/Automattic/Co-Authors-Plus/issues/551
+	 * @param string $column_name The name of the column being edited.
+	 * @param string $post_type The post type being edited.
+	 * @return void
+	 */
+	public function _action_bulk_edit_custom_box( string $column_name, string $post_type ): void {
+		if ( 'coauthors' != $column_name || ! $this->is_post_type_enabled( $post_type ) || ! $this->current_user_can_set_authors() ) {
```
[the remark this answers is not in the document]
mattradford-sage: Done

php/class-coauthors-plus.php
GaryJones: Ah, OK. If we didn't have the version check then the UI would show it, but it wouldn't get saved. 

Leave as is then.

php/class-coauthors-plus.php
```diff
@@ -1999,4 +2009,57 @@ public function conditionally_hide_author_term_description( WP_REST_Response $re
...
+			return;
+		}
+		?>
+		<label class="bulk-edit-group bulk-edit-coauthors">
+			<span class="title"><?php esc_html_e( 'Authors', 'co-authors-plus' ) ?></span>
+			<div id="coauthors-edit" class="hide-if-no-js">
+				<p><?php echo wp_kses( __( 'Leave the field below blank to keep the Authors unchanged. Any change here will overwrite all previously assigned Authors.', 'co-authors-plus' ), array( 'strong' => array() ) ); ?></p>
```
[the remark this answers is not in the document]
GaryJones: Screenshot 2025-06-09 at 13 21 54

As can be seen here - the input (which is the thing we're trying to expose to assistive devices correctly) - doesn't have the correct label yet - it's now _only_ the description. (`Name: "Leave the..."`)

Why?

Partly because the `aria-*` is on the wrong element. It should be on the `input`, and not the `label`. 

Secondly, because the `aria-labelledby` and `aria-describedby` are being mixed up. The `aria-labelledby` should point to an ID of the span containing just "Authors", and the `aria-describedby` should point to the longer paragraph description.

See:

Screenshot 2025-06-09 at 13 45 29

(Don't miss the additional `id` element on the _Authors_ `span`.)

We would now have `Name: "Authors"` and `Description: "Leave the ..."`.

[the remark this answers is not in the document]
GaryJones: Let's go ahead with implementing the aria-* attributes as suggested here, please, so we can merge in the ability to bulk-edit, and then create a follow-up that does discovery on whether a JS function that outputs both the input and the label is feasible.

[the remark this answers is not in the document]
mattradford-sage: Addressed in https://github.com/Automattic/Co-Authors-Plus/pull/1120/commits/6bbd3298dc382481b02d220f23795296ad8dcc60

GaryJones:
I don't know when a 3.7 release may be made, but this code can be used in production now as it will be coming at some point!

PR #1127, opened 2025-06-26
Fix Gutenberg saving authors after PHP save post hooks have already run

carosioa:
## Description

Currently, when editing a post in Gutenberg, the author terms are not saved to the post itself until after the editor has finished saving: https://github.com/Automattic/Co-Authors-Plus/blob/14725a6f877fecd8853f40bf4cc5e9262f71e580/src/index.js#L39-L48

This creates issues when we have PHP methods hooked into any of the actions related to post saving and publishing (like save_post, post_status_transition, etc), as those hooks have already ran once the authors are saved, so the PHP code does not have access to the correct / up to date author terms for the post.

See issues: https://github.com/Automattic/Co-Authors-Plus/issues/1024 & https://github.com/Automattic/Co-Authors-Plus/pull/1053

To reproduce this, add a hook on save_post with 9999 priority and then do wp_get_post_terms( $post_id, 'author'). If the post hadn't been saved before, that will return empty. If it has been saved, but the authors have been edited since save, it will return outdated info.

My suggested PR is to add the author terms in the post at the same time as the React data store is updated, here:

https://github.com/carosioa/Co-Authors-Plus/blob/37dccd11c057d92646c569b3fb6ece55f169594c/src/components/co-authors/index.jsx#L102-L118

Then you can totally remove the subscribe action that is making api calls to add the authors and the state of the post in WordPress will always match with the Coauthors box. 

I did have to add each author term ID to the API response for the search box so it that was easily accessible in React. Not sure if there is a better place for looking that up than here:

https://github.com/carosioa/Co-Authors-Plus/blob/37dccd11c057d92646c569b3fb6ece55f169594c/php/class-coauthors-endpoint.php#L210-L224

## Deploy Notes

No new dependencies.

## Steps to Test

1. Switch to branch and start a new post in Gutenberg
2. Add multiple authors
3. Check in Console and wp.data.select('core/editor').getEditedPostAttribute( 'coauthors' ) should be up to date / show the same users as wp.data.select('cap/authors').getAuthors()


GaryJones:
Thanks for the patch, @carosioa — and sorry for the slow follow-up.

Closing this in favour of #1217, which has since landed on `develop` and takes the same architectural approach: routing author changes through `wp.data.dispatch('core/editor').editPost({ coauthors: [...termIDs] })`, exposing term IDs via REST, and removing the custom `subscribe()`-based save loop. #1217 is more comprehensive — it adds a dedicated `authors-by-term-ids` endpoint, a `useCoauthorDetails` hook with request-scoped caching, PHP-side bridging via `sync_coauthors_on_rest_save` / `backfill_coauthor_terms_in_rest`, `post_author` synchronisation, and deprecation notices for the legacy filters. It will ship in the upcoming 4.0 release.

One note worth flagging for anyone who finds this issue later: the specific symptom you described — a `save_post` callback at priority 9999 not seeing the current author terms — is not actually solved by either approach. WordPress's REST posts controller calls `wp_update_post()` (which fires `save_post`) BEFORE it calls `wp_set_object_terms()` to assign the payload's taxonomy terms. Pushing the terms into the payload gets them onto the post, but doesn't change that ordering. Consumers that need to read authors during save should migrate to `wp_after_insert_post` or `rest_after_insert_{post_type}`, which fire after taxonomies are assigned. That's the same hook #1217 uses for its own bridging logic.

Thanks again for the patch and the clear reproduction — both fed into the 4.0 design.

PR #1137, opened 2025-10-16
fix: avoid posts without authors on CLI and REST

leogermani:
## Description

The current checks in the save_post callback will create posts without proper coauthor terms in case an authenticated user creates a post via REST or CLI.

The issue only happens with users who can assign authors though. If you do the same thing with an `author`, it works just fine.

This PR makes sure that we only check for the user capabilities and nonce if `$_POST['coauthors']` is set, which is the scope of that check anyways.

All other cases should check if the post doesn't have an author and create author terms if needed. 

## Deploy Notes

Nothing special

## Steps to Test

On the current release or `develop`:

1. Take note of a admin or editor ID
2. Open WP Shell
3. Run:

```
wp_set_current_user( $admin_or_editor_id );
var_dump( wp_insert_post( [ 'post_title' => 'test', 'post_content' => 'test' ]) );
```

4. Check the database and confirm that this post does not have author terms associated with it.
5. Checkout this branch and repeat the steps above
6. Confirm the post now correctly has the logged in user as an author

You can run a similar test by adding a post via a REST API request, using an application password to authenticate.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -986,17 +986,15 @@ class CoAuthors_Plus {
 			return;
 		}
 
-		if ( $this->current_user_can_set_authors() ) {
+		if ( isset( $_POST['coauthors-nonce'], $_POST['coauthors'] ) && $this->current_user_can_set_authors() ) {
 			// if current_user_can_set_authors and nonce valid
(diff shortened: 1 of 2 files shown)


GaryJones (approved):
I added in some tests too.

PR #1164, opened 2025-12-05
Fix #1163: Create author command does not accept required arguments

### LINKED ISSUE #1163: CLI command create-author doesn't accept user arguments
The `create-author` command does not recognise all optional args, such as `--user_login=<user_login>`. When passing these arguments, the CLI returns an unrecognised arguments error and does not proceed. 

However, if the command is executed without the `user_email` or `user_login` arguments, the  `create_guest_author()` method will fail. This is because the method looks for existing users. After the first user with an empty `user_email` or `user_login` is created, subsequent users created without an email address will always throw a "user already exists" error because it is matching `user_email=""`.

NateWr:
## Description

See https://github.com/Automattic/co-authors-plus/issues/1163.

This solves the underlying problem, but leaves a few usability issues. 

1. It's possible that a site will want two different guest authors with the same display name. This PR checks for false matches against display name by also checking the email and login.
2. In fact, the `user_login` is a required argument, not an optional one. However, the CLI command returns a sensible error message if you try to create a user without one. So I haven't made any adjustments to this.

## Deploy Notes

No.

## Steps to Test

Try running `wp co-authors-plus create-author --display_name="Any user here"`. To see the original bug, you may need to run this twice with two different display names (and no email address).

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -981,13 +981,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	 *
 	 * @subcommand create-author
 	 * @synopsis
-	 * [--display_name=<display_name>]
-	 * [--user_login=<user_login>]
-	 * [--first_name=<first_name>]
-	 * [--last_name=<last_name>]
-	 * [--website=<website>]
-	 * [--user_email=<user_email>]
-	 * [--description=<description>]
+	 * [--display_name=<display_name>] [--user_login=<user_login>] [--first_name=<first_name>] [--last_name=<last_name>] [--website=<website>] [--user_email=<user_email>] [--description=<description>]
 	 */
 	public function create_author( $args, $assoc_args ): void {
 		$this->create_guest_author( $assoc_args );
@@ -1082,10 +1076,21 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	 */
 	private function create_guest_author( $author ): void {
 		global $coauthors_plus;
-		$guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'user_email', $author['user_email'], true );
 
-		if ( ! $guest_author ) {
+		if ( ! empty( $author['user_email'] ) ) {
+			$guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'user_email', $author['user_email'], true );
+		}
+
+		if ( ! $guest_author && ! empty ( $author['user_login'] ) ) {
 			$guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'user_login', $author['user_login'], true );
+		}
+
+		if ( ! $guest_author && ! empty ( $author['display_name'] ) ) {
+			$guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'display_name', $author['display_name'], true );
+
+			if ( $guest_author->user_email !== $author['user_email'] || $guest_author->user_login !== $author['user_login'] ) {
+				$guest_author = null;
+			}
 		}
 
 		if ( $guest_author ) {
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1077,20 +1077,14 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 	private function create_guest_author( $author ): void {
 		global $coauthors_plus;
 
+		$guest_author = false;
+
 		if ( ! empty( $author['user_email'] ) ) {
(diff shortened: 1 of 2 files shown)


GaryJones:
Thanks for tackling this issue, @NateWr! The approach of checking for non-empty fields before querying is exactly the right fix for the duplicate detection bug. I've spotted a couple of issues that should be addressed before merging:

### Critical: Missing null check (line 1091)

After fetching by `display_name` on line 1089, we need to verify `$guest_author` exists before accessing its properties:

**Current code:**
```php
if ( ! $guest_author && ! empty ( $author['display_name'] ) ) {
    $guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'display_name', $author['display_name'], true );

    if ( $guest_author->user_email !== $author['user_email'] || $guest_author->user_login !== $author['user_login'] ) {
        $guest_author = null;
    }
}
```

If `get_guest_author_by()` returns null/false (no match found), line 1091 will throw a PHP warning.

**Suggested fix:**
```php
if ( ! $guest_author && ! empty( $author['display_name'] ) ) {
    $guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'display_name', $author['display_name'], true );

    if ( $guest_author && ( $guest_author->user_email !== $author['user_email'] || $guest_author->user_login !== $author['user_login'] ) ) {
        $guest_author = null;
    }
}
```

### Logic issue: Validation should only check provided fields

The current validation logic has an edge case issue. Consider this scenario:

- User runs: `wp co-authors-plus create-author --display_name="John Smith"` (no email/login)
- System finds existing "John Smith" with email "john@example.com"  
- Comparison: `"john@example.com" !== ""` evaluates to `true`
- Result: `$guest_author` is set to null, preventing the match

This means you can't match existing authors by display name alone, even when it should be the same person. The validation should only check fields that were actually provided:

```php
if ( $guest_author ) {
    $mismatch = false;
    
    if ( ! empty( $author['user_email'] ) && $guest_author->user_email !== $author['user_email'] ) {
        $mismatch = true;
    }
    if ( ! empty( $author['user_login'] ) && $guest_author->user_login !== $author['user_login'] ) {
        $mismatch = true;
    }
    
    if ( $mismatch ) {
        $guest_author = null;
    }
}
```

### Minor: Coding standards

Extra spaces before parentheses on lines 1084 and 1088:
- `! empty ( $author` → `! empty( $author`

Happy to help test these changes if needed!

NateWr:
Thanks @GaryJones! I have a small concern about validating only checked fields. In the edge case that you describe, it would be possible to create two authors with the same display name and no other distinguishing data. From a UI perspective, it would be impossible to assign the "correct" author. Is this desirable?

I'm happy to make the change either way. For what it's worth, I think the command would fail in the case that you specify, because `$coauthors_plus->guest_authors->create()` will return an error if it is missing the `user_login` field. I believe the correct solution is to mark `user_login` as a required field or auto-generate it from the `display_name` if it is missing.

GaryJones:
Thanks for the discussion @NateWr - you raise a good point about duplicate authors being indistinguishable.

I think we can simplify this PR by dropping the `display_name` matching entirely. The core fixes needed are:

1. **Single-line synopsis** (required for WP-CLI to parse args)
2. **Empty checks before querying** (prevents false "already exists" matches)

The `display_name` fallback adds complexity and has the bug we discussed. Since `user_login` is effectively required by `create()` anyway, the CLI will fail gracefully with a clear error if it's missing.

**Suggested simplified diff for `create_guest_author()`:**

```php
private function create_guest_author( $author ): void {
    global $coauthors_plus;

    $guest_author = false;

    if ( ! empty( $author['user_email'] ) ) {
        $guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'user_email', $author['user_email'], true );
    }

    if ( ! $guest_author && ! empty( $author['user_login'] ) ) {
        $guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'user_login', $author['user_login'], true );
    }

    if ( $guest_author ) {
        /* translators: Guest Author ID. */
        WP_CLI::warning( sprintf( esc_html__( '-- Author already exists (ID #%s); skipping.', 'co-authors-plus' ), $guest_author->ID ) );
        return;
    }

    // ... rest unchanged
}
```

This keeps the fix focused and avoids the edge cases around display name matching. Thoughts?

I'll push these changes with tests.

NateWr:
Thanks @GaryJones, that sounds good. I've pushed the suggested changes to the PR. Let me know if you need anything else.

GaryJones:
> I've pushed the suggested changes to the PR. 

Thank you! I've added some tests to cover, but all looking good now.

PR #1206, opened 2026-02-20
Fix: allow guest author creation on WordPress 7.0

### LINKED ISSUE #1205: Update new Guest Author creation for WordPress 7.0
Creating a new Guest Author in WordPress 7.0 fails with the messege "“Content, title, and excerpt are empty”

smithjw1:
## Description

Fixes guest author creation on WordPress 7.0 by allowing empty post content for the guest author post type.

Two changes in WordPress core combine to break guest author creation:

1. **[#45516](https://core.trac.wordpress.org/ticket/45516)** (WP 7.0): `get_default_post_to_edit()` now passes an empty `post_title` for CPTs without `title` support, instead of `"Auto Draft"`.
2. **[#37441](https://core.trac.wordpress.org/ticket/37441)** (WP 6.9): `get_default_post_to_edit()` now passes `$wp_error = true` to `wp_insert_post()` and calls `wp_die()` on failure.

The old `filter_wp_insert_post_empty_content()` returned `true` (blocking insertion) when `post_title` was empty. Previously this never triggered because auto-drafts always had `"Auto Draft"` as the title. On WP 7.0, the empty title trips the check and the error is now fatal.

The fix unconditionally returns `false` for the guest author post type, since guest authors store data in post meta and don't need post content. Display name validation is handled separately in `manage_guest_author_filter_post_data()` and the `create()` method's required-field validation.

Fixes #1205

## Deploy Notes

No new dependencies added.

## Steps to Test

1. Check out PR.
2. Run against WordPress trunk (7.0).
3. Navigate to the guest authors admin page.
4. Click "Add New" to create a new guest author.
5. Verify the auto-draft is created successfully and the edit form loads.
6. Fill in the Display Name and other required fields, then save.
7. Verify the guest author is created correctly.
8. Also verify on WP 6.4/6.8 that existing behavior is unchanged.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -115,6 +115,7 @@
 			"version": "7.25.7",
 			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.25.7.tgz",
 			"integrity": "sha512-yJ474Zv3cwiSOO9nXJuqzvwEeM+chDuQ8GJirw+pZ91sCGCyOZ3dJkVE09fTV0VEVzXyLWhh3G/AolYTPX7Mow==",
+			"peer": true,
 			"dependencies": {
 				"@ampproject/remapping": "^2.2.0",
 				"@babel/code-frame": "^7.25.7",
@@ -1936,6 +1937,7 @@
 			"resolved": "https://registry.npmjs.org/keyv/-/keyv-5.5.5.tgz",
 			"integrity": "sha512-FA5LmZVF1VziNc0bIdCSA1IoSVnDCqE8HJIZZv2/W8YmoAM50+tnUgJR/gQZwEeIMleuIOnRnHA/UaZRNeV4iQ==",
 			"dev": true,
+			"peer": true,
 			"dependencies": {
 				"@keyv/serialize": "^1.1.1"
 			}
@@ -2043,6 +2045,7 @@
 					"url": "https://opencollective.com/csstools"
 				}
 			],
+			"peer": true,
 			"engines": {
 				"node": ">=18"
 			},
@@ -2084,6 +2087,7 @@
 					"url": "https://opencollective.com/csstools"
 				}
 			],
+			"peer": true,
 			"engines": {
 				"node": ">=18"
 			}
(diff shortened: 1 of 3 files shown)


GaryJones:
Thanks for this, Jacob. I've pushed a commit adding test coverage for `filter_wp_insert_post_empty_content()` — all tests pass on both WP 6.4 and WP master.

The fix itself is sound: guest authors store their data in post meta, so the empty content guard is unnecessary for this CPT. However, I want to flag something about the framing.

The `$maybe_empty` calculation in `wp_insert_post()` is identical across WP 6.4, 6.7, and trunk:

```php
$maybe_empty = 'attachment' !== $post_type
    && ! $post_content && ! $post_title && ! $post_excerpt
    && post_type_supports( $post_type, 'editor' )
    && post_type_supports( $post_type, 'title' )
    && post_type_supports( $post_type, 'excerpt' );
```

Since the guest author CPT removes `title` and `editor` support, core always passes `$maybe_empty = false` to the filter. This means the old code's `return true` for empty titles has blocked empty-title auto-drafts on **every** WP version, not just trunk.

Could you clarify what changed in WP trunk that exposes this? Is it a change in how the admin UI or REST API creates auto-drafts (e.g. the block editor now creates them with an empty title where the classic editor used "Auto Draft")? Understanding the root cause would help us document this properly in the PR description and ensure we're not missing a wider issue.

The fix is fine to merge either way — I'm just keen to get the "why now" documented accurately.

smithjw1:
Thanks Gary, appreciate the test coverage and the sharp observation.

You're right that the `$maybe_empty` calculation itself hasn't changed. After digging in, the root cause is two changes in `get_default_post_to_edit()`:

1. **[#45516](https://core.trac.wordpress.org/ticket/45516)** (trunk/7.0): Auto-draft titles are now empty for CPTs that don't support `title`:
   ```php
   'post_title' => post_type_supports( $post_type, 'title' ) ? __( 'Auto Draft' ) : '',
   ```
   Since guest authors remove `title` support, the auto-draft now gets `post_title = ''` instead of `"Auto Draft"`.

2. **[#37441](https://core.trac.wordpress.org/ticket/37441)** (6.9): `get_default_post_to_edit()` now passes `$wp_error = true` to `wp_insert_post()` and calls `wp_die()` on failure, instead of silently continuing with `get_post(0)`.

So the old filter's `empty($postarr['post_title'])` check never triggered before because auto-drafts always had `"Auto Draft"` as the title. On 7.0, the empty title trips the check, and the error propagation change makes it fatal.

I'll update the PR description to reflect this.

GaryJones:
Merged, thank you!

Issue #1198, opened 2026-01-20
PHP Warnings during user delete action


leedxw:
We are getting a PHP warning seemingly whenever a user is deleted.

```
PHP message: PHP Warning:  Attempt to read property "term_id" on false in /var/www/html/wp-content/plugins/co-authors-plus/php/class-coauthors-plus.php on line 1127

request: "POST /wp-admin/users.php?action=delete&user=9999&_wpnonce=f34f0b190e HTTP/1.1"
```

PR #1209, opened 2026-03-04
refactor: simplify guest author lookup in delete_user_action()

faisalahammad:
## Summary

Defensive cleanup for the guest author deletion block in `delete_user_action()`.

**Note:** The original PHP warning (`Attempt to read property "term_id" on false`) from #1198 was already fixed on `develop` in commit `9ecae097` with an `if ( $term )` guard. This PR is a related cleanup of a nearby code path.

## Problem

The guest author deletion block in `delete_user_action()` used a separate `get_user_by()` call to retrieve the user being deleted, even though the same data was already available in the `$delete_user` variable fetched earlier in the same method. Additionally, it accessed `$user_data->data->user_login` without validating that `get_user_by()` had returned a valid object.

## Solution

- Reused the existing `$delete_user` variable instead of making a duplicate `get_user_by()` call.
- Added an `if ( $delete_user )` guard before accessing `$delete_user->user_login` to make the intent explicit and prevent potential issues if `$delete_user` is ever false.
- Simplified `$associated_user` access to use `isset( $associated_user->ID )` directly, relying on PHP's `isset()` language construct which safely handles `false`/`null` without warnings.

### php/class-coauthors-plus.php

**Before:**
```php
if ( $this->is_guest_authors_enabled() ) {
    // Get the deleted user data by user id.
    $user_data = get_user_by( 'id', $delete_id );

    // Get the associated user.
    $associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $user_data->data->user_login );

    if ( isset( $associated_user->ID ) ) {
        // Delete associated guest user.
        $this->guest_authors->delete( $associated_user->ID );
    }
}
```

**After:**
```php
if ( $this->is_guest_authors_enabled() ) {
    // Get the associated user from the already fetched $delete_user object.
    if ( $delete_user ) {
        $associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $delete_user->user_login );

        if ( isset( $associated_user->ID ) ) {
            // Delete associated guest user.
            $this->guest_authors->delete( $associated_user->ID );
        }
    }
}
```

**Why:** Eliminates a redundant database call, makes the null-guard intent explicit, and simplifies the `$associated_user` check to rely on PHP's built-in `isset()` safety.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1144,11 +1144,13 @@ class CoAuthors_Plus {
 			$user_data = get_user_by( 'id', $delete_id );
 
 			// Get the associated user.
-			$associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $user_data->data->user_login );
+			if ( $user_data && ! empty( $user_data->data->user_login ) ) {
+				$associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $user_data->data->user_login );
 
-			if ( isset( $associated_user->ID ) ) {
-				// Delete associated guest user.
-				$this->guest_authors->delete( $associated_user->ID );
+				if ( $associated_user && isset( $associated_user->ID ) ) {
+					// Delete associated guest user.
+					$this->guest_authors->delete( $associated_user->ID );
+				}
 			}
 		}
 	}
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1140,12 +1140,9 @@ class CoAuthors_Plus {
 		}
 
 		if ( $this->is_guest_authors_enabled() ) {
-			// Get the deleted user data by user id.
-			$user_data = get_user_by( 'id', $delete_id );
-
-			// Get the associated user.
-			if ( $user_data && ! empty( $user_data->data->user_login ) ) {
-				$associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $user_data->data->user_login );
+			// Get the associated user from the already fetched $delete_user object.
+			if ( $delete_user ) {
+				$associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $delete_user->user_login );
 
 				if ( $associated_user && isset( $associated_user->ID ) ) {
 					// Delete associated guest user.
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1144,7 +1144,7 @@ class CoAuthors_Plus {
 			if ( $delete_user ) {
 				$associated_user = $this->guest_authors->get_guest_author_by( 'linked_account', $delete_user->user_login );
 
-				if ( $associated_user && isset( $associated_user->ID ) ) {
+				if ( isset( $associated_user->ID ) ) {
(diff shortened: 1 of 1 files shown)


GaryJones (commented):
Thanks for the contribution! The defensive null checks are a reasonable improvement, but I have a few concerns:

**1. Does this fix the reported issue?**

Issue #1198 reports `Attempt to read property "term_id" on false` at line 1127, which points to the `wp_delete_term( $term->term_id, ... )` call in the author term deletion block (around current line 1138). This PR fixes the *guest author* block (lines 1144–1151), which is a different code path. Could you confirm which line is actually triggering the warning in your environment?

**2. Duplicate `get_user_by()` call**

Line 1144 calls `get_user_by( 'id', $delete_id )` but the same call already happens on line 1133, stored in `$delete_user`. Since the `delete_user` hook fires *before* the user is removed from the database, both calls return the same object. We could reuse `$delete_user` and avoid the extra database query.

**3. Overly cautious guard**

The `! empty( $user_data->data->user_login )` check isn't needed — if `get_user_by()` returns a valid `WP_User`, `user_login` is always a non-empty string (WordPress requires it). A simple `if ( $delete_user )` would be sufficient and clearer.

faisalahammad:
Thanks for the review, @GaryJones! 

I double-checked the history and the `wp_delete_term( $term->term_id, ... )` warning reported in #1198 seems to have already been fixed by your commit `9ecae097` on the `develop` branch, which added the `if ( $term )` guard. I had mistakenly added defensive checks to the guest author block instead.

However, since simplifying the guest author block is still a good defensive cleanup, I've updated this PR to address your feedback (points 2 and 3). I removed the duplicate `get_user_by()` call, reused the existing `$delete_user` object, and simplified the guard clause.

Let me know if this looks good to merge as a cleanup/enhancement!

GaryJones:
Small note on the `$associated_user && isset( $associated_user->ID )` check: the `$associated_user &&` is technically redundant here — `isset()` is a language construct that suppresses warnings, so `isset( $false_value->ID )` returns `false` without triggering a PHP warning even in PHP 8.x. The existing `isset()` was already safe.

That said, it does make the intent more explicit, so I'm not going to ask you to change it.

GaryJones:
Could you update the PR description to remove the "Fixes #1198" reference? The `term_id` warning from that issue was already fixed on `develop` in commit 9ecae097. Keeping "Fixes #1198" here would auto-close the issue on merge, which would be misleading since this PR addresses a different code path.

Happy to frame this as a related defensive cleanup instead.

faisalahammad:
Thanks for the follow-up notes, @GaryJones!

I've addressed both points:

1. **PR description updated** — Removed the `Fixes #1198` reference and reframed this as a defensive cleanup, since `9ecae097` already handled the original `term_id` warning.

2. **Code simplified** — Removed the redundant `$associated_user &&` prefix from the `isset()` check, letting PHP's built-in `isset()` safety handle `false`/`null` values as you noted.

Happy for this to be merged whenever you're ready!

GaryJones (approved):
Thanks for addressing all the feedback — both the description update and the `isset()` simplification look great. Nice commit message on that last one too.

LGTM.

PR #1217, opened 2026-03-24
Work with real-time collaboration: Replace custom Redux store with core entity store

### LINKED ISSUE #1204: Update for compatibility with real-time collaboration
Real-time collaboration is coming in WordPress 7.0, Co-authors Plus uses local state and communication via a custom Rest API endpoint to update authors. These methods aren't compatible with real-time collaboration.

smithjw1:
## Rationale

Real-time collaboration observes the core data store to sync entities among collaborators. By switching the storage mechanism, Co-Authors Plus becomes compatible with real-time collaboration. This only affects how co-authors are manipulated and saved in the editor. The underlying data shape stays the same.

## Description

Replaces the custom `cap/authors` Redux store and `subscribe()`-based save mechanism with WordPress's core entity store (`getEditedPostAttribute` / `editPost`). Co-author taxonomy term IDs now flow through the standard post entity.

Adds a new `authors-by-term-ids` REST endpoint so the editor sidebar can resolve term IDs to rich author data, a `useCoauthorDetails` hook with request-scoped caching, and PHP-side bridging logic (`sync_coauthors_on_rest_save`, `backfill_coauthor_terms_in_rest`) to keep `post_author` in sync and fire deprecation notices for legacy filters.

Closes #1204

## Deploy Notes

No new dependencies. The `prop-types` import was removed from the JS side. The custom `cap/authors` data store is no longer registered — any third-party code relying on `select('cap/authors')` will need to migrate to the core entity store.

## Steps to Test

1. Check out PR and run `npm run build`.
2. Open the block editor for a post with multiple co-authors — verify authors load in the sidebar panel.
3. Add/remove/reorder co-authors, save the post, and reload — verify changes persist.
4. Create a brand-new post (auto-draft) — verify the current user is assigned as the default co-author.
5. Open a legacy post (created before Co-Authors Plus was active) — verify the `post_author` is backfilled as a co-author term.
6. Verify `post_author` stays in sync with the first co-author after saves.


GaryJones:
I'll defer to @douglas-johnson , @leogermani , @claudiulodro for the review, but a key bit that's missing is the rationale behind this proposed change.

smithjw1:
> ... but a key bit that's missing is the rationale behind this proposed change.

Great point @GaryJones. I've updated the title and added a rationale section. Sorry, I knew why I was doing it, but that isn't clear to everyone looking at the PR.

leogermani:
Thanks @smithjw1 , I think this change makes sense and is a good move. We just need to test it thoroughly and evaluate the extent of the impact on things that we and other people might be relying on... We'll have a closer look!

leogermani:
> Real-time collaboration observes the core data store to sync entities among collaborators. By switching the storage mechanism, Co-Authors Plus becomes compatible with real-time collaboration. 

Can you expand on what exactly doesn't work?

In the current state if multiple people try to change authors at the same time they will override each others changes? Is that what we are solving?

Or is the idea that collaborative editing will automatically assign a person as a co-author when they make edits to it?

smithjw1:
> In the current state if multiple people try to change authors at the same time they will override each others changes? Is that what we are solving?

Yes, that's what happens. Today, if User A and User B are both editing, the "last save wins." With this change, updates made by User B will be immediately reflected to User A.

douglas-johnson:
I tested in a full site editor theme and a hybrid theme with WordPress 7.0-RC2. I did not experience any issues while using the sidebar to add / update co-authors in the post editor or using the co-author blocks in the site editor.

php/class-coauthors-endpoint.php
```diff
@@ -181,6 +201,35 @@ public function update_coauthors( $request ): WP_REST_Response {
...
+
+		foreach ( $ids as $term_id ) {
+			$term = get_term( $term_id, $this->coauthors->coauthor_taxonomy );
+
+			if ( ! $term || is_wp_error( $term ) ) {
+				continue;
+			}
+
+			$slug     = $term->slug;
+			$nicename = str_starts_with( $slug, 'cap-' ) ? substr( $slug, 4 ) : $slug;
```
leogermani: No need to remove the prefix here. CAP already handles it in `$cap->get_user_by()` and `$guest_authors->get_post_meta_key()`

php/class-coauthors-plus.php
```diff
@@ -978,13 +1048,59 @@ public function coauthors_set_post_author_field( $data, $postarr ) {
...
+		);
+
+		if ( empty( $terms ) || is_wp_error( $terms ) ) {
+			return;
+		}
+
+		$coauthor_nicenames = array();
+		foreach ( $terms as $term ) {
+			$slug = $term->slug;
+			$coauthor_nicenames[] = str_starts_with( $slug, 'cap-' ) ? substr( $slug, 4 ) : $slug;
```
leogermani: No need to remove the prefix here. CAP already handles it in `$cap->get_user_by()` and `$guest_authors->get_post_meta_key()`

leogermani (commented):
This looks really good and is a great improvement regardless of collaborative editing. Using term-id in the front end was a good idea to avoid ID collisions between authors and guest authors.

I gave it a full round of testing and it seems to be working fine. I left only two minor comments... I'm still unsure if this can have any bugs with really old guest authors because of what I see [here](https://github.com/Automattic/co-authors-plus/blob/7221a90a1e7b4d6546becd811a639461124623d8/php/class-coauthors-plus.php#L1616). But I think it's fine.

I'll ask some help to review the JS/Store piece of it from more experienced folks on that front.

One thing I know is that this will break any integrations that are relying on the CAP's store for some reason. I know [we are](https://github.com/Automattic/newspack-plugin/pull/3906). But we can fix it.

src/components/co-authors/index.jsx
```diff
@@ -47,38 +37,32 @@ register( coauthorsStore );
...
-	 * CoAuthor select functions.
+	 * Read co-author term IDs from the core entity store.
+	 * Returns undefined until the post entity has loaded, then an array of term IDs.
 	 */
-	const saveAuthors = useSelect(
-		( select ) => select( 'cap/authors' )?.saveAuthors,
-		[]
-	);
+	const { coauthorTermIds, hasResolvedPost } = useSelect( ( select ) => {
+		const { getEditedPostAttribute, hasFinishedResolution, getCurrentPostId } = select( 'core/editor' );
```
chickenn00dle: Nit: It doesn't look like `hasFinishedResolution` is used anywhere?

src/components/co-authors/index.jsx
```diff
@@ -47,38 +37,32 @@ register( coauthorsStore );
...
+		const postId = getCurrentPostId();
+		return {
+			coauthorTermIds: getEditedPostAttribute( 'coauthors' ),
+			hasResolvedPost: !! postId,
```
chickenn00dle: Nit: This is non-blocking because it doesn't impact the functionality of this PR but is checking `postId` enough to verify whether the post has been resolved? I think its still possible for post attributes to be undefined despite this being set.

chickenn00dle (commented):
Just noting I had a look at the JS changes which look good. I left two non-blocking comments. Feel free to ignore.

src/components/co-authors/index.jsx
smithjw1: I think @leogermani commented that it should be.

smithjw1:
I've addressed the PR comments to date, but it would be good to get this out there and tested before releasing a breaking change.

What's the best way to communicate that to plugin users?

leogermani (approved):

smithjw1:
@GaryJones I should have addressed this question to you:

I've addressed the PR comments to date, but it would be good to get this out there and tested before releasing a breaking change.

What's the best way to communicate that to plugin users?

GaryJones:
@smithjw1 We can tag a beta/RC release, and you can send the link to whichever customers you think are willing to test.

smithjw1:
> @smithjw1 We can tag a beta/RC release, and you can send the link to whichever customers you think are willing to test.

Forgive my ignorance, but is merging this to the develop branch the first step in that process? Having never run a release of this plugin, I'm unsure.

GaryJones:
With Gutenberg v23, and two tabs open (Eddie the Editor, and Admin the Administrator), I was able to show that updating of co-authors was reflected in RTC.


https://github.com/user-attachments/assets/d6854068-88ea-4ebb-928d-ee265081ae7c

PR #1224, opened 2026-04-06
Fix: Mark post dirty when co-authors change in the block editor

### LINKED ISSUE #1223: Block editor Update button not enabled when authors are the only change on a published post
## Description

When editing an already-published post in the block editor, changing authors in the Co-Authors Plus sidebar panel does **not** mark the post as dirty. The Update button remains disabled, making it impossible to save the author change unless another edit (content, category, etc.) is also made.

## Steps to Reproduce

1. Open any published post in the block editor
2. In the Co-Authors Plus sidebar panel, change the author(s) (add, remove, swap, or reorder)
3. Observe: the **Update** button remains disabled/grayed out
4. Navigate away — the author change is silently lost

**Expected:** The Update button should enable when authors are changed, same as any other post edit.

**Actual:** The Update button stays disabled. Authors can only be saved by also making an unrelated edit to trigger the save cycle.

## Root Cause

The Co-Authors Plus Gutenberg sidebar manages author state in its own isolated Redux store (`cap/authors` registered in `src/store.js`) and local React `useState`. When authors change, the update is dispatched to `cap/authors` — but **never** to `core/editor` or `core`'s entity record system.

WordPress's block editor determines whether the post is "dirty" (has unsaved changes) via `isEditedPostDirty()`, which checks `hasEditsForEntityRecord("postType", ...)` in the `core` data store. Since CAP only writes to its own store, the core dirty detection never sees any change.

### The save mechanism (`src/index.js`)

The current save approach uses a `subscribe()` piggyback that watches for `isSavingPost()`:

```js
subscribe(() => {
    if (isSavingPost()) {
        checked = false;
    } else if (!checked) {
        const { id } = getCurrentPost();
        const authors = getAuthors(id);
        saveAuthors(id, authors);
        checked = true;
    }
});
```

This only fires when the post is **already being saved** for another reason. If the author change is the only edit, `isSavingPost()` never becomes true, and the author save never executes.

### The core author panel is hidden

`class-coauthors-endpoint.php` removes the `wp:action-assign-author` link from REST responses (lines 291-325), which hides the core Author dropdown in the block editor. The core dropdown **does** use `editPost({ author: ... })` and **would** trigger dirty detection, but CAP replaces it with its own panel that doesn't integrate with the editor's change tracking.

## Proposed Fix

When authors are changed in the sidebar, dispatch a synthetic edit to `core/editor` to mark the post as dirty. For example, in the `updateAuthors` handler in `src/components/co-authors/index.jsx`:

```js
import { dispatch } from '@wordpress/data';

// After updating the cap/authors store:
dispatch('core/editor').editPost({ meta: { _cap_authors_changed: Date.now() } });
```

Or alternatively, use `editPost()` with a non-persistent field that simply signals "something changed" to the dirty detection system.

This ensures:
- The Update button enables when authors change
- The existing `subscribe()` piggyback fires when the user clicks Update
- No changes needed to the save/REST layer

## Affected Versions

This affects all versions of Co-Authors Plus with the block editor sidebar panel, since the sidebar was introduced in 2021. It is not a regression from a specific version — the Gutenberg integration was never connected to the editor's dirty state detection.

Tested on:
- WordPress 6.9.4
- Gutenberg 22.8.2
- Co-Authors Plus 3.7.0

## Related Issues

- #1024 — "Does co-authors plus fire its own HTTP request to store coauthors?" (same architectural concern)
- #919 — "Guest Author Not Saving on Draft Posts" (same root cause — piggyback save not firing)
- WordPress/gutenberg#17632 — "Add hook for successful post save" (referenced in CAP source code comment)
- WordPress/gutenberg#12903 — "How to differentiate between REST and POST requests of Gutenberg inside save_post"

chubes4:
## Summary

When changing authors in the Co-Authors Plus sidebar panel is the **only edit** on a published post, the Update button remains disabled. The editor never detects a change because co-author state is managed in an isolated Redux store (`cap/authors`) that does not notify `core/editor`'s entity record system.

This fix dispatches a synthetic edit to `core/editor` via `editPost()` when authors are changed by the user, marking the post as dirty and enabling the save button.

Fixes #1223

## The Problem

1. User opens a published post in the block editor
2. Changes authors in the Co-Authors Plus sidebar (add, remove, swap, or reorder)
3. The Update button stays disabled — the change cannot be saved
4. Navigating away silently loses the author change

### Root cause

- Co-author changes go to `cap/authors` (custom Redux store) and local `useState`, but **never** to `core/editor` or `core`'s entity record system
- `isEditedPostDirty()` checks `hasEditsForEntityRecord("postType", ...)` — since CAP only writes to its own store, core never sees any change
- The existing `subscribe()` piggyback in `src/index.js` only fires when `isSavingPost()` becomes true, which requires the post to already be dirty from another edit

### Why a synthetic edit is necessary

Core sidebar fields mark the post dirty by passing real data through `editPost()`:
- Author (without CAP): `editPost({ author: userId })`
- Categories: `editPost({ categories: [1, 5, 12] })`
- Tags: `editPost({ tags: [3, 7] })`

CAP cannot do this because its data model doesn't map to core post fields — co-authors are stored in a custom `author` taxonomy via a separate REST endpoint (`/coauthors/v1/authors/`), and guest authors don't have WordPress user IDs. A synthetic edit signal is the minimal way to bridge this gap without restructuring the save architecture.

## The Fix

**One file changed:** `src/components/co-authors/index.jsx`

1. Import `dispatch` from `@wordpress/data` and `useRef` from `@wordpress/element`
2. Add `isHydrated` ref to track whether initial author data has loaded from the API
3. In `updateAuthors`, after updating the CAP store and local state, dispatch `editPost({ coauthors_modified: Date.now() })` to mark the post dirty — but only after hydration is complete
4. Set `isHydrated.current = true` after the first `useEffect` load, so the initial data fetch doesn't falsely mark the post dirty

The `coauthors_modified` key is a non-persistent signal — it marks the entity as edited but is ignored during the REST save since it doesn't correspond to a registered post field.

## Testing

Tested manually on a WordPress 6.9.4 multisite with Gutenberg 22.8.2:

- ✅ Open published post → change author → Update button enables → save persists
- ✅ Open published post → no changes → Update button stays disabled (no false dirty)
- ✅ Add co-author → Update enables
- ✅ Remove co-author → Update enables
- ✅ Reorder co-authors → Update enables
- ✅ Existing save behavior unchanged — `subscribe()` piggyback fires as before

## AI Disclosure

```
AI assistance: Yes
Tool(s): Claude Code
Used for: Codebase analysis to trace the root cause across CAP and
Gutenberg source, drafting the fix, and building. Investigation,
approach, and final review done by the contributor.
```

### DIFF
--- a/src/components/co-authors/index.jsx
+++ b/src/components/co-authors/index.jsx
@@ -10,8 +10,8 @@ import apiFetch from '@wordpress/api-fetch';
 import { ComboboxControl, Spinner } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { applyFilters } from '@wordpress/hooks';
-import { useDispatch, useSelect, register } from '@wordpress/data';
-import { useEffect, useState } from '@wordpress/element';
+import { dispatch as wpDispatch, useDispatch, useSelect, register } from '@wordpress/data';
+import { useEffect, useRef, useState } from '@wordpress/element';
 import { useDebounce } from '@wordpress/compose';
 
 /**
@@ -51,6 +51,7 @@ const CoAuthors = () => {
 	 */
 	const [ selectedAuthors, setSelectedAuthors ] = useState( [] ); // Currently selected options.
 	const [ dropdownOptions, setDropdownOptions ] = useState( [] ); // Options that are available in the dropdown.
+	const isHydrated = useRef( false ); // Tracks whether initial author data has loaded.
 
 	/**
 	 * Retrieve post id.
@@ -89,12 +90,22 @@ const CoAuthors = () => {
 
 	/**
 	 * Setter for updating authors and selected authors simultaneously.
+	 * When the change is user-initiated (after initial hydration),
+	 * also marks the post as dirty so the editor enables the save button.
 	 *
 	 * @param {Array} newAuthors array of new authors.
 	 */
 	const updateAuthors = ( newAuthors ) => {
 		setAuthorsStore( newAuthors );
 		setSelectedAuthors( newAuthors );
+
+		// Mark the post dirty in the core editor store so the
+		// Update/Publish button enables when authors are the only change.
+		if ( isHydrated.current ) {
+			wpDispatch( 'core/editor' ).editPost( {
+				coauthors_modified: Date.now(),
+			} );
+		}
 	};
 
 	/**
@@ -162,6 +173,11 @@ const CoAuthors = () => {
 
 		updateAuthors( authors );
 
+		// Mark hydration complete after the first load so that
+		// subsequent calls to updateAuthors mark the post dirty.
+		if ( ! isHydrated.current ) {
+			isHydrated.current = true;
+		}
 	}, [ authors ] );
 
 	return (


GaryJones:
Hi @chubes4,

This is being handled by #1217, which replaces the custom `cap/authors` Redux store with WordPress's core entity store. Routing author changes through `editPost()` naturally marks the post dirty, so the Update button enables as expected — without needing a bespoke dirty-tracking subscription.

Closing this PR in favour of that approach. Thanks for the patch — it flagged the problem clearly and made the fix easier to scope.

GaryJones:
Closing in favour of #1217 — see updated note above.

PR #1242, opened 2026-04-24
Remove dynamic term backfill on REST API

### LINKED ISSUE #1241: Disappearing co-authors after update to v4
Following an auto-update to v4 from v3.7 about eight hours ago, co-authors disappeared from two recent posts, though not all posts. Rolling back to v3.7 fixed this.

claudiulodro:
## Description

Closes https://github.com/Automattic/co-authors-plus/issues/1241

This PR solves the problem by removing the backfill. IMO a dynamic backfill when reading posts is risky (there isn't a good way to restore any data which has been affected by it, nor does it log when it changes data), and there's already a CLI command that backfills terms, so I don't think it's necessary.

## Deploy Notes

Nope.

## Steps to Test

1. Verify author saving and guest author saving works nicely in general.
2. Follow[ the steps to reproduce the issue here](https://github.com/Automattic/co-authors-plus/issues/1241#issuecomment-4314800938) before and after applying this PR. The issue should be resolved and authors should stay correctly assigned with this PR.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -249,58 +249,9 @@ class CoAuthors_Plus {
 		register_taxonomy( $this->coauthor_taxonomy, $this->supported_post_types(), $args );
 
 		// Bridge REST API saves to add_coauthors() for post_author sync and legacy filter compatibility.
-		// Backfill empty coauthor terms from post_author in REST responses (handles legacy posts).
 		foreach ( $this->supported_post_types() as $post_type ) {
 			add_action( "rest_after_insert_{$post_type}", array( $this, 'sync_coauthors_on_rest_save' ), 10, 2 );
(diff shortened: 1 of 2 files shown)


GaryJones:
Pushed a regression test (b3de169) covering the exact scenario from your reproducer: an authenticated `GET /wp/v2/posts/{id}?_fields=id` against a post with a guest author, asserting the coauthor terms are unchanged after the request. Verified locally that it fails against develop (with the backfill filter) and passes against this branch.

misfist:
Glad to hear this resolved. 

But, is there some reasonable way to restore the authors? I am working with a few sites that have thousands of posts that now have the wrong authors.

GaryJones:
@misfist  We wrote up https://github.com/Automattic/Co-Authors-Plus/wiki/Recovering-from-the-4.0.0-REST-backfill-regression to help solve this. I hope it helps.

PR #1249, opened 2026-04-26
fix: extend author SQL filters to support author__in and comma-separated author IDs

### LINKED ISSUE #1102: Author query parameters not working as expected
I'm seeing some unexpected behavior when I try to use some standard WP_Query parameters with Co-Authors Plus.

For a normal WP_Query with one id specified in the "author" parameter, I see that Co-Authors inserts some additional SQL to query the custom author taxonomy (to make sure that posts are found even if the author is a co-author and not the main author on the post).

However, when I use the "author" parameter with a comma separated list of ids or the "author__in" parameter with an array of ids (as specified here: https://developer.wordpress.org/reference/classes/wp_query/#author-parameters) the query is not updated with a reference to the custom authors taxonomy.

This means that if I am trying to query for posts that have any of multiple authors, only posts that have the main author of the post as the author are returned (not any posts where that author is a coauthor).

I can write my own code to accomplish this if necessary, but it seems like something it would be useful to have in the plugin itself. I tried searching the repository, but didn't find a similar issue. Are there any plans in the works to add this? Or am I missing some setting somewhere to accomplish this?

faisalahammad:
## Summary

`WP_Query`'s `author__in` parameter and the comma-separated form of `author` (e.g. `'author' => '1,2,3'`) do not trigger Co-Authors Plus's taxonomy SQL modifications, so posts where the queried user is a co-author — but not `post_author` — are silently missing from results. The three SQL filter methods (`posts_join_filter`, `posts_where_filter`, `posts_groupby_filter`) all gate on `$query->is_author()`, which is only `true` for single-author URL archives.

Fixes #1102

**Before:** all three SQL filters exit early for non-archive queries
```php
// posts_join_filter, posts_where_filter, posts_groupby_filter
if ( $query->is_author() ) { ... }
```

**After:** a new helper covers both archive and programmatic multi-author queries
```php
if ( $this->is_author_query( $query ) ) { ... }
```

**Why:** `is_author()` is `false` for programmatic queries. The new `is_author_query()` also returns `true` when `author__in` is a non-empty array or `author` contains a comma, so the taxonomy JOIN/WHERE/GROUPBY kicks in for those cases too.

Three methods added:

- **`is_author_query( WP_Query )`** — detects archive and programmatic author queries
- **`get_author_ids_from_query( WP_Query )`** — extracts integer IDs from `author` and `author__in`
- **`posts_where_filter_multi_author( string, WP_Query )`** — resolves each author ID to its CAP taxonomy term and rewrites the `post_author IN (...)` WHERE clause

The existing single-author archive path is unchanged; multi-author queries are dispatched to the new method.

### `tests/Integration/AuthorMultiQueryTest.php`

Six new integration tests:

| Test | Scenario |
|------|----------|
| `test_author_in_with_single_id_finds_post_when_user_is_post_author` | Regression: single `author__in` still works |
| `test_author_in_finds_post_where_user_is_coauthor_not_post_author` | Core bug: `author__in` finds co-authored posts |
| `test_author_in_with_multiple_ids_finds_posts_for_all_coauthors` | Multiple IDs return posts for all authors |
| `test_author_in_does_not_return_unrelated_posts` | No false positives |
| `test_author_comma_string_finds_post_where_user_is_coauthor_not_post_author` | Comma-separated `author` finds co-authored posts |
| `test_single_author_id_still_finds_coauthored_post_after_fix` | Regression: single integer `author` still works |

## Testing

**Test: `author__in` returns co-authored post (the bug)**

1. Create two users — `author1` (ID 2) and `author2` (ID 3)
2. Create a post with `post_author = author1`
3. Use `$coauthors_plus->add_coauthors()` to set both as co-authors
4. Run `new WP_Query(['author__in' => [3]])`

Before fix: 0 posts returned  
After fix: 1 post returned ✅

**Test: comma-separated `author` IDs**

Same setup, run `new WP_Query(['author' => '2,3'])`.

Before fix: only returns the post if author2 is `post_author`  
After fix: returns the post for both co-authors ✅


GaryJones (commented):
Thanks for the thorough write-up and the test scaffolding — the `author__in` half of this looks solid and is a worthwhile fix.

I do have one significant correctness concern, plus a few smaller observations.

## The comma-separated `author` case isn't actually being fixed

Walking the dispatch logic against WordPress core (`wp-includes/class-wp-query.php:971-975`):

```php
if ( empty( $qv['author'] ) || ( '0' == $qv['author'] ) ) {
    $this->is_author = false;
} else {
    $this->is_author = true;
}
```

So for `new WP_Query( [ 'author' => '1,2,3' ] )`, `parse_query()` sets `is_author = true`. WP core then transforms the comma string into `author__in = [1,2,3]` (`class-wp-query.php:2337-2353`) and emits `post_author IN (1,2,3)` into the WHERE clause.

In the new `posts_where_filter` flow:

```php
if ( $this->is_author_query( $query ) ) {
    if ( ! $query->is_author() ) {
        return $this->posts_where_filter_multi_author( $where, $query );
    }
    // existing single-author path…
```

For `author => '1,2,3'`, `is_author()` is **true**, so the new multi-author method is bypassed. The existing path then runs `get_userdata( $query->get( 'author' ) )`, which is `get_userdata( '1,2,3' )` → `false`, and returns `$where` unchanged. No taxonomy expansion happens, and the bug from #1102 remains for this form.

### Why the test for it passes anyway

`test_author_comma_string_finds_post_where_user_is_coauthor_not_post_author` creates a post with `post_author = $author1->ID` and queries `'author' => $author1->ID . ',' . $author2->ID`. WordPress's default `post_author IN (author1, author2)` already matches via `post_author = author1`, so the assertion holds without any CAP modification. The test name suggests it's exercising the co-author-but-not-post-author case, but the post_author is in the comma list, so the bug isn't being triggered.

To actually expose the bug, you would want something like:

```php
$other  = $this->create_author( 'comma_other' );
$post   = $this->create_post( $other ); // post_author = $other (NOT in the comma list)
$this->_cap->add_coauthors( $post->ID, [ $other->user_login, $author2->user_login ] );

$query = new \WP_Query( [
    'author' => $author2->ID . ',' . $unrelated->ID,
] );
$this->assertCount( 1, $query->posts );
```

That test would fail against the current PR.

### A possible direction

Since WP populates `author__in` from the comma string by the time `posts_where_filter` runs, you could dispatch to the multi-author method whenever `get_author_ids_from_query()` returns more than one ID, irrespective of `is_author()`. That also eliminates the dead branch in `is_author_query()` (see below).

## Dead code in `is_author_query()`

Because WP sets `is_author = true` whenever `qv['author']` is non-empty (whether `5`, `'5'`, or `'1,2,3'`), the first `if ( $query->is_author() ) return true;` already covers every comma-separated case. The `str_contains( $author_var, ',' )` branch is unreachable in practice. Worth removing or — better — moving the comma logic into the dispatch so it actually does something.

## Other observations

- **Duplicated taxonomy-clause assembly.** Term lookup, linked-account fallback, the `coauthors_plus_should_query_post_author` filter, and the `force_guest_authors` short-circuit are now in two places. A small private helper that builds the taxonomy clause from a list of co-authors would let both paths share it and reduce drift.
- **Regex character class.** `[\d ,]+` is loose enough to accept `,, ,` or stray whitespace. WordPress's own SQL won't produce that, but tightening to something like `\(?\d+(?:\s*,\s*\d+)*\)?` is safer and self-documenting.
- **Performance.** `posts_where_filter_multi_author` does `get_userdata` + `get_coauthor_by` + `get_author_term` per requested ID (and again for any linked account). Fine at typical sizes; worth a brief comment if you expect this on large arrays.
- **Style nit.** `//end foreach` is a Squiz convention; the rest of the file doesn't use it. Worth running phpcs to confirm.
- **Drive-by Yoda flip.** The change at line 891 (`$id !== '\d+'` → `'\d+' !== $id`) is unrelated to the PR's scope.
- **`@since 3.6.0`.** The plugin header is `Version: 4.0.0`, so the docblock tag may want updating to whatever the next release will be.

## What's good

- Clean separation of detection, extraction, and rewriting.
- Type hints and reasonable docblocks throughout.
- Linked guest-author handling is consistent with the existing path.
- Regression coverage for the single-ID forms guards the legacy path.
- The `author__in` case is genuinely fixed and well covered.

Happy to look again once the comma-separated case has been addressed (and its test updated to fail without the fix). Thanks again for tackling this one.

faisalahammad:
@GaryJones Thanks for the detailed review! I have addressed your feedback:

1. Fixed the dispatch logic so the comma separated author case is correctly routed to the multi author path.
2. Replaced the flawed comma test with one that genuinely fails without the fix (post author is not in the queried list).
3. Removed the dead comma branch in is_author_query().
4. Tightened the regex in posts_where_filter_multi_author().
5. Removed the Squiz style comment.
6. Updated the since tags to 4.0.0.
7. Reverted the unrelated Yoda flip.

All tests are passing across the different WP versions. Let me know if you have any other feedback!

GaryJones (commented):
Thanks for the quick turnaround — this is a much tighter revision.

## What's now correct

The new dispatch in `posts_where_filter` resolves the main concern:

```php
$author_ids = $this->get_author_ids_from_query( $query );
if ( ! empty( $author_ids ) && ( ! $query->is_author() || count( $author_ids ) > 1 ) ) {
    return $this->posts_where_filter_multi_author( $where, $query );
}
```

Tracing the cases:

- **`author__in => [3]`** (programmatic, single ID) — `is_author = false`, `! is_author()` triggers dispatch. ✓
- **`author__in => [3, 4]`** (programmatic, multi-ID) — same path. ✓
- **`author => '3,4'`** — WP expands to `author__in = [3, 4]`, sets `is_author = true`. `count > 1` triggers dispatch. ✓
- **`/author/john/` archive** — `is_author = true`, `author_name` set, `author` resolved to a single integer, no comma. `get_author_ids_from_query` returns `[]` (the integer fails the `is_string` check and `author__in` is empty). Falls through to the existing single-author path. ✓
- **`author => 5`** (programmatic, single ID) — WP populates `author__in = [5]`, `is_author = true`. `count == 1` and `is_author()` true → falls through to existing path, which handles single IDs correctly. ✓

The strengthened comma-separated test now genuinely fails without the fix: `post_author = $other` is excluded from the comma list, so the only way the post is returned is via the taxonomy rewrite. Good.

The other points are addressed too:

- Dead `str_contains( ',' )` branch removed from `is_author_query()`, with a docblock note explaining where the comma case is now handled.
- Regex tightened to `\(?\b(?:…)?post_author\s*(?:=\s*\d+|IN\s*\(\d+(?:\s*,\s*\d+)*\))\)?`.
- `@since` updated to `4.0.0`.
- The unrelated Yoda flip on the `$id !== '\d+'` line has been reverted.
- `//end foreach` style comment removed.

## Remaining (non-blocking)

- The taxonomy-clause assembly is still duplicated between the existing single-author block and the new method. Worth a follow-up to extract a small helper, but happy for that to be a separate PR.
- A brief comment on `posts_where_filter_multi_author` noting the per-ID lookup cost (and that it's bounded by typical `author__in` sizes) would help future readers.

Looks good to me from a logic standpoint — leaving the final approval to the maintainers, but I'm content this addresses #1102 properly now.

faisalahammad:
Thanks for the thorough second review, @GaryJones — glad the dispatch logic checks out.

Addressed the remaining item in 3197ee4:

**Performance note** — added a docblock comment on `posts_where_filter_multi_author` documenting the per-ID lookup cost and why it's acceptable (small `author__in` arrays, WP object cache).

**Taxonomy-clause deduplication** — agreed, worth extracting a shared helper. I'll open a follow-up PR for that to keep this one focused on the #1102 fix.

GaryJones:
Excellent, thank you!

PR #1250, opened 2026-04-26
fix: clear conflicting WP_Query flags on guest author pages

### LINKED ISSUE #1109: Warnings when author page somehow treated as category
Note that this issue relates to warnings generated via abnormal and unexpected requests.

We had a number of PHP warnings following someone using a vulnerability scanner against a site we run.

```
PHP Warning:  Undefined property: stdClass::$name in /var/www/html/wp-includes/general-template.php on line 1610
PHP Deprecated:  strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/wp-includes/class-wp-hook.php on line 324
PHP Warning:  Undefined property: stdClass::$name in /var/www/html/wp-includes/general-template.php on line 3284
PHP Warning:  Undefined property: stdClass::$term_id in /var/www/html/wp-includes/general-template.php on line 3287
```

I could eventually reproduce the issue by appending `?x=&cat=1` to an author page for a guest author (using co-authors-plus) but not a system author. I suspect the `cat=` argument is causing `is_category()` to return true, but I can't see where the logic issue is, so I don't know if this is a bug in core or the plugin.

````
http://localhost/author/guest/?x=&cat=1
````

Triggering a stack dump:

```
#0 /var/www/html/wp-includes/general-template.php(1610):  errHandle()
#1 /var/www/html/wp-includes/general-template.php(1229): single_term_title()
#2 /var/www/html/wp-includes/general-template.php(1311): wp_get_document_title()
#3 /var/www/html/wp-includes/class-wp-hook.php(324): _wp_render_title_tag()
#4 /var/www/html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
#5 /var/www/html/wp-includes/plugin.php(517): WP_Hook->do_action()
#6 /var/www/html/wp-includes/general-template.php(3064): do_action()
#7 /var/www/html/wp-content/themes/example/layouts/main.php(18): wp_head()
#8 /var/www/html/wp-includes/template-loader.php(106): include('...')
#9 /var/www/html/wp-blog-header.php(19): require_once('...')
#10 /var/www/html/index.php(17): require('...')
```

### PR DESCRIPTION
**Pull request by faisalahammad**
## Summary

When a guest author URL is visited with unexpected query parameters (e.g. `?cat=1`, common from vulnerability scanners), WordPress sets both `is_category` and `is_author` to `true` simultaneously. `fix_author_page()` correctly overwrites `queried_object` with the guest author `stdClass`, but previously left the conflicting `is_category` (and related) flags intact — causing core functions like `single_term_title()` to access `queried_object->name` and `->term_id`, which don't exist on a guest author object.

This only affects guest authors, not standard WP users.

Fixes #1109

### `php/class-coauthors-plus.php` — `fix_author_page()`

**Before:**
```php
if ( is_object( $authordata ) || ! empty( $term ) ) {
    $wp_query->queried_object    = $authordata;
    $wp_query->queried_object_id = (int) $authordata->ID;
    if ( ! is_paged() ) {
        add_filter( 'pre_handle_404', '__return_true' );
    }
}
```

**After:**
```php
if ( is_object( $authordata ) || ! empty( $term ) ) {
    $wp_query->queried_object    = $authordata;
    $wp_query->queried_object_id = (int) $authordata->ID;

    // Ensure conflicting query flags are cleared. When unexpected query vars
    // (e.g. ?cat=1) arrive alongside an author URL, WordPress may simultaneously
    // set is_category (or other flags) that conflict with is_author. This causes
    // core functions like single_term_title() to read properties that don't exist
    // on the guest-author stdClass object, triggering PHP warnings.
    // See https://github.com/Automattic/co-authors-plus/issues/1109.
    $wp_query->is_category = false;
    $wp_query->is_tag      = false;
    $wp_query->is_tax      = false;
    $wp_query->is_singular = false;
    $wp_query->is_single   = false;
    $wp_query->is_page     = false;
    $wp_query->is_home     = false;
    $wp_query->is_search   = false;
    $wp_query->is_feed     = false;

    if ( ! is_paged() ) {
        add_filter( 'pre_handle_404', '__return_true' );
    }
}
```

**Why:** Once `fix_author_page()` takes ownership of `queried_object`, the request is definitively an author archive — any other conditional flags left over from the initial query parse are wrong. Resetting them here prevents downstream consumers (core or otherwise) from acting on contradictory state.

### `tests/Integration/AuthorQueriedObjectTest.php`

Added `test__guest_author_page_with_cat_query_var_does_not_set_is_category()`: creates a guest author, simulates visiting their URL with `?cat=1`, and asserts `is_category()` / `is_tag()` / `is_tax()` are `false` while `is_author()` and `is_archive()` remain `true`.

## Testing

**Test 1: Guest author page with `?cat=1` query param**
1. Install the plugin with guest authors enabled
2. Create a guest author (e.g. slug `testguest`)
3. Visit `/author/testguest/?cat=1`
4. Check the PHP error log

Result: No `Undefined property: stdClass::$name` warnings; page renders as the author archive.

**Test 2: Normal guest author page (regression)**
1. Visit `/author/testguest/` (no extra params)

Result: Page still renders correctly as the author archive.

**Test 3: Automated**
```bash
composer test:integration -- --filter AuthorQueriedObjectTest
```

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1440,21 +1440,16 @@ class CoAuthors_Plus {
 			$wp_query->queried_object    = $authordata;
 			$wp_query->queried_object_id = (int) $authordata->ID;
 
-			// Ensure conflicting query flags are cleared. When unexpected query vars
-			// (e.g. ?cat=1) arrive alongside an author URL, WordPress may simultaneously
-			// set is_category (or other flags) that conflict with is_author. This causes
-			// core functions like single_term_title() to read properties that don't exist
-			// on the guest-author stdClass object, triggering PHP warnings.
+			// Once fix_author_page() takes ownership of queried_object, this request
+			// is definitively an author archive. Reset all query flags to a clean
+			// state — mirroring how core handles flag transitions internally — then
+			// re-assert only the two flags that are true. This prevents downstream
+			// consumers from acting on contradictory state left over from unexpected
+			// query vars (e.g. ?cat=1 setting is_category=true alongside is_author=true).
 			// See https://github.com/Automattic/co-authors-plus/issues/1109.
-			$wp_query->is_category = false;
-			$wp_query->is_tag      = false;
-			$wp_query->is_tax      = false;
-			$wp_query->is_singular = false;
-			$wp_query->is_single   = false;
-			$wp_query->is_page     = false;
-			$wp_query->is_home     = false;
-			$wp_query->is_search   = false;
-			$wp_query->is_feed     = false;
+			$wp_query->init_query_flags();
+			$wp_query->is_author  = true;
+			$wp_query->is_archive = true;
 
 			if ( ! is_paged() ) {
 				add_filter( 'pre_handle_404', '__return_true' );
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1443,13 +1443,14 @@ class CoAuthors_Plus {
 			// Once fix_author_page() takes ownership of queried_object, this request
 			// is definitively an author archive. Reset all query flags to a clean
 			// state — mirroring how core handles flag transitions internally — then
-			// re-assert only the two flags that are true. This prevents downstream
-			// consumers from acting on contradictory state left over from unexpected
-			// query vars (e.g. ?cat=1 setting is_category=true alongside is_author=true).
+			// re-assert only the flags that should remain true. Preserve is_paged so
+			// that paginated author archives can still trigger 404 when out of range.
 			// See https://github.com/Automattic/co-authors-plus/issues/1109.
+			$is_paged = $wp_query->is_paged;
 			$wp_query->init_query_flags();
 			$wp_query->is_author  = true;
 			$wp_query->is_archive = true;
+			$wp_query->is_paged   = $is_paged;
 
 			if ( ! is_paged() ) {
 				add_filter( 'pre_handle_404', '__return_true' );
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1439,6 +1439,23 @@ class CoAuthors_Plus {
 		if ( is_object( $authordata ) || ! empty( $term ) ) {
 			$wp_query->queried_object    = $authordata;
 			$wp_query->queried_object_id = (int) $authordata->ID;
+
+			// Ensure conflicting query flags are cleared. When unexpected query vars
+			// (e.g. ?cat=1) arrive alongside an author URL, WordPress may simultaneously
+			// set is_category (or other flags) that conflict with is_author. This causes
+			// core functions like single_term_title() to read properties that don't exist
+			// on the guest-author stdClass object, triggering PHP warnings.
+			// See https://github.com/Automattic/co-authors-plus/issues/1109.
+			$wp_query->is_category = false;
+			$wp_query->is_tag      = false;
+			$wp_query->is_tax      = false;
+			$wp_query->is_singular = false;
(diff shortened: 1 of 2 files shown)


GaryJones:
Thanks for tackling this, @faisalahammad. The diagnosis is spot on, the inline comment really helps explain the *why*, and the link back to #1109 makes the trail easy to follow for future maintainers. A few thoughts that might strengthen the fix further:

### Consider broadening the flag reset

The nine flags you've cleared cover the exact symptom from the issue (`?cat=1`), but the same underlying bug is reachable through other query vars. For example:

- `?year=2024` would set `is_year` and `is_date`
- `?attachment_id=123` would set `is_attachment`
- `?post_type=foo` could set `is_post_type_archive`

Each of these would leave `fix_author_page()` in the same contradictory state, just with a different flag. Two ways forward:

1. **Expand the explicit list** to include `is_attachment`, `is_post_type_archive`, `is_year`, `is_month`, `is_day`, `is_time`, and `is_date`.
2. **Use `init_query_flags()`** then set only `is_author = true` and `is_archive = true`. This mirrors how core resets state internally and means we don't have to chase new flags as WordPress evolves. Slightly more invasive, but more robust to future changes.

Option 2 is my preference, since the invariant we actually want is "this is an author archive, full stop", and explicitly listing every false flag is the long way around.

### Test reproducibility

I noticed the test comment mentions that `go_to()` with `?cat=1` doesn't reproduce the conflicting state reliably, so the state is injected manually. That's pragmatic, but it does mean the test verifies the post-fix flag state rather than the actual user journey. A couple of options worth considering:

- Add a second test that calls `$wp->parse_request()` with a synthesised request URI to get closer to the real path, even if it's a bit more setup.
- Add an assertion that no PHP warning is raised when `single_term_title()` runs against the queried object. That captures the original symptom from #1109 directly, so the test stays meaningful even if WordPress changes how it parses these query strings in the future.

Both are nice-to-haves rather than blockers, given the value of catching the regression at all.

### Tiny nit

The phrase "is_category (or other flags)" in the inline comment is slightly redundant given the list directly below it. You could trim that sentence without losing any context.

Overall a solid, well-scoped fix with a clear test. The core change is in the right place and the reasoning is sound; the suggestions above are about future-proofing rather than correctness.

faisalahammad:
Thanks for the thorough review, @GaryJones — all three points addressed in 8d3ba47.

**Flag reset** — switched to `init_query_flags()` then re-assert `is_author = true` and `is_archive = true`. Agree this is the right invariant: "this is an author archive, full stop", and it's robust to any future flags WordPress adds.

**Test** — expanded the assertions to cover the full relevant set (`is_year`, `is_date`, `is_attachment`, `is_post_type_archive`, `is_single`, `is_singular`, `is_page` in addition to the original three) and added a `single_term_title( '', false )` assertion that directly captures the original #1109 symptom. Skipped the `parse_request()` second test for now given environment fragility.

**Comment nit** — trimmed the redundant phrase.

GaryJones:
Lovely work, @faisalahammad. The switch to `init_query_flags()` is exactly what I had in mind, and the catch on is_paged is excellent — that would have been a subtle regression for paginated author archives, and you spotted it independently. The strengthened test, particularly the `single_term_title()` assertion, ties the verification right back to the original symptom.

faisalahammad:
Thank you very much @GaryJones for the review and your kind words. I am happy that you liked the solution with `init_query_flags` for this fix. I am also glad that I found the `is_paged` issue so we do not have a regression for paginated archives. It is good that the test is now stronger with `single_term_title`. Thank you for your help to make this better.

PR #1251, opened 2026-04-27
refactor: extract shared taxonomy clause helpers

faisalahammad:
## Summary

Extract shared helper methods from the duplicated taxonomy-clause assembly logic in `posts_where_filter()` and `posts_where_filter_multi_author()`. Pure refactoring — no behavior change.

Follow-up to #1249 as [discussed with @GaryJones](https://github.com/Automattic/co-authors-plus/pull/1249#pullrequestreview-4180220165).

> [!IMPORTANT]
> **Stacked on #1249.** Only the last commit (`e84770a`) belongs to this PR. The other commits are from #1249 and will disappear from this diff once #1249 merges to `develop`. To review just this PR's changes: [view commit e84770a](https://github.com/Automattic/co-authors-plus/pull/1251/commits/e84770a).

Two new `protected` helpers extracted from duplicated code:

**`collect_coauthor_terms( $coauthor ): WP_Term[]`**

Resolves a coauthor's primary term and linked account term into an array. Previously inlined in both the single-author and multi-author paths.

**`build_terms_clauses( WP_Term[] $terms ): string`**

Builds the OR-chain WHERE clause fragment and populates `$this->having_terms` for the GROUPBY HAVING clause. Returns `terms_implode` trimmed; leaves `having_terms` untrimmed so the single-author path can append private post terms before the final `rtrim()`.

**Before:** term resolution + SQL assembly duplicated across two methods (~40 lines each).

**After:** both paths delegate to the shared helpers:

```php
// Single-author path
$coauthor = $this->get_coauthor_by( 'user_nicename', $author_name );
$terms    = $this->collect_coauthor_terms( $coauthor );
// ...
$terms_implode = $this->build_terms_clauses( $terms );

// Multi-author path
foreach ( $author_ids as $author_id ) {
    // ...
    $all_terms = array_merge( $all_terms, $this->collect_coauthor_terms( $coauthor ) );
}
$terms_implode = $this->build_terms_clauses( $all_terms );
```

**Why:** Eliminates structural duplication so future changes to term resolution or SQL construction only need to happen in one place, reducing drift between the two query paths.

## Testing

No new tests — this is a behavior-preserving refactoring. All existing integration tests (single-site + multisite) validate correctness.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -939,18 +939,8 @@ class CoAuthors_Plus {
 				}
 			}
 
-			$terms    = array();
 			$coauthor = $this->get_coauthor_by( 'user_nicename', $author_name );
-			if ( $author_term = $this->get_author_term( $coauthor ) ) {
-				$terms[] = $author_term;
-			}
-			// If this co-author has a linked account, we also need to get posts with those terms
-			if ( ! empty( $coauthor->linked_account ) ) {
-				$linked_account = get_user_by( 'login', $coauthor->linked_account );
-				if ( $guest_author_term = $this->get_author_term( $linked_account ) ) {
-					$terms[] = $guest_author_term;
-				}
-			}
+			$terms    = $this->collect_coauthor_terms( $coauthor );
 
 			// Whether to include the original 'post_author' value in the query.
 			// Don't include it if we're forcing guest authors, or it's obvious our query is for a guest author's posts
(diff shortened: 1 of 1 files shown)


GaryJones:
Quick note now that #1249 has merged: would you mind rebasing this onto `develop` and force-pushing? That'll drop the four already-merged commits from the diff and leave just `e84770a`, which makes the review noticeably easier. Thanks again for breaking the refactor out into its own PR.

faisalahammad:
Done! Rebased onto `develop` and force-pushed. The diff now shows only the single refactor commit (`81f5446`). Thanks for the quick review and for merging #1249 so cleanly.

GaryJones (approved):
Looks good. Behaviour is preserved, the helpers are sensibly scoped, and the asymmetric trim contract on `build_terms_clauses` is clearly documented. CI is green across all four matrices, and the rebase has the diff down to just the refactor commit.

Two non-blocking nits for a future tidy-up if you fancy:

- `build_terms_clauses` returns `terms_implode` trimmed but leaves `$this->having_terms` untrimmed for the caller to finalise. It's documented, but a slightly "leaky" abstraction — returning both strings (or an array) would make the contract self-evident.
- A small `if ( ! $coauthor ) return array();` guard in `collect_coauthor_terms` would harden the linked-account access against null inputs. Pre-existing risk, not a regression.

Thanks for splitting the refactor out into its own PR — made this very easy to review.

GaryJones:
I went ahead and merged, as we're doing a new release.

PR #1261, opened 2026-04-30
fix(search): raise search_authors() result cap and gate back-fill by capability

### LINKED ISSUE #1057: search_authors() function doesn't return all results, if there are more than 10 results
When you have a combined number of more than 10 WP Users + Guest Authors that return for the same search string, not all of the results are displayed for the user to choose from.  (I.E. 9 WP users  and 4 Guest Authors all with "Max" somewhere in the name) 

The fix is:
```
add_filter( 'coauthors_search_authors_get_terms_args', 'how_coauthors_update_term_args', 10, 1 );
function how_coauthors_update_term_args( $args ) {
	$args['number'] = 0;
	return $args;
}
```

Which modifies the `$args` used by `get_terms()` here: https://github.com/Automattic/Co-Authors-Plus/blob/37065d289a9da07aa5228aa4969c8f96576a58c0/php/class-coauthors-plus.php#L1343

So my question is, [why are we specifying](https://github.com/Automattic/Co-Authors-Plus/blob/37065d289a9da07aa5228aa4969c8f96576a58c0/php/class-coauthors-plus.php#L1339C4-L1339C19) `'number' => 10,` in the `$args` in the first place? And if there's no reason, then can we remove that? If there is a reason, can we nevertheless have all results returned as expected?

How to reproduce:
Create 6 WP users (editor role) all with first name "max" 
Create 5 Guest Authors all with the first name "max"

Either use the UI (search in the Authors metabox) or use the endpoint: `/wp-json/coauthors/v1/search/?max`

However, you need to make the endpoint public first with:
```
function make_coauthors_search_route_public( $endpoints ) {
    if ( isset( $endpoints['/coauthors/v1/search'] ) ) {
        // Set the permission callback to always return true, making the endpoint public
        $endpoints['/coauthors/v1/search'][0]['permission_callback'] = '__return_true';
    }
    return $endpoints;
}
add_filter( 'rest_endpoints', 'make_coauthors_search_route_public' );
```
Either way, you should observe that not all 11 results are displayed. At least that's what I saw on multiple sites.

faisalahammad:
## Summary

`search_authors()` uses a two-phase approach: first `get_users()` back-fills author taxonomy terms for matching WP users, then `get_terms()` searches those term descriptions. Two separate bugs meant search results were silently incomplete on many sites.

Fixes #1057

**Change 1 — Raise the 10-result cap to 50 in `get_terms()`**

**Before:**
```php
$args = array(
    'search' => $search,
    'get'    => 'all',
    'number' => 10,
);
```

**After:**
```php
$args = array(
    'search' => $search,
    'get'    => 'all',
    'number' => 50,
);
```

**Why:** A hard cap of 10 silently truncated autocomplete results whenever more than 10 authors matched. 50 keeps the query bounded while fixing the common case of 11+ matches. Sites can still adjust the cap via `coauthors_search_authors_get_terms_args`.

---

**Change 2 — Keep the `capability` constraint on the back-fill `get_users()` query, gate term creation by `has_cap()`**

**Before (this PR's earlier revision):**
```php
$cap = apply_filters( 'coauthors_edit_author_cap', 'edit_posts' );
$args = array(
    'count_total'    => false,
    'search'         => sprintf( '*%s*', $search ),
    'search_columns' => array( 'ID', 'display_name', 'user_email', 'user_login' ),
    'capability'     => array( $cap ),
    'role__in'       => $this->get_roles_with_cap( $cap ),
    'fields'         => 'all_with_meta',
);
$args = apply_filters( 'coauthors_search_authors_get_users_args', $args );
$found_users = get_users( $args );
```

**After:**
```php
$cap = apply_filters( 'coauthors_edit_author_cap', 'edit_posts' );
$args = array(
    'count_total'    => false,
    'search'         => sprintf( '*%s*', $search ),
    'search_columns' => array( 'ID', 'display_name', 'user_email', 'user_login' ),
    'capability'     => array( $cap ),
    'fields'         => 'all_with_meta',
);
$args = apply_filters( 'coauthors_search_authors_get_users_args', $args );
$found_users = get_users( $args );

foreach ( $found_users as $found_user ) {
    if ( ! $found_user->has_cap( $cap ) ) {
        continue;
    }
    // ...back-fill as before
}
```

**Why:** `role__in` (derived from the capability via a new `get_roles_with_cap()` helper) was added in an earlier revision to make the back-fill query more reliable on multisite/wp-env. Tracing `WP_User_Query::prepare_query()` and reproducing on wp-env (WP 7.1) showed this doesn't hold up: `capability` already expands internally to match every role that grants it, so `role__in` selects the same users on the happy path. The two arguments combine with AND, so `role__in` can never rescue a case where `capability` returns empty, only narrow it — and it does narrow it: a user granted the capability directly (not through a role) passes `capability` but fails `role__in`, and was incorrectly dropped from back-fill.

`role__in` and `get_roles_with_cap()` are removed. What stays: `capability` on the back-fill query, a `has_cap( $cap )` check in the back-fill loop before any term is created or updated (so a low-capability user can't get a term back-filled even if one somehow surfaced from `get_users()`), and the `coauthors_search_authors_get_users_args` filter so sites can further customize or short-circuit the back-fill query.

## Deploy Notes

No new dependencies. This is a pure PHP behavior change to `CoAuthors_Plus::search_authors()`:
- The `get_terms()` result limit for author-search autocomplete rises from 10 to 50. Both `coauthors_search_authors_get_terms_args` (existing) and `coauthors_search_authors_get_users_args` (new) remain available for sites that need different limits or a different back-fill query.
- The back-fill `get_users()` call runs a `has_cap()` check per user before creating or updating a term — this doesn't change term-creation behavior versus current `develop`, it's a safeguard against a side effect a low-capability user's term back-fill would otherwise leave behind.
- Sites with very large low-privilege user bases that find the back-fill query slow can use `coauthors_search_authors_get_users_args` to further constrain or short-circuit it.

## Testing

Four integration tests cover the new behavior:

1. `test_search_authors_returns_all_results_beyond_default_limit` — 13 matching authors (7 WP + 6 guest) are all returned.
2. `test_search_authors_excludes_low_cap_users_from_keyword_search` — a subscriber sharing a search prefix with an editor is excluded.
3. `test_search_authors_does_not_backfill_terms_for_low_cap_users` — subscribers do not get author terms created during search.
4. `test_search_authors_includes_user_with_directly_granted_capability` — a user granted the capability directly (not through a role) still appears in results, guarding against the regression `role__in` introduced.

Manual tests:

1. Create 7+ editors and 6+ guest authors with `maxtest` in their display name, then search `maxtest` in the Authors panel. All 13+ should appear.
2. Create a subscriber and an editor with `capcheck` in their display name, search `capcheck`. Only the editor should appear.
3. Create a subscriber with `termgrowth` in their display name, search `termgrowth`, and confirm no author term is created for the subscriber.

## Build

[co-authors-plus-fix-1057-v2.zip](https://github.com/user-attachments/files/placeholder.zip) available for manual testing.

Install via _WP Admin → Plugins → Upload Plugin_.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -2078,6 +2078,7 @@ class CoAuthors_Plus {
 		// instead of the user details. If the term is missing, we probably need to
 		// back-fill with user details. Let's do this first... easier than running
 		// an upgrade script that could break on a lot of users
+		$cap         = apply_filters( 'coauthors_edit_author_cap', 'edit_posts' );
 		$args        = array(
 			'count_total'    => false,
 			'search'         => sprintf( '*%s*', $search ),
@@ -2087,22 +2088,31 @@ class CoAuthors_Plus {
 				'user_email',
 				'user_login',
 			),
-			'capability'     => array( apply_filters( 'coauthors_edit_author_cap', 'edit_posts' ) ),
+			'capability'     => array( $cap ),
 			'fields'         => 'all_with_meta',
 		);
+		$args        = apply_filters( 'coauthors_search_authors_get_users_args', $args );
 		$found_users = get_users( $args );
 
 		foreach ( $found_users as $found_user ) {
+			// get_users()'s 'capability' arg bounds the query but isn't the
+			// correctness guarantee; skip back-filling a term for anyone who
+			// doesn't actually hold the capability.
+			if ( ! $found_user->has_cap( $cap ) ) {
+				continue;
+			}
 			$term = $this->get_author_term( $found_user );
 			if ( empty( $term ) || empty( $term->description ) ) {
 				$this->update_author_term( $found_user );
 			}
 		}
 
+		// Raised from 10 to 50 (#1057) so large sites' autocomplete stops silently
+		// truncating matches. Adjust further via the filter below.
 		$args = array(
 			'search' => $search,
 			'get'    => 'all',
-			'number' => 10,
+			'number' => 50,
 		);
 		$args = apply_filters( 'coauthors_search_authors_get_terms_args', $args );
 		add_filter( 'terms_clauses', array( $this, 'filter_terms_clauses' ) );
--- a/tests/Integration/Bylines/SearchAuthorsTest.php
+++ b/tests/Integration/Bylines/SearchAuthorsTest.php
@@ -171,4 +171,101 @@ class SearchAuthorsTest extends TestCase {
 		$this->assertArrayNotHasKey( $this->author1->user_login, $authors );
 		$this->assertArrayHasKey( $author2->user_login, $authors );
 	}
+
+	/**
+	 * Checks search_authors() returns all matches when more than 10 share a prefix.
+	 *
+	 * @covers ::search_authors
+	 */
+	public function test_search_authors_returns_all_results_beyond_default_limit(): void {
+
+		global $coauthors_plus;
+
+		$prefix  = 'maxtest';
+		$editors = array();
+
+		for ( $i = 1; $i <= 7; $i++ ) {
+			$editors[] = $this->create_editor( "{$prefix}_editor_{$i}" );
+		}
+
+		$guest_logins = array();
+		for ( $i = 1; $i <= 6; $i++ ) {
+			$guest_logins[] = "{$prefix}_guest_{$i}";
+			$this->create_guest_author( "{$prefix}_guest_{$i}" );
+		}
+
+		$authors = $coauthors_plus->search_authors( $prefix );
+
+		$this->assertGreaterThanOrEqual( 13, count( $authors ) );
+
+		foreach ( $editors as $editor ) {
+			$this->assertArrayHasKey( $editor->user_login, $authors );
+		}
+		foreach ( $guest_logins as $guest_login ) {
+			$this->assertArrayHasKey( $guest_login, $authors );
+		}
+	}
+
+	/**
+	 * Checks search_authors() excludes a subscriber sharing a prefix with an editor.
+	 *
+	 * @covers ::search_authors
+	 */
+	public function test_search_authors_excludes_low_cap_users_from_keyword_search(): void {
+
+		global $coauthors_plus;
+
+		$prefix     = 'capcheck';
+		$editor     = $this->create_editor( "{$prefix}_editor" );
+		$subscriber = $this->create_subscriber( "{$prefix}_subscriber" );
+
+		$authors = $coauthors_plus->search_authors( $prefix );
+
+		$this->assertArrayHasKey( $editor->user_login, $authors );
+		$this->assertArrayNotHasKey( $subscriber->user_login, $authors );
+	}
+
+	/**
+	 * Checks search_authors() does not create an author term for a subscriber.
+	 *
+	 * @covers ::search_authors
+	 */
+	public function test_search_authors_does_not_backfill_terms_for_low_cap_users(): void {
+
+		global $coauthors_plus;
+
+		$prefix     = 'termgrowth';
+		$subscriber = $this->create_subscriber( "{$prefix}_subscriber" );
+
+		$this->assertEmpty( $coauthors_plus->get_author_term( $subscriber ) );
+
+		$coauthors_plus->search_authors( $prefix );
+
+		$this->assertEmpty( $coauthors_plus->get_author_term( $subscriber ) );
+	}
+
+	/**
+	 * Checks search_authors() still includes a user granted the capability
+	 * directly rather than through a role, now that role__in is gone.
+	 *
+	 * Regression test for the case GaryJones reproduced on wp-env WP 7.1:
+	 * capability combined with role__in incorrectly dropped a subscriber
+	 * granted edit_posts directly as a per-user capability.
+	 *
+	 * @covers ::search_authors
+	 */
+	public function test_search_authors_includes_user_with_directly_granted_capability(): void {
+
+		global $coauthors_plus;
+
+		$prefix     = 'directcap';
+		$editor     = $this->create_editor( "{$prefix}_editor" );
+		$subscriber = $this->create_subscriber( "{$prefix}_subscriber" );
+		$subscriber->add_cap( 'edit_posts' );
+
+		$authors = $coauthors_plus->search_authors( $prefix );
+
+		$this->assertArrayHasKey( $editor->user_login, $authors );
+		$this->assertArrayHasKey( $subscriber->user_login, $authors );
+	}
 }


GaryJones (changes requested):
Thanks for digging into #1057, Faisal — the underlying bug is real and worth fixing. I have some concerns with the current approach on both changes that I'd like to talk through before we merge.

## Change 1 — removing the 10-result cap

The reasoning in the PR description ties the existence of the `coauthors_search_authors_get_terms_args` filter to evidence that the cap was always a workaround. I'd be cautious about that framing — the filter lets sites modify any of the args, not specifically raise the limit, so it's equally consistent with the cap being a deliberate UX/performance choice.

The practical worry is that this turns an autocomplete-style query into an unbounded one. On a large site, a broad search such as "a" could now return thousands of terms; the consumer probably can't usefully display them, and the call simply gets slower with no user-visible benefit.

A safer middle ground would be to raise the default to a bounded value (say 50 or 100) rather than removing the cap entirely. That fixes the reported scenario in #1057 (truncation at 10 with 11+ matches) without removing the safety rail for very large sites.

## Change 2 — removing the `capability` arg from `get_users()`

This is the change I'd like to push back on most. The reasoning that `has_cap()` in the cleanup loop is the authoritative gate is correct **for the returned results**, but there is a side effect I think has been missed:

The `get_users()` call also drives the back-fill of author taxonomy terms in the loop immediately after it (`update_author_term( $found_user )`). With the `capability` arg removed, every matching user — including subscribers, customers, and any other low-cap user — now gets an author term created for them on every search.

On a WooCommerce site with 100k customers, repeated searches for common name fragments will silently grow `wp_terms` and `wp_term_taxonomy` with author terms for users who can never be authors. Those terms are never cleaned up. The `has_cap()` check at the bottom only filters the return value; it doesn't undo the side effect that has already happened.

I think the right fix is to keep the capability gate but move it to a place that doesn't depend on the unreliable `get_users( 'capability' => ... )` arg — i.e. filter inside the back-fill loop:

```php
$cap = apply_filters( 'coauthors_edit_author_cap', 'edit_posts' );
foreach ( $found_users as $found_user ) {
    if ( ! $found_user->has_cap( $cap ) ) {
        continue;
    }
    $term = $this->get_author_term( $found_user );
    if ( empty( $term ) || empty( $term->description ) ) {
        $this->update_author_term( $found_user );
    }
}
```

That preserves the capability check, sidesteps the wp-env / multisite unreliability, and avoids back-filling junk terms for low-cap users.

## Test coverage

The new test is great for Change 1 (count > 10). It doesn't yet cover Change 2's behavioural claim, though, and I think a couple of additions would make the PR much stronger:

1. A test where a subscriber and an editor share a search-matching name, asserting only the editor is returned. The existing `test_search_authors_no_args` covers the empty-search path, but not the keyword path with the capability filter removed.
2. A test that confirms author terms are not created for users without the required capability (i.e. that the back-fill does not have the side effect described above).

## Summary

The bug in #1057 is real — 10 is too low — but I think the current fix is reaching a bit further than it needs to, and Change 2 has a non-obvious data-growth side effect on sites with many low-cap users (WooCommerce, BuddyPress, membership sites, etc.). Could you have a look at the suggested approach above and the additional tests? Happy to discuss further if any of this doesn't sit right.

Ritesh-patel:
> Change 2 — Remove capability from the back-fill get_users() query

I recently came across this issue where it is taking more than 10 seconds to get results from this query. We have many user roles and more than 500k users with subscriber role. Should this user back-fill be optional using a filter? In many instances only guest authors are used instead of WordPress users. Should this be handled on user creation/update instead?

GaryJones (changes requested):
Thanks for this, @faisalahammad — and for the unusually thorough tests. The three you added (results beyond 10, subscribers excluded, no term back-fill for low-cap users) cover exactly the behaviour I'd worry about, and they're the reason I'm comfortable with most of this.

The `number` 10 → 50 change for #1057 is clean and I'm happy with it as-is — bounded, and still adjustable via `coauthors_search_authors_get_terms_args`.

My concern is the capability change. Removing the `capability` arg from the back-fill `get_users()` leaves that call with **neither a capability filter nor a `number`**, so on a large site with many low-privilege accounts (membership / WooCommerce, tens of thousands of subscribers) a short search now hydrates every matching user with `all_with_meta` and then loops `has_cap()` over all of them — on every keystroke. Today the `capability` arg keeps that set down to the handful of editors/authors that match the string.

The tricky part is we can't simply put a `number` cap on it. The back-fill exists only to lazily create an author *term* for a matching user who doesn't have one yet, so the `get_terms()` search can find them. If we cap `get_users()` at, say, 200 and a brand-new editor sits behind 200 same-prefix subscribers in the result order, they'd be cut off, never get a term, and silently never appear. So a raw row cap trades the perf problem for a correctness one.

Worth noting the result side is already correctly gated — the final `has_cap()` loop plus `get_terms( … 'number' )` are the authoritative, bounded filter — so the back-fill is really just term-priming. That means the right shape is to keep the back-fill's *input* small by capability, not by row count.

Two ways forward, whichever you prefer:

1. **Keep a reliable capability/role constraint on the back-fill `get_users()`** rather than dropping it — e.g. `role__in` of the roles that hold the cap, which is reliable across single-site and multisite. If the original `capability` arg was genuinely returning empty for you, I'd be glad to know the exact environment (multisite? wp-env version? how the test users got their roles?), because fixing *that* is better than removing the filter — the filter is what keeps this query cheap at scale.
2. **Split the PR** so the `number` 10 → 50 fix lands now and the capability rework is worked out separately under its own issue.

Marking as changes-requested for that one point; the rest is good, and the tests make me confident the security behaviour is preserved. Thanks again.

faisalahammad:
Hi @GaryJones — updated the change to keep the capability constraint on the back-fill get_users() query while still addressing the wp-env/multisite reliability issue. The query now uses both capability and role__in derived from coauthors_edit_author_cap, plus a new coauthors_search_authors_get_users_args filter. The three integration tests still pass. Could you re-review when you have a moment?

GaryJones:
Thanks for turning this around so quickly, @faisalahammad. Deriving `role__in` from the capability and keeping it as a bound on the back-fill query is the right shape, and it addresses the concern from my review.

One thing to verify before we merge, because it's the crux of the reliability claim: `WP_User_Query` combines `capability` and `role__in` with an **AND** relation. If the `capability` clause is the part that was silently returning empty on multisite/wp-env, then `capability AND role__in` could still come back empty — and where `capability` does work, `role__in` is redundant. Worth confirming on an actual multisite / wp-env setup that the combined query reliably returns capable users; if `role__in` is the dependable bound, we may want it to stand on its own rather than be ANDed with the flaky `capability`.

On timing: 4.1.0 is a stabilisation release and is being cut now, and this is an enhancement (the 10→50 cap lifts a long-standing limit, not a 4.0.x regression) rather than a blocker. So I'm targeting it for **4.2.0**, which gives the multisite verification room to be done properly. Milestoned accordingly — thanks again, this is close.

GaryJones:
Thanks again for turning the earlier round around so quickly, @faisalahammad, and apologies for the wait on the re-review.

I've now traced this through WordPress core and confirmed it on wp-env, and I'd like to split the PR so the genuinely valuable half can land cleanly.

**Change 1 (10 → 50) — let's keep this.** It's the actual fix for #1057, it's low risk, and the default stays overridable through the existing `coauthors_search_authors_get_terms_args` filter. It stands on its own, together with the `..._beyond_default_limit` test.

**Change 2 (adding `role__in` to the back-fill `get_users()`) — I'd like to drop this one.** Reading `WP_User_Query::prepare_query()`, the `capability` argument doesn't behave the way the change assumes:

- `capability => ['edit_posts']` already expands internally to a meta query matching users whose `wp_capabilities` contains `"edit_posts"` **or any role that grants it** (administrator, editor, author, contributor). So on a standard site the added `role__in` selects the same users. It's redundant on the happy path.
- The capability clause and the role clause are combined with **AND**. So if `capability` really is returning empty in some environment, `capability AND role__in` is still empty, and `role__in` can't rescue it. That's the crux I flagged last time, now confirmed in core.
- It also narrows behaviour: a user granted the edit cap **directly** (a per-user cap, or through a custom role that isn't in the derived list) passes the capability clause but fails `role__in`, so the AND drops them from the back-fill where they're included today.

I confirmed that last point on wp-env (WordPress 7.1), searching an editor alongside a subscriber who has `edit_posts` added directly as a per-user cap:

```
roles_with_cap(edit_posts): administrator, editor, author, contributor

capability only        => editor, direct-cap subscriber
capability + role__in  => editor                        (direct-cap subscriber dropped)
```

The direct-cap subscriber returns `has_cap( 'edit_posts' ) === true`, so they legitimately qualify. Today's back-fill surfaces them; the change would silently stop surfacing them. That's a small regression for exactly the flexible setups `coauthors_edit_author_cap` exists to serve.

And the correctness that matters is already covered without any of this: the `has_cap()` filter at the end of `search_authors()`, plus your new back-fill gate, excludes non-capable wpusers regardless of what `get_users()` returns. So that query only needs to be a performance bound, not a correctness one.

If you are seeing `capability` genuinely return empty on wp-env or multisite, I'd very much like to get to the bottom of it, as that would be a core-level issue worth a proper reproduction. A test that isolates it, a capability-only query returning nothing where the user clearly qualifies, would be the thing to anchor any fix. The three tests here don't currently exercise that path: the first back-fills terms directly before searching, and the exclusion test already passes on `develop` thanks to the final `has_cap()` filter.

Practically, the branch now conflicts with `develop`, so it needs a rebase either way. If you're happy to reduce this PR to Change 1 (the cap lift and its test), that's a clean, self-contained change we can land in 4.2.0. If you'd rather keep pursuing the capability side, let's spin that out into a separate issue with a reproduction so we can size it properly.

Thanks again — the #1057 half is a real improvement and I'd like to ship it.

faisalahammad:
@GaryJones Thanks for tracing this through `WP_User_Query::prepare_query()`. Removed `role__in` and the `get_roles_with_cap()` helper. You are right it was redundant on the happy path and dropped users with a direct capability grant, and the AND with `capability` could never rescue an empty result anyway.

Final shape of `search_authors()`:
- `get_terms()` cap raised from 10 to 50.
- Back-fill `get_users()` keeps `capability` only, no `role__in`.
- Back-fill loop skips term creation for any user that fails `has_cap($cap)`.
- `coauthors_search_authors_get_users_args` filter stays, so sites can adjust or skip the back-fill query.

Also rebased onto current `develop`. The old test file (`tests/Integration/CoAuthorsPlusTest.php`) does not exist anymore after the test suite got split up, so the tests now live in `tests/Integration/Bylines/SearchAuthorsTest.php` with the rest of the `search_authors()` coverage. Added a fourth test for the direct-capability-grant case from your repro, so it can not come back.

Ready for another look.

faisalahammad:
@Ritesh-patel Thanks for the report on the back-fill query time on your 500k-subscriber site. This PR does not change when the back-fill runs, but the `coauthors_search_authors_get_users_args` filter added here lets you adjust or skip that `get_users()` call for your site, for example by tightening the query or returning empty to skip it entirely. Worth its own issue if you want a built-in default for very large sites.

PR #1262, opened 2026-04-30
Enhancement: Add export-coauthors and import-coauthors WP-CLI commands

### LINKED ISSUE #1067: Can someone write an export and import script?
Hello Co-Authors-Plus Community,

While migrating blog posts from a multisite to a single site, I'm encountering an issue and need help with the Co-Authors Plus plugin. 

The plugin documentation mentions that "guest authors will exist on a site-by-site basis," which may be why the co-author information is not being correctly exported and imported during the migration. The co-authors do not seem to be correctly linked to the blog posts after the import.

I'd like to know if anyone in the community has written or can point me to a script or method for exporting and importing this co-author data along with the blog post information. This would help streamline merging multisite blog posts into a single site while preserving guest author information.

Thank you for any insights or guidance you can provide! The blog posts are housed in the following URLs for reference:

https://reformedjournal.com/
https://blog.reformedjournal.com/

Best regards,
Andrea Napierkowski
Creative Director at CurlyHost

faisalahammad:
## Summary

Adds `export-coauthors` and `import-coauthors` WP-CLI subcommands for migrating guest author profiles and post associations across environments (e.g. multisite-to-single-site). Post relationships are keyed by slug + post type, so they survive the ID changes that happen during a standard WP export/import cycle.

Fixes #1067

**Scope note:** Only guest-author CPT profiles are handled. WordPress-user coauthor relationships are documented as out of scope in the help text.

---

#### `export_coauthors()` — performance

**Before:**
```php
$posts = get_posts( array( 'posts_per_page' => -1, ... ) );
foreach ( $posts as $post ) {
    $coauthors = get_coauthors( $post->ID ); // per-post query
    foreach ( $coauthors as $index => $coauthor ) { ... }
}
```

**After:**
```php
// Guest authors fetched in batches of 500 + stop_the_insanity() between batches.
// Position derived from a single SQL query on term_relationships.term_order per author.
do {
    $batch_query = new WP_Query( array( 'posts_per_page' => 500, ... ) );
    $this->stop_the_insanity();
} while ( $batch_count === $batch_size );
```

**Why:** The original approach loaded all posts for every guest author into memory and issued a `get_coauthors()` call per post — on a site with 5,000 posts per author this would OOM or time out. The new code matches the batching pattern used in the rest of this file and derives position from `term_order` in a single query.

#### `import_coauthors()` — correctness and safety

**Before:**
```php
$coauthors_plus->add_coauthors( $post_id, $current_logins ); // append=default
```

**After:**
```php
// Explicit append=false replaces the full ordered list; no duplicates introduced.
$coauthors_plus->add_coauthors( $post->ID, $current_logins, false );
```

**Why:** Without the explicit `false`, the default append behaviour could duplicate entries on a second import run. Passing `false` makes the full reordered list the authoritative state.

Additional import guards:
- `is_array()` check on `guest_authors` before `count()`/`foreach` — prevents PHP 8 TypeError on malformed input (`"guest_authors": null`)
- Version mismatch warning when the export was produced by a different plugin version
- Idempotency check skips posts where the author is already assigned

#### Scope isolation

No changes to existing methods. The diff vs `develop` is purely additive — zero lines deleted from pre-existing code.

---

### `tests/Integration/CliExportImportTest.php` — new file

10 integration tests covering all reviewer-requested scenarios:

| Test | Scenario |
|---|---|
| `test_round_trip_reproduces_coauthor_state` | Full export → delete → import restores exact state |
| `test_idempotency_running_import_twice` | Second import is a no-op; no duplicates |
| `test_dry_run_makes_no_db_writes` | `--dry-run` leaves guest-author count unchanged |
| `test_skip_create_links_existing_profile_without_creating` | `--skip-create` re-links posts where profile exists |
| `test_post_not_found_does_not_fatal` | Missing slug increments counter, completes cleanly |
| `test_malformed_json_fails_cleanly` | Non-JSON input exits with error, no crash |
| `test_guest_authors_null_fails_cleanly` | `"guest_authors": null` caught before foreach |
| `test_author_with_no_post_refs_creates_profile` | Empty `post_refs` still creates the profile |
| `test_add_coauthors_does_not_duplicate` | Two import runs → author appears exactly once |
| `test_export_default_file_is_absolute` | Default `--file` path is under `WP_CONTENT_DIR` |

---

## Testing

**Test 1: Round-trip migration**
1. `wp co-authors-plus export-coauthors --file=/tmp/cap-export.json`
2. Transfer JSON to destination site
3. `wp co-authors-plus import-coauthors --file=/tmp/cap-export.json --dry-run` — preview
4. `wp co-authors-plus import-coauthors --file=/tmp/cap-export.json` — execute

Result: Missing profiles created, existing profiles skipped, posts linked at correct byline positions.

**Test 2: Idempotency**
1. Run import twice against the same site
2. Check `get_coauthors()` on any post

Result: Author appears exactly once; no duplicates.

**Test 3: Dry run**
1. `wp co-authors-plus import-coauthors --file=/tmp/cap-export.json --dry-run`
2. Check database for new guest-author CPT posts

Result: No database writes; only log output.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1210,35 +1210,19 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 		WP_CLI::log( sprintf( 'Found %d guest author(s).', $total ) );
 
 		// make_progress_bar is only available when WP-CLI is fully bootstrapped
-		// (not during PHPUnit integration tests). Provide a no-op fallback so
-		// both contexts work without branching the test helpers.
-		if ( function_exists( '\WP_CLI\Utils\make_progress_bar' ) ) {
-			$progress = \WP_CLI\Utils\make_progress_bar( 'Exporting guest authors...', $total );
-		} else {
-			// No-op fallback: tick()/finish() calls become safe no-ops in test context.
-			$progress = new class() {
-				/**
-				 * Advance the progress bar (no-op in test context).
-				 *
-				 * @param int $increment Unused.
-				 * @return void
-				 */
-				public function tick( int $increment = 1 ): void {}
-				/**
-				 * Finish the progress bar (no-op in test context).
-				 *
-				 * @return void
-				 */
-				public function finish(): void {}
-			};
-		}//end if
+		// (not during PHPUnit integration tests). Use null + guards instead.
+		$progress = function_exists( '\WP_CLI\Utils\make_progress_bar' )
+			? \WP_CLI\Utils\make_progress_bar( 'Exporting guest authors...', $total )
+			: null;
 
 		foreach ( $guest_author_ids as $ga_id ) {
 			$guest_author = $coauthors_plus->guest_authors->get_guest_author_by( 'ID', $ga_id );
 
 			if ( ! $guest_author ) {
 				WP_CLI::warning( sprintf( 'Could not load guest author with ID %d; skipping.', $ga_id ) );
-				$progress->tick();
+				if ( $progress ) {
+					$progress->tick();
+				}
 				continue;
 			}//end if
 
@@ -1288,7 +1272,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 						$wpdb->prepare(
 							// phpcs:ignore WordPress.DB.PreparedSQLPlaceholders.UnquotedComplexPlaceholder,WordPress.DB.PreparedSQL.NotPrepared
 							'SELECT ID, post_name, post_type FROM ' . $wpdb->posts . ' WHERE ID IN (' . $id_placeholders . ')',
-							...$post_ids
+							$post_ids
 						),
 						OBJECT_K
 					);
@@ -1308,10 +1292,14 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 				'post_refs' => $post_refs,
 			);
 
-			$progress->tick();
+			if ( $progress ) {
+				$progress->tick();
+			}
 		}//end foreach
 
-		$progress->finish();
+		if ( $progress ) {
+			$progress->finish();
+		}
 
 		// Write JSON to file.
 		$json = wp_json_encode( $export, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES );
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -106,7 +106,7 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 				wp_set_post_terms( $single_post->ID, array( $author_term->slug ), $coauthors_plus->coauthor_taxonomy );
 				WP_CLI::log( "{$count}/{$total_posts}) Added - Post #{$single_post->ID} '{$single_post->post_title}' now has an author term for: " . $author->user_nicename );
 				$affected++;
-			}
+			}//end foreach
 
 			if ( $count && 0 === $count % 500 ) {
 				$this->stop_the_insanity();
@@ -115,14 +115,13 @@ class CoAuthorsPlus_Command extends WP_CLI_Command {
 
 			$args['paged']++;
 			$posts = new WP_Query( $args );
-		}
+		}//end while
 		WP_CLI::log( 'Updating author terms with new counts' );
 		foreach ( $authors as $author ) {
 			$coauthors_plus->update_author_term( $author );
 		}
 
 		WP_CLI::success( "Done! Of {$total_posts} posts, {$affected} now have author terms." );
-
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


GaryJones (dismissed):
Thanks for tackling this, Faisal — #1067 is a real pain point and an export/import command pair is the right shape of solution. I have a few concerns I'd like to work through before we merge though.

## Scope — feature mixed with style cleanup

A large fraction of the diff isn't the new commands. It's `[]` → `array()` conversion, `// end foreach` / `// end while` markers, comment reformatting, and trailing-blank-line removal across methods that have nothing to do with the new feature (`create_terms_for_posts`, `assign_coauthors`, `swap_coauthors`, `migrate_author_terms`, the private SQL helpers, `stop_the_insanity`). Reviewing the actual feature is harder as a result, `git blame` on those lines now points here, and the merge-conflict surface against other branches is wider than it needs to be.

Could you split the style cleanup into its own PR — ideally driven by an actual phpcs config change so the rules are visible — and keep this PR focused on the new commands?

## No automated tests

This is the bit I'd push back on hardest. 442 new lines of CLI logic — file I/O, JSON parsing, post matching, idempotency, dry-run, position handling — and the "Testing" section is a single manual round-trip. For something users will run against production data during migrations, this is genuinely the highest-risk code per line in the plugin and the most in need of tests.

At minimum I think we'd want integration coverage for:

- Round-trip on the same site reproduces coauthor state exactly.
- Idempotency: running `import-coauthors` twice doesn't duplicate or shift positions.
- `--dry-run` truly makes no DB writes.
- `--skip-create` doesn't create profiles, but does still link posts where the profile already exists.
- Posts not found by slug increment the counter but don't fatal.
- Malformed JSON fails cleanly with a useful error.
- An author with no `post_refs` still creates the profile cleanly.

## Performance — likely to OOM on the exact use case it targets

The export does this for every guest author:

```php
$posts = get_posts( array( ..., 'posts_per_page' => -1, ... ) );
foreach ( $posts as $post ) {
    $coauthors = get_coauthors( $post->ID );  // per-post query
    foreach ( $coauthors as $index => $coauthor ) { ... }
}
```

The reported issue is for reformedjournal.com, a multi-blog migration. An author with 5,000 posts means a 5,000-row `get_posts()` load into memory and 5,000 `get_coauthors()` calls just to compute the byline position. That's likely to OOM or run for hours on the sites that most need this command.

Two changes would help:

1. Batch the export loop and call `stop_the_insanity()` between batches, matching the pattern used elsewhere in this file.
2. Derive position from the term-relationship `term_order` in a single query per author, rather than calling `get_coauthors()` per post.

## `add_coauthors()` append/replace semantics

```php
$current_logins = wp_list_pluck( $current_coauthors, 'user_login' );
$position = min( $position, count( $current_logins ) );
array_splice( $current_logins, $position, 0, array( $user_login ) );
$coauthors_plus->add_coauthors( $post->ID, $current_logins );
```

`CoAuthors_Plus::add_coauthors( $post_id, $coauthors, $append = false )` — whether the third arg is append or replace determines whether passing the full reordered list is correct. It's not passed here, so behaviour depends on the default. Worth passing `false` explicitly and covering it with a test that asserts existing coauthors aren't duplicated.

## JSON schema versioning recorded but unused

Export writes `'version' => COAUTHORS_PLUS_VERSION`, but import never reads it. If the schema evolves later, old export files will silently misbehave. Either check the version on import (with a clear error if the major version differs) or drop the field — recording state without acting on it is worse than not recording it at all.

## Type safety on the import path

```php
$total = count( $data['guest_authors'] );
foreach ( $data['guest_authors'] as $entry ) {
```

If a malformed file has `"guest_authors": null` or a string, `count()` and `foreach` both error on PHP 8+. The existing `! isset( $data['guest_authors'] )` check doesn't catch this. An `is_array()` guard would close the gap.

## File path defaults

Export defaults `--file` to `co-authors-export.json`, a relative path resolved against the WP-CLI working directory — brittle. Either require `--file` or default to a predictable absolute location (e.g. somewhere under `WP_CONTENT_DIR`).

## Scope clarification — WP-user coauthors

The export iterates over guest-author CPT posts, so WP-user coauthor relationships on a post don't appear in the output. That's a reasonable scoping decision (the title says "guest authors"), but the linked issue talks about preserving "co-author information" generally, which a user could read either way. Worth being explicit in the PR description and command help text that this command handles guest authors only.

## Build artefact

The `.zip` link in the description is fine for manual testing in passing, but please don't expect reviewers to download binary attachments to evaluate a PR — happy to test from a checkout of the branch.

## Summary

Feature is welcome, but I'd like to see:

1. Style cleanup split into its own PR.
2. Integration tests covering round-trip, dry-run, idempotency, `--skip-create`, malformed JSON, and posts-not-found counter.
3. Batching in the export loop and a single-query position lookup per author.
4. Explicit append/replace flag on `add_coauthors()`, with a test asserting no duplicates.
5. `is_array()` guard on `guest_authors` during import.
6. Either enforce or drop the `version` field on import.
7. Either require `--file` or default to a predictable absolute location.
8. A line in the PR description and CLI help making it explicit that only guest authors are handled.

Happy to discuss any of this if you'd rather take a different approach.

GaryJones:
Thanks, @faisalahammad — this is in good shape: code-complete, tested, and green. The only reason we're not merging now is timing. 4.1.0 is a stabilisation release, and a pair of new WP-CLI commands is a feature rather than a fix.

We're deferring it to 4.2.0, which we're shaping into a more CLI-focused release, so this should be a natural fit there. Nothing for you to do right now — we'll pick it up and re-review (superseding the earlier change request) when 4.2.0 work begins. Thanks for the contribution.

GaryJones:
Hi @faisalahammad,

Thank you again for this. Coming back to it as we start shaping 4.2.0, your PR did something genuinely useful: it put the spotlight on the WP-CLI command class itself. Every subcommand currently lives in one large `CoAuthorsPlus_Command` class, and adding export and import as two more methods would only have grown that monolith.

Rather than do that, we've decided to take your PR as the prompt to improve the underlying structure first. The direction for 4.2.0 is:

- Pin the existing command behaviours with tests, so we can refactor safely.
- Introduce small service classes for the areas the commands touch (guest authors, coauthor assignment, author terms), so the logic is shared and testable rather than tied to the CLI.
- Split the command class into one class per command, in line with our plugin standards.
- Reimplement export and import cleanly on top of that structure.

To be clear, this is not a reflection on your work. It was complete, tested and green; it simply gave us the nudge to tidy the foundations before building on them.

Because the feature will be rebuilt on that new structure, I'm going to close this PR rather than leave it open against code that's about to change underneath it. Issue #1067 stays open and will be closed by the new implementation. We'll make sure your contribution is credited when it lands.

Thanks again for moving this forward, and for your patience.

PR #1283, opened 2026-05-30
fix(editor): normalize coauthors entity store values to term IDs

### LINKED ISSUE #1277: Co-Authors look-up from post.php broken post 4.x upgrade
I checked and this seems not directly related to #1241.

After upgrading from 3.7.0 to 4.0.2, I noticed that co-authors were still visible on `edit.php` in the list, and also rendered correctly on the front-end template for posts, but would no longer show up correctly on `post.php`.

Seems to be a serialization/deserialization issue:

```js
fetch("https://www.site.tld/wp-json/coauthors/v1/authors-by-term-ids?ids=%5Bobject%20Object%5D%2C%5Bobject%20Object%5D&_locale=user"
```

Even with a single author, the ids are still incorrect serialized: 

```
https://www.site.tld/wp-json/coauthors/v1/authors-by-term-ids?ids=%5Bobject%20Object%5D&_locale=user
```

I will attempt to figure out where and why this is happening on this site and will update this ticket once I have more information.

The issue seems to stem from the fact that [`useCoauthorDetails( termIds )`](https://github.com/Automattic/co-authors-plus/blob/cc6f7ffb02a04e6566eec25a216c8167d9e6675f/src/hooks/use-coauthor-details.js#L18) expects a list of termIds, but in our case, receives the full coauthor objects:

```
wp.data.select('core/editor').getEditedPostAttribute('coauthors')
[
    {
        "display_name": "John Doe",
        "user_nicename": "john-doe",
        "user_url": "https://www.site.tld/wp-content/uploads/2024/10/***-150x150.jpg",
        "user_id": 8703
    },
    {
        "display_name": "Jane Doe",
        "user_nicename": "jane-doe",
        "user_url": "https://www.site.tld/wp-content/uploads/2022/06/***-150x150.jpg",
        "user_id": 16
    }
]
```

Which boils down to this returning objects instead of a list of term ids.

```ts
const { getEditedPostAttribute } = select( 'core/editor' );
const coauthors = getEditedPostAttribute( 'coauthors' );
```

When would this have changed and why would it return full objects in our case instead of just the Ids ?

This is as far as I get with limited knowledge about the code. Could you please check or hint at where the `coauthors` attribute is added to the post object when `post.php` is queried?

### See also

- https://github.com/Automattic/co-authors-plus/blob/develop/src/components/co-authors/index.jsx#L54
- https://github.com/Automattic/co-authors-plus/blob/develop/src/components/co-authors/index.jsx#L64

faisalahammad:
## Summary

After the 4.0.0 migration to WordPress core entity store, `getEditedPostAttribute('coauthors')` returns full author objects instead of integer term IDs in some environments. This causes the `/coauthors/v1/authors-by-term-ids` REST call to send `[object Object]` in the query string, resulting in empty responses and a permanent spinner in the sidebar.

Fixes #1277

### src/utils.js — Tighten `extractTermIds()` to trust only `term_id`

After feedback from @GaryJones, the fallback chain is now restricted to plain integers and the `term_id` property of objects. The previous `term_id ?? id ?? user_id ?? ID` chain could resolve to a different author's term because `/coauthors/v1/authors-by-term-ids` calls `get_term($id, 'author')` — which treats every value as a taxonomy **term** ID, not a user ID. Substituting a `user_id` therefore risks rendering the wrong authors silently.

Objects that only carry `user_id`, `id`, or `ID` are now filtered out. Rendering nothing is safer than rendering the wrong author.

```js
export const extractTermIds = ( coauthors ) => {
	if ( ! Array.isArray( coauthors ) || 0 === coauthors.length ) {
		return [];
	}

	return coauthors
		.map( ( item ) => {
			if ( Number.isInteger( item ) ) {
				return item;
			}
			if ( item && 'object' === typeof item ) {
				return item.term_id ?? null;
			}
			return null;
		} )
		.filter( ( id ) => Number.isInteger( id ) );
};
```

### src/components/co-authors/index.jsx — Normalize at read point

**Before:**
```js
coauthorTermIds: coauthors,
```

**After:**
```js
coauthorTermIds: extractTermIds( coauthors ),
```

**Why:** The `useCoauthorDetails` hook and `buildCoauthorTermIds()` both expect and produce integer arrays. Normalizing here preserves that contract without modifying the hook.

### src/__tests__/utils.test.js — Updated tests

Removed tests that asserted the unsafe `user_id`/`id`/`ID` fallback behaviour. Added a test that confirms these shapes are now filtered out, and updated the mixed-entry test to reflect the new contract. All 22 unit tests pass.

## Root cause note (for follow-up)

The objects reported in #1277 carry only `user_id` (`{display_name, user_nicename, user_url, user_id}`) — no `term_id`. With the tightened helper, affected sites will see an empty panel rather than the previous spinner or the silent-wrong-author outcome. The boundary fix here papers over the symptom; the underlying cause (what serialises the `coauthors` taxonomy field as user objects on certain sites) is worth a separate investigation. Likely candidates: a WordPress core REST taxonomy representation change or a third-party plugin filtering `rest_prepare_{$post_type}`.

## Deploy Notes

No new dependencies. No changes to PHP runtime requirements, no schema changes, no database migrations. JS-only fix in the editor sidebar. The `/coauthors/v1/authors-by-term-ids` REST route is unchanged.

## Steps to Test

1. Check out the branch and build: `npm install && npm run build`.
2. On a site with co-authors, open the block editor for a post that has at least one co-author assigned.
3. Verify the Co-Authors sidebar panel renders the assigned authors (no spinner, no permanent loading state).
4. DevTools → Network → filter for `authors-by-term-ids` → confirm the `ids` query parameter contains integers, not `[object Object]`.
5. Add, remove, and reorder co-authors from the sidebar. Save the post and reload — changes persist.
6. Run unit tests: `npm run test:unit` — expect 22 passed.
7. Edge case (regression check): on a site where the entity store returns `user_id`-only objects (see #1277), the panel should now render empty rather than show the wrong author or spin forever.

## AI Usage Disclosure

- [x] This PR was developed with AI assistance.
- [x] AI tools were used for code review feedback analysis, test design, and PR description drafting.

AI was used to:
- Diagnose the namespace mismatch between `user_id` and `term_id` raised by @GaryJones.
- Draft the tightened `extractTermIds` contract and corresponding test updates.
- Audit the PR description against the repo template.

### DIFF
--- a/src/__tests__/utils.test.js
+++ b/src/__tests__/utils.test.js
@@ -157,24 +157,24 @@ describe( 'Utility - extractTermIds', () => {
 		expect( result ).toStrictEqual( [ 42, 43 ] );
 	} );
 
-	it( 'extracts user_id from objects (co-author objects)', () => {
-		const result = extractTermIds( [ { user_id: 8703 }, { user_id: 16 } ] );
-		expect( result ).toStrictEqual( [ 8703, 16 ] );
-	} );
-
-	it( 'extracts id from objects', () => {
-		const result = extractTermIds( [ { id: 5 } ] );
-		expect( result ).toStrictEqual( [ 5 ] );
-	} );
-
-	it( 'extracts ID from objects', () => {
-		const result = extractTermIds( [ { ID: 99 } ] );
-		expect( result ).toStrictEqual( [ 99 ] );
+	it( 'filters out objects that only have user_id, id or ID', () => {
+		const result = extractTermIds( [
+			{ user_id: 8703 },
+			{ id: 5 },
+			{ ID: 99 },
+		] );
+		expect( result ).toStrictEqual( [] );
 	} );
 
 	it( 'handles mixed integer and object entries', () => {
-		const result = extractTermIds( [ 42, { user_id: 43 }, { term_id: 44 } ] );
-		expect( result ).toStrictEqual( [ 42, 43, 44 ] );
+		const result = extractTermIds( [
+			42,
+			{ user_id: 43 },
+			{ id: 5 },
+			{ ID: 99 },
+			{ term_id: 44 },
+		] );
+		expect( result ).toStrictEqual( [ 42, 44 ] );
 	} );
 
 	it( 'filters out objects with no recognizable ID property', () => {
--- a/src/__tests__/utils.test.js
+++ b/src/__tests__/utils.test.js
@@ -3,6 +3,7 @@ import {
 	removeItem,
 	addItemByValue,
 	buildCoauthorTermIds,
+	extractTermIds,
 } from '../utils';
 import {
 	selectedAuthors,
@@ -142,5 +143,57 @@ describe( 'Utility - buildCoauthorTermIds', () => {
 			undefined
 		);
 		expect( result ).toStrictEqual( [ 10 ] );
+	} );
+} );
+
+describe( 'Utility - extractTermIds', () => {
+	it( 'returns integer IDs unchanged', () => {
+		const result = extractTermIds( [ 42, 43, 44 ] );
+		expect( result ).toStrictEqual( [ 42, 43, 44 ] );
+	} );
+
+	it( 'extracts term_id from objects', () => {
+		const result = extractTermIds( [ { term_id: 42 }, { term_id: 43 } ] );
+		expect( result ).toStrictEqual( [ 42, 43 ] );
+	} );
+
+	it( 'extracts user_id from objects (co-author objects)', () => {
+		const result = extractTermIds( [ { user_id: 8703 }, { user_id: 16 } ] );
+		expect( result ).toStrictEqual( [ 8703, 16 ] );
+	} );
+
+	it( 'extracts id from objects', () => {
+		const result = extractTermIds( [ { id: 5 } ] );
+		expect( result ).toStrictEqual( [ 5 ] );
+	} );
+
+	it( 'extracts ID from objects', () => {
+		const result = extractTermIds( [ { ID: 99 } ] );
+		expect( result ).toStrictEqual( [ 99 ] );
+	} );
+
+	it( 'handles mixed integer and object entries', () => {
+		const result = extractTermIds( [ 42, { user_id: 43 }, { term_id: 44 } ] );
+		expect( result ).toStrictEqual( [ 42, 43, 44 ] );
+	} );
+
+	it( 'filters out objects with no recognizable ID property', () => {
+		const result = extractTermIds( [
+			{ display_name: 'John Doe', user_nicename: 'john-doe' },
+			42,
+		] );
+		expect( result ).toStrictEqual( [ 42 ] );
+	} );
+
+	it( 'returns empty array for undefined input', () => {
+		expect( extractTermIds( undefined ) ).toStrictEqual( [] );
+	} );
+
+	it( 'returns empty array for empty array input', () => {
+		expect( extractTermIds( [] ) ).toStrictEqual( [] );
+	} );
+
+	it( 'returns empty array for non-array input', () => {
+		expect( extractTermIds( 'not an array' ) ).toStrictEqual( [] );
 	} );
 } );
--- a/src/components/co-authors/index.jsx
+++ b/src/components/co-authors/index.jsx
@@ -20,6 +20,7 @@ import AuthorsSelection from '../author-selection';
 import {
 	addItemByValue,
 	buildCoauthorTermIds,
+	extractTermIds,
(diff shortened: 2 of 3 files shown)


GaryJones (changes requested):
Thanks for this, @faisalahammad. I've tested the branch and the approach is sound: with co-authors assigned the sidebar renders correctly and the `[object Object]` request is gone, so normalising at this boundary is the right idea.

One thing needs resolving before we merge, though, because as it stands the fallback list can render the *wrong* authors rather than fixing the panel.

`extractTermIds` resolves `term_id ?? id ?? user_id ?? ID`, and the value flows to `/coauthors/v1/authors-by-term-ids`, which does `get_term( $id, 'author' )` — it treats every value as a taxonomy **term** ID. Term IDs and user IDs are different namespaces, so substituting a `user_id` resolves to a different author's term (or none). On a test site:

```
alice: user_id=2  term_id=3        bob: user_id=3  term_id=4
get_term(2,'author') => cap-admin     ← user_id 2 lands on admin's term
get_term(3,'author') => cap-alice
get_term(4,'author') => cap-bob
```

So if the entity-store objects in #1277 carry `user_id` without a `term_id` (which is how the report reads), `extractTermIds` would emit `[2, 3]` and the panel would show **admin + alice** instead of alice + bob — silently wrong authorship, which is harder to notice than the original spinner.

Could we:

1. Trust only `term_id` and plain integers, and drop the `id` / `user_id` / `ID` fallbacks? If an object has no `term_id`, filter it out — rendering nothing is safer than rendering the wrong author. (The `extracts user_id from objects` test would change accordingly.)
2. Confirm the exact object shape you saw in #1277 — does it include `term_id` alongside `user_id`? If the objects are genuinely `user_id`-only, it's worth pinning down what serialises the `coauthors` attribute as objects in the first place, since the boundary fix can only paper over that.

Happy to pair on it if useful. The diagnosis and the test scaffolding are a good base — it's really just the fallback priority that needs tightening.

faisalahammad (commented):
Thanks for catching this — agreed, the `user_id` fallback was unsafe. Pulled `id`, `user_id`, and `ID` out of the chain in `extractTermIds` (src/utils.js:25). Only `Number.isInteger` and `term_id` are now trusted. Objects without `term_id` are filtered out, which renders an empty panel rather than the wrong author.

Confirmed the shape from #1277: the objects carry `display_name`, `user_nicename`, `user_url`, and `user_id` — no `term_id`. So sites affected by that symptom will now see an empty panel instead of the original spinner or the silent-wrong-author outcome. The boundary fix here papers over it; worth a separate look at what serialises the `coauthors` taxonomy field as user objects. My guess is a core REST taxonomy representation change or a third-party filter on `rest_prepare_{$post_type}`, but I haven't tracked it down.

Updated the tests: removed the three cases that asserted the unsafe behaviour, added a test that confirms `user_id`/`id`/`ID`-only objects are now filtered, and updated the mixed-entry case. All 22 tests pass.

Updated the PR description to call out the root-cause gap as a follow-up.

faisalahammad (commented):
Addressed the review feedback in commit 5ae2d77.

- `extractTermIds()` now trusts only `Number.isInteger(item)` and `item.term_id`; removed `id` / `user_id` / `ID` fallbacks (src/utils.js:25).
- Updated `src/__tests__/utils.test.js` to remove the unsafe-fallback tests and confirm objects with only `user_id`/`id`/`ID` are filtered out.
- All 22 unit tests pass. CodeRabbit review returned 0 findings.

Requesting re-review from @GaryJones.

faisalahammad:
@GaryJones — changes addressed and pushed. `extractTermIds` now trusts only `term_id` and plain integers; tests updated. Ready for re-review when you have a moment.

GaryJones (approved):
Re-reviewed — this is the right fix now. `extractTermIds` trusts only `term_id` and plain integers, so it can never feed a `user_id` to `authors-by-term-ids` (the term-vs-user namespace mix-up from the previous version is gone), and the tests cover it well. For a stock install it's an identity passthrough; for an environment whose store returns author objects, it now degrades gracefully — the malformed `?ids=[object Object]` request is gone — instead of hanging the panel.

One thing for context, not a blocker: where those objects carry only `user_id` and no `term_id` (as in #1277), the sidebar will now show an empty author list rather than the wrong authors. That's the correct, safe behaviour, but it means this doesn't resolve the underlying cause in #1277 (something on that site overriding the `coauthors` REST field to return objects rather than the taxonomy's default term-ID array). I'll keep #1277 open for that root cause.

Approving — thanks for the quick turnaround.

PR #1286, opened 2026-05-31
fix(pcp): add translator comments for block sprintf __()

faisalahammad:
## What

Adds `/* translators: %s: author display name */` comments above `sprintf( __( 'Posts by %s' ) )` calls in 3 block files.

## Why

PHPCS flagged `WordPress.WP.I18n.MissingTranslatorsComment` — 3 instances in block files using `sprintf` with `__()` placeholders but no translator comment.

## Files changed
- `php/blocks/block-coauthor-image/class-block-coauthor-image.php`
- `php/blocks/block-coauthor-name/class-block-coauthor-name.php`
- `php/blocks/block-coauthor-avatar/class-block-coauthor-avatar.php`

## Verification

- `php -l` on all 3 files — no syntax errors
- `composer cs` — 0 i18n translator violations

### DIFF
--- a/php/blocks/block-coauthor-avatar/class-block-coauthor-avatar.php
+++ b/php/blocks/block-coauthor-avatar/class-block-coauthor-avatar.php
@@ -113,6 +113,7 @@ class Block_CoAuthor_Avatar {
 				array(
 					'href'  => $link,
 					'rel'   => $rel,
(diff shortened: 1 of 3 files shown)


GaryJones:
Thank you 😄

PR #1287, opened 2026-05-31
fix(pcp): migrate get_terms() to single-array form

faisalahammad:
## What

Migrates 4 instances of deprecated two-parameter `get_terms( $taxonomy, $args )` to the modern single-array form.

## Files changed
- `php/class-wp-cli.php` — 2 calls (lines 842, 879)
- `php/class-coauthors-plus.php` — 1 call (line 1827)
- `template-tags.php` — 1 call (line 610)

## Verification
- `php -l` on all 3 files — no syntax errors
- Modern `get_terms()` accepts single array with `taxonomy` key


GaryJones:
Thank you 😄

PR #1290, opened 2026-06-07
fix: author archive feed outputting as HTML (Regression in 4.0.2)

kawarano:
## Description

This PR fixes a critical regression introduced in **v4.0.2 (via #1250)**, where author archive feeds (e.g., `/author/username/feed/`) unexpectedly render as the standard author archive HTML page instead of the XML/RSS feed output.

In **PR #1250**, `CoAuthors_Plus::fix_author_page()` introduced `$wp_query->init_query_flags();` to clear conflicting query flags on guest author pages. However, this sequence permanently drops the necessary `is_feed` flag to `false` because the method does not preserve or re-assert it after initialization. 

Just like `$is_paged`, this PR preserves `$wp_query->is_feed` before running `init_query_flags()` and safely restores it right after. This ensures that author feeds continue to serve valid XML without interfering with the original fix implemented in #1250.

## Deploy Notes

There are no new dependencies or structural database changes added in this PR. It is a pure logic fix for an existing class method.

## Steps to Test

1. Ensure you are using Co-Authors Plus **v4.0.2** or higher.
1. Create a post and assign it to an author (either a standard WP User or a Guest Author).
1. Visit the author's feed URL in your browser or via cURL (e.g., `https://your-local-site.test/author/username/feed/`).
1. **Expected Behavior (Before this PR):** The URL returns the full HTML page of the author's archive, even though the HTTP header `Content-Type` is set to `application/rss+xml`.
1. **Expected Behavior (After this PR):** The URL correctly outputs the raw XML/RSS feed markup as expected.

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1857,10 +1857,12 @@ class CoAuthors_Plus {
 			// that paginated author archives can still trigger 404 when out of range.
 			// See https://github.com/Automattic/co-authors-plus/issues/1109.
 			$is_paged = $wp_query->is_paged;
+			$is_feed = $wp_query->is_feed;
 			$wp_query->init_query_flags();
 			$wp_query->is_author  = true;
 			$wp_query->is_archive = true;
 			$wp_query->is_paged   = $is_paged;
+			$wp_query->is_feed    = $is_feed;
 
 			if ( ! is_paged() ) {
 				add_filter( 'pre_handle_404', '__return_true' );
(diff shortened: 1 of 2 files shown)


GaryJones:
Thank you for this, @kawarano — and for spotting the regression. We're closing this in favour of #1303, which fixes the same 4.0.2 author-feed regression (introduced by the query-flag reset in #1250) but is slightly more complete: it preserves the feed flags in `fix_author_page()` while also guarding against forcing `is_feed` on a plain author archive, and it adds integration coverage for both cases. The two PRs touch the same function and tests, so we can only take one.

Your report and diff directly informed the fix and you're credited as co-reporter. #1303 is merged to `develop` and will ship in 4.1.0.

PR #1311, opened 2026-06-22
Apply the default byline delimiter when 'between' is an empty string

### LINKED ISSUE #1310: Byline separators dropped between middle authors when the 'between' delimiter is an empty string
Reported on the WordPress.org forum (Soledad theme): with 3+ authors, the separators between the middle authors are missing — e.g. `AnnaBenCara and Dan` instead of `Anna, Ben, Cara and Dan`. https://wordpress.org/support/topic/missing-delimiters-with-3-authors/

## Diagnosis

The byline join in `coauthors__echo()` (template-tags.php) is **correct** with the default separators and is unchanged since 3.7.0 — `coauthors()`, `coauthors_posts_links()` etc. all render `Anna, Ben, Cara and Dan` for 3–5 authors and for guest-author mixes. So this is **not** a core regression in the join itself.

The exact symptom reproduces only when the `between` delimiter is an **empty string**:

```php
coauthors( '' );          // => "AnnaBenCara and Dan"   (middle separators gone, " and " kept)
coauthors( null );        // => "Anna, Ben, Cara and Dan" (correct)
```

The cause is the default-application guard:

```php
if ( ! isset( $separators['between'] ) ) {
    $separators['between'] = apply_filters( 'coauthors_default_between', $default_between );
}
```

`isset()` returns **false** for `null` (so the `, ` default is applied) but **true** for `''` (so no default is applied, leaving an empty separator). The callers default `$between` to `null`, which works — but a theme or snippet passing `coauthors( '' )`, or filtering `coauthors_default_between` to `''`, produces the mangled byline above. The Soledad theme (or a site snippet) is the most likely source of the empty value here.

## Proposed fix

Treat a caller-supplied empty-string `between` the same as unset, so the default is applied:

```php
if ( ! isset( $separators['between'] ) || '' === $separators['between'] ) {
    $separators['between'] = apply_filters( 'coauthors_default_between', $default_between );
}
```

This still honours an intentional empty separator set via the `coauthors_default_between` filter (the filter result is what's applied), and only guards against a loosely-passed empty value from a caller. `betweenLast` has the same theoretical fragility and is worth considering, but the reported symptom is `between`.

## Notes
- Affects 3.7.0 through current `develop` (the logic is identical across them), so it is long-standing rather than a 4.0.x regression.
- Forum reporter has a local snippet adding ` & ` separators as a workaround.

GaryJones:
## Summary

A forum report ([missing delimiters with 3 authors](https://wordpress.org/support/topic/missing-delimiters-with-3-authors/), Soledad theme) described bylines losing the separators between the middle authors — `AnnaBenCara and Dan` instead of `Anna, Ben, Cara and Dan`.

The byline join in `coauthors__echo()` is correct and unchanged since 3.7.0, and renders properly with the default separators for any number of authors, so this isn't a regression in the join itself. The symptom reproduces only when the `between` delimiter is an **empty string**: `coauthors( '' )`. The default is applied with `! isset( $separators['between'] )`, and `isset( '' )` is `true`, so a caller-supplied empty string skips the default and leaves the middle authors run together. A theme or site snippet passing `coauthors( '' )` (or filtering `coauthors_default_between` to `''`) is the source of the empty value.

This treats a caller-supplied empty string the same as unset, so the default `, ` is applied. An empty separator set deliberately through the `coauthors_default_between` filter is still honoured, because the filter result is what gets applied — only a loosely-passed empty value from a caller is corrected.

## Test plan

- [x] New `BylineSeparatorTest`: `coauthors( '' )` over four authors now yields `Anna, Ben, Cara and Dan` (fails without the fix with `AnnaBenCara and Dan`), and an empty separator set via the filter is still respected.
- [x] Existing byline output for default (`null`) separators is unchanged.

Fixes #1310.

### DIFF
--- a/template-tags.php
+++ b/template-tags.php
@@ -140,7 +140,11 @@ function coauthors__echo( $tag, $type = 'tag', $separators = array(), $tag_args 
 	if ( ! isset( $separators['before'] ) ) {
 		$separators['before'] = apply_filters( 'coauthors_default_before', $default_before );
 	}
-	if ( ! isset( $separators['between'] ) ) {
+	// Treat a caller-supplied empty string the same as unset, so the default
+	// delimiter is applied. Otherwise `coauthors( '' )` drops the separator
+	// between the middle authors (e.g. "AnnaBenCara and Dan"). An intentional
+	// empty separator can still be set via the coauthors_default_between filter.
+	if ( ! isset( $separators['between'] ) || '' === $separators['between'] ) {
 		$separators['between'] = apply_filters( 'coauthors_default_between', $default_between );
 	}
 	if ( ! isset( $separators['betweenLast'] ) ) {
--- a/tests/Integration/BylineSeparatorTest.php
+++ b/tests/Integration/BylineSeparatorTest.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Automattic\CoAuthorsPlus\Tests\Integration;
+
+/**
+ * Regression coverage for issue #1310.
+ *
+ * Previously, coauthors__echo() applied the default "between" delimiter only
+ * when the separator was unset. A caller passing an empty string — coauthors( '' ) —
+ * left the middle authors run together ("AnnaBenCara and Dan"), because
+ * isset( '' ) is true. A caller-supplied empty string should fall back to the
+ * default, while an empty separator set deliberately through the
+ * coauthors_default_between filter must still be honoured.
+ *
+ * @covers \coauthors__echo
+ */
+class BylineSeparatorTest extends TestCase {
+
+	/**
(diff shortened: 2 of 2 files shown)


GaryJones:
Closing this without merging. On reflection, the empty-string behaviour is intentional and shouldn't be second-guessed: `coauthors( null )` means "use the default separator" while `coauthors( '' )` means "use no separator". The mangled byline in the linked report comes from an empty string being supplied by the reporter's theme or a site snippet, not from Co-Authors Plus — so the right fix is on their side (pass a real separator, or use the `coauthors_default_between` filter), not a behaviour change here that would override an explicit choice for everyone.

Leaving the contract as-is. The forum thread will be answered with guidance to locate and amend the offending call.

PR #1325, opened 2026-07-01
fix(search): create author terms proactively to avoid stale cache on new users

### LINKED ISSUE #1314: New user doesn't appear in the author search until the object cache is cleared (multisite)
Reported on the WordPress.org forum: https://wordpress.org/support/topic/new-authors-not-showing-up/

## Symptom

A **newly added** user doesn't appear in the author box in the block editor (and possibly in Quick Edit on the posts list). In one instance a new user stayed missing for ~24 hours, then appeared immediately after a manual object-cache flush. It's intermittent — a later attempt to reproduce showed the new user straight away.

## Environment

- WP Engine, **multisite** (~510 users network-wide, ~475 on the main site).
- WP Engine's full caching stack: page cache, edge full-page cache, and a **persistent object cache**.

## Likely cause

A brand-new user has no author term yet, so the author search only finds them once `search_authors()` back-fills one. If the back-fill's `get_users()` pass doesn't surface the new user, no searchable term is created and they stay invisible. Two candidates:

1. The `get_users()` back-fill not returning the new user — the multisite `capability` argument has been reported as unreliable (see #1261), which would skip the back-fill entirely.
2. Core's `get_terms()` query-cache `last_changed` invalidation not propagating reliably on the host's persistent object cache.

That a full cache flush fixes it points at a stale cached query rather than a logic error.

## Note

This is distinct from #849, which concerns an **existing** co-author whose search term goes stale after a profile change. This issue is specifically about a **new** user never becoming searchable in the first place.

faisalahammad:
## Summary
On multisite with a persistent object cache, a newly added user doesn't appear in the author search box until the cache is manually flushed. This happens because the lazy back-fill in `search_authors()` creates terms too late and the term query cache may not propagate the `last_changed` bump across cache nodes fast enough.
Fixes #1314

**Before:** Author terms were only created lazily when `search_authors()` ran its `get_users()` back-fill. If the object cache held stale data, the term creation happened but `get_terms()` missed it.

**After:**
1. `user_register`, `add_user_to_blog`, `set_user_role` and `add_user_role` hooks now call `create_author_term_on_user_registration()` to create the term at registration, blog-add or role-promotion time for users with `edit_posts` capability.
2. The `search_authors()` back-fill is retained as a safety net, but the transient `last_changed` bump is removed since `wp_insert_term()`/`wp_update_term()` already invalidate the terms cache.

**Why:** Proactive creation ensures the term exists in the database before anyone searches. Gating on the co-author capability keeps subscribers and other low-privilege roles from getting terms.

### `tests/Integration/Bylines/AuthorTermRegistrationTest.php` (new)

Integration tests covering: term creation on registration for author-role users, no term for subscribers, idempotency of the callback, role promotion (`set_user_role`), and the `search_authors()` path when the lazy back-fill returns no users.

### `tests/Integration/Bylines/AuthorTermMultisiteTest.php` (new)

Multisite test covering the real `add_user_to_blog()` flow, which fires after `user_register` has already run (before the user is a member of any blog).

## Deploy Notes

No new dependencies. This change only adds WordPress action hooks and integration tests; no special deployment steps are required.

## Testing
**Test 1: Single-site — new author appears immediately**
1. Create a new user with role Author
2. Search for them in the Authors meta box on any post
Result: The user appears immediately in results

**Test 2: Subscriber excluded**
1. Create a new subscriber user
2. Check `get_author_term()` returns null
Result: No term created for non-editors

**Test 3: Automated tests**
```bash
composer test:integration       # all pass
composer test:integration-ms    # all pass
```


GaryJones:
Thanks @faisalahammad. The core idea is sound: creating the term at registration neatly sidesteps the unreliable multisite `get_users( capability )` back-fill that #1261 flagged as the likely culprit, and it follows the `update_author_term_on_profile_update()` precedent. A few things before this goes in, including some history.

### We have created terms on `user_register` before, and reverted it

In 2018, #544 ("Stop handling wp_terms in search_authors, do it in install and profile update", fixing #486/#235) added `add_action( 'user_register', … )` for exactly this purpose. It was reverted the next day by #590. The reasons are reassuring rather than damning: it was reverted to restore a green build with the intent to redo it, and that version also created a **guest author** per user and **removed** the lazy back-fill in favour of a self-respawning activation cron. This PR avoids all three: term only, back-fill retained, and `edit_posts`-gated so we don't reintroduce the subscriber term flooding #486/#235 were about. So I'm comfortable in principle, but given the history I'd like us to be deliberate about the points below.

### The `last_changed` bump looks redundant, and is gated the wrong way

`update_author_term()` calls `wp_insert_term()`/`wp_update_term()`, which already run `clean_term_cache()` → `wp_cache_set_terms_last_changed()` and bump `last_changed` for the `terms` group. `WP_Term_Query` salts its cache key with that value, so the `get_terms()` call just below already sees a fresh key. The manual `wp_cache_set( 'last_changed', … )` writes the same key to the same backend microseconds later, so functionally it changes nothing, and it cannot make anything "propagate faster" across cache nodes (the stated rationale), since it suffers the identical propagation path core just used.

More importantly, it is gated behind `$backfilled`, which is only true when a term was just created, i.e. exactly when core already bumped `last_changed`. The stale-cached-query scenario it describes happens when the term **already exists** and the back-fill does nothing, in which case `$backfilled` stays false and this line never runs. So it is redundant when it fires and absent when its own scenario would need it. I'd drop the bump and the `$backfilled` flag and rely on proactive creation.

### Role promotions aren't covered

The fix covers registration and blog-add, but a user promoted later (subscriber → author) gets no term: `update_author_term_on_profile_update()` only refreshes an existing term, it won't create one. That reproduces the same "not searchable until a cache flush" symptom for promotions, which are about as common as new authors. Strictly #1314 is about new users, so this need not block, but once we adopt "create terms proactively" it's a conspicuous gap. Could we cover `set_user_role`/`add_user_role`, or at least note it as out of scope?

### Tests

A few gaps, some of which the history makes me keen to close:

- No test proves the fix works **when the `get_users()` back-fill misses the user**, which is the actual mechanism of #1314. Filtering `pre_get_users` to return empty and asserting the proactively-created author is still returned by `search_authors()` would pin it.
- `test_author_term_created_on_add_user_to_blog` fakes the multisite path with `do_action( 'add_user_to_blog', … )` after manually adding a role. It never exercises core's real `add_user_to_blog()` (with its `switch_to_blog()`). A `-ms` test calling the real function would be far stronger, especially given 2018's multisite fragility.
- `test_invalid_user_id_returns_void` asserts `assertNull()` on a `: void` method, which can only ever be null. Assert the observable behaviour instead (no term created, no warnings).
- The `search_authors()` `last_changed` half has no coverage. Moot if we drop it.

### Smaller things

- The docblock's cross-node propagation framing overstates what the code does; worth tidying so future readers aren't misled about the mechanism.
- For consistency with the sibling `AuthorTermProfileUpdateTest`, prefer `$this->_cap` over `global $coauthors_plus` in the new test.
- Minor performance note: this adds a synchronous term lookup and possible insert to every qualifying `user_register`/`add_user_to_blog`. Fine for normal use; just worth being aware of for bulk user imports.

### Questions

1. Did we confirm which of #1314's two candidate causes is actually responsible (the #1261 back-fill unreliability vs. `get_terms()` cache propagation)? It affects whether this fully closes it or just mitigates.
2. Should role promotions be in scope, given we're now creating terms proactively?
3. Any objection to dropping the `last_changed` bump?

GaryJones (changes requested):
As above.

faisalahammad:
Addressed the review feedback in 750775c:

- Removed the redundant terms last_changed bump.
- Added proactive term creation for role promotions.
- Added back-fill-miss coverage with filter cleanup.
- Added a real multisite add_user_to_blog test.
- Fixed invalid user coverage and aligned tests with _cap.
- Updated the callback documentation.

Manual testing passed with the PR 1325 build.

faisalahammad (commented):
Implemented the requested changes in 750775c. The redundant terms cache bump was removed, role promotions now create terms proactively, and registration, back-fill miss, invalid user, and real multisite add_user_to_blog coverage were added. Manual testing and both integration suites passed.

faisalahammad:
## CI Fix Summary — 2 failures resolved

The two Behat failures in `Behat Tests (WP trunk, PHP 8.4)` were stale test expectations, not code bugs. The PR's purpose is to create the `cap-<login>` author term the moment a user is created (the fix for #1314), but two scenarios still asserted the old lazy behaviour.

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | features/reassign-terms.feature:110 | RuntimeException on STDOUT mismatch (expected rename branch, got merge branch because `cap-newuser` is pre-created at `wp user create`) | Updated assertion to the merge-branch output and term list `cap-admin` / `cap-newuser` |
| 2 | features/swap-coauthors.feature:143 | RuntimeException on count mismatch (expected `0`, got `1` because `cap-author2` is pre-created at `wp user create`) | Changed count to `1`; surrounding assertions proving the dry-run made no changes are untouched |

Tests ✅ · Verification ✅ · CodeRabbit ✅ · PHP 7.4+ ✅ · Full Behat 162/162 scenarios pass

PR #1326, opened 2026-07-01
fix(author-query): optimize multi-author SQL rewrite for author__in queries

### LINKED ISSUE #1308: Reduce query cost of the author__in / co-author SQL rewrite (OR-chain → IN, trim LEFT JOIN / SQL_CALC_FOUND_ROWS)
Follow-up to #1296, split out now that the opt-out filter has shipped in #1304.

## Problem

When Co-Authors Plus rewrites an author / `author__in` query to include co-authored posts (the behaviour added in #1249 for #1102), the generated SQL is heavier than it needs to be:

- a per-term `OR`-chain — one `( wp_term_taxonomy.taxonomy = 'author' AND wp_term_taxonomy.term_id = 'N' )` clause per resolved co-author term, which can't use a useful index;
- a `LEFT JOIN wp_term_relationships + wp_term_taxonomy`, which multiplies rows;
- participation in `SQL_CALC_FOUND_ROWS`, forcing a full result count.

For queries spanning many authors this is a significant, repeated DB hit. See #1296 for a full reproduction and example SQL.

The opt-out filter `coauthors_plus_is_author_query` (#1304) lets callers skip the rewrite entirely, but does nothing for callers who legitimately want it. This issue covers making the rewrite itself cheaper.

## Proposed direction

- Replace the per-term `OR`-chain with `wp_term_taxonomy.term_id IN ( … )` (a single, indexable clause).
- Use an `INNER JOIN` rather than `LEFT JOIN` where the query is purely co-author-scoped, to avoid row multiplication — keeping the `post_author = N OR …` branch in mind for when `coauthors_plus_should_query_post_author` is true.
- Avoid forcing `SQL_CALC_FOUND_ROWS` participation where it isn't needed.

## Acceptance

- Identical result set to today for author archives and `author__in` / comma-separated `author` queries (regression coverage in `AuthorMultiQueryTest` / `AuthorQueriedObjectTest`).
- A materially cheaper query plan for the multi-author case, ideally demonstrated with `EXPLAIN` before/after.

Reported by @lubo-makky in #1296.

faisalahammad:
## Summary

Optimizes the author/co-author SQL rewrite for multi-author queries (`author__in` / comma-separated `author`). Replaces the per-term OR-chain with a single indexable `IN (...)` clause, and uses INNER JOIN when `post_author` is excluded from the query.

Fixes #1308

**`posts_join_filter()`** — Choose JOIN type based on multi-author path:
- When `coauthors_plus_should_query_post_author` returns false: INNER JOIN
- Otherwise: LEFT JOIN (default, preserves post_author fallback)

**`posts_where_filter_multi_author()`** — Replace per-term OR-chain:
- Before: `(taxonomy = 'author' AND term_id = 'N') OR (taxonomy = 'author' AND term_id = 'M')`
- After: `taxonomy = 'author' AND term_id IN ('N', 'M')`

**`posts_groupby_filter()`** — Add GROUP BY for multi-author deduplication when JOIN produces multiple rows per post.

### `tests/Integration/Queries/AuthorQueriesTest.php`

3 new SQL structure tests:
- `test_multi_author_query_uses_in_clause_not_or_chain`
- `test_multi_author_query_uses_inner_join_when_post_author_excluded`
- `test_multi_author_query_uses_left_join_by_default`

2 new GROUP BY dedup tests, added per review feedback (no existing test had a post matching two queried co-authors at once, so the dedup branch was never exercised):
- `test_author_in_returns_a_shared_coauthored_post_only_once`
- `test_author_in_returns_a_shared_coauthored_post_only_once_when_post_author_excluded`

## Deploy Notes

None — no new dependencies. Internal query-building change only (SQL string construction plus a GROUP BY clause), no schema or data changes.

## Steps to Test

1. Check out this PR.
2. Run `composer test:integration` and `composer test:integration-ms` — all tests pass, including the 5 new ones covering the `IN (...)` clause, JOIN type, and GROUP BY dedup.
3. Optional manual check: co-author a post with two users, then query `author__in` with both of their IDs (e.g. via `WP_Query` or an author archive covering both) — the post must come back exactly once, not duplicated.

## Testing

- `composer test:integration` — 274 tests pass (5 new)
- `composer test:integration-ms` — 276 tests pass
- `composer cs` — clean on modified files
- CodeRabbit review: 0 findings

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -986,6 +986,36 @@ class CoAuthors_Plus {
 				return $join;
 			}
 
+			// Multi-author path (author__in / comma-separated author).
+			$author_ids      = $this->get_author_ids_from_query( $query );
+			$is_multi_author = ! empty( $author_ids ) && ( ! $query->is_author() || count( $author_ids ) > 1 );
+
+			if ( $is_multi_author ) {
+				$maybe_both = $this->force_guest_authors
+					? false
+					: apply_filters( 'coauthors_plus_should_query_post_author', true );
+
+				if ( $maybe_both ) {
+					// Need post_author fallback -- keep LEFT JOIN.
+					$rel_join = " LEFT JOIN {$wpdb->term_relationships} AS tr1 ON ({$wpdb->posts}.ID = tr1.object_id)";
+					$tax_join = " LEFT JOIN {$wpdb->term_taxonomy} ON ( tr1.term_taxonomy_id = {$wpdb->term_taxonomy}.term_taxonomy_id )";
+				} else {
+					// Only co-author terms -- INNER JOIN is sufficient.
+					$rel_join = " INNER JOIN {$wpdb->term_relationships} ON ({$wpdb->posts}.ID = {$wpdb->term_relationships}.object_id)";
+					$tax_join = " INNER JOIN {$wpdb->term_taxonomy} ON ( {$wpdb->term_relationships}.term_taxonomy_id = {$wpdb->term_taxonomy}.term_taxonomy_id )";
+				}
+
+				if ( false === strpos( $join, trim( $rel_join ) ) ) {
+					$join .= $rel_join;
+				}
+				if ( false === strpos( $join, trim( $tax_join ) ) ) {
+					$join .= $tax_join;
+				}
+
+				return $join;
+			}
+
+			// Single-author path: only add ONCE having_terms is populated.
 			if ( empty( $this->having_terms ) ) {
 				return $join;
 			}
@@ -1240,8 +1270,20 @@ class CoAuthors_Plus {
 			return $where;
 		}
 
-		$terms_implode = $this->build_terms_clauses( $all_terms );
-		$this->having_terms = rtrim( $this->having_terms, ' OR' );
+		// Build a single IN clause instead of a per-term OR-chain.
+		$term_ids = array_unique(
+			array_map(
+				function ( $t ) {
+					return $t->term_id;
+				},
+				$all_terms
+			)
+		);
+		$placeholders = implode( ', ', array_fill( 0, count( $term_ids ), '%d' ) );
+		$terms_implode = $wpdb->prepare(
+			"{$wpdb->term_taxonomy}.taxonomy = %s AND {$wpdb->term_taxonomy}.term_id IN ({$placeholders})",  // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+			array_merge( array( $this->coauthor_taxonomy ), $term_ids )
+		);
 
 		$maybe_both = $this->force_guest_authors
 			? false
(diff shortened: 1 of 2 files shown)


GaryJones:
Thanks for this — I reviewed the change and ran it locally, and the multi-author rewrite looks correct. All of the existing `AuthorQueriesTest` cases pass against the branch, and the key structural points hold up:

- The `IN (...)` clause is parameterised correctly (`%s` for the taxonomy, `%d` per term, arguments in matching order).
- The JOIN aliases stay consistent with the WHERE clause: the LEFT path keeps the `tr1` alias, the INNER path uses the unaliased table, and both provide the unaliased `wp_term_taxonomy` that the WHERE references.
- `maybe_both` is derived the same way in `posts_join_filter()` and `posts_where_filter_multi_author()`, so the JOIN type and the `post_author` fallback never disagree.
- No `having_terms` staleness: the multi-author path never reads it, and `build_terms_clauses()` resets it for the single-author path.

## One coverage gap worth closing before merge

Nothing currently exercises the new `GROUP BY` in `posts_groupby_filter()`. None of the existing tests have a post that matches two queried authors' terms at the same time, so the JOIN never produces duplicate rows and the dedup branch is never hit.

I confirmed this by disabling the `GROUP BY`: the whole suite still passed, and only once I added a test that shares a post between two queried co-authors did the post come back twice (`[5, 5]`).

Here are two tests that close the gap, covering both the default LEFT JOIN path and the INNER JOIN path taken when `coauthors_plus_should_query_post_author` opts out. Both pass on this branch, and both fail if the `GROUP BY` is removed:

```php
	/**
	 * A single post co-authored by two of the queried authors must be returned
	 * exactly once. The per-term taxonomy JOIN yields one row per matching term,
	 * so this guards the GROUP BY deduplication in posts_groupby_filter().
	 */
	public function test_author_in_returns_a_shared_coauthored_post_only_once(): void {
		$author1 = $this->create_author( 'dedup_a1' );
		$author2 = $this->create_author( 'dedup_a2' );

		$post = $this->create_post( $author1 );
		$this->_cap->add_coauthors( $post->ID, array( $author1->user_login, $author2->user_login ) );

		// Both queried authors co-author the same post, so the JOIN produces one
		// row per matching term. Without GROUP BY the post would appear twice.
		$query = new WP_Query( array( 'author__in' => array( $author1->ID, $author2->ID ) ) );

		$this->assertQueryReturns(
			array( $post->ID ),
			$query,
			'A post shared by two queried co-authors must be returned once, not duplicated by the JOIN.'
		);
	}

	/**
	 * The GROUP BY deduplication must also hold on the INNER JOIN path taken when
	 * `coauthors_plus_should_query_post_author` opts out of the post_author fallback.
	 */
	public function test_author_in_returns_a_shared_coauthored_post_only_once_when_post_author_excluded(): void {
		$author1 = $this->create_author( 'dedup_inner_a1' );
		$author2 = $this->create_author( 'dedup_inner_a2' );

		$post = $this->create_post( $author1 );
		$this->_cap->add_coauthors( $post->ID, array( $author1->user_login, $author2->user_login ) );

		add_filter( 'coauthors_plus_should_query_post_author', '__return_false' );

		$query = new WP_Query( array( 'author__in' => array( $author1->ID, $author2->ID ) ) );

		remove_all_filters( 'coauthors_plus_should_query_post_author' );

		$this->assertQueryReturns(
			array( $post->ID ),
			$query,
			'A shared co-authored post must be returned once on the INNER JOIN path too.'
		);
	}
```

Happy to push these for you if you enable "Allow edits by maintainers" on the PR, or I can open them as a quick follow-up — whichever suits.

## Two minor, non-blocking notes

- `coauthors_plus_should_query_post_author` now fires twice per multi-author query (once in the where filter, once in the join) rather than once. Only relevant for a filter callback with side effects, so fine to leave.
- The `GROUP BY` is only added when `$groupby` is empty, so a `groupby` set by another plugin would skip the dedup. An edge case, not something to hold this up for.

GaryJones (changes requested):
As above.

faisalahammad:
Thanks for the review and for writing the tests.

Added both tests as you wrote them, in `tests/Integration/Queries/AuthorQueriesTest.php`:
- `test_author_in_returns_a_shared_coauthored_post_only_once`
- `test_author_in_returns_a_shared_coauthored_post_only_once_when_post_author_excluded`

Checked they catch the gap. Disabled the GROUP BY line, both tests failed. Restored it, both pass again. Full suite is green: `composer test:integration` (274 pass, 5 new) and `test:integration-ms` (276 pass). CodeRabbit is clean too.

The two minor notes (filter firing twice, groupby only set when empty) are fine to leave as you said. No change made there.

Pushed in 0c02510. Ready for another look.

Issue #1135, opened 2025-09-19
Fatal Error when creating a guest author with a name of a user that already exists.


mattpramschufer:
We ran into an issue while trying to create a Guest Author with the display name Joe Greene. When clicking “Create Guest Author,” the site immediately returned a 500 Internal Server Error.

No errors were logged in debug.log, and no fatal error notification was sent to the site administrator, which was unexpected. The only clue came after checking the wp_users table, where I discovered that an existing subscriber already has the display name and slug Joe Greene.

Unfortunately, since nothing was logged, I don’t have a stack trace or specific error message to share. The only output shown was the generic Internal Server Error page.

Just figured I would report it.

PR #1341, opened 2026-08-09
Allow creating a new guest author from the block-editor Co-Authors sidebar

### LINKED ISSUE #53: Create new user or guest author while assigning to a post

faisalahammad:
## Description

Adds a `Create new guest author` action to the Co-Authors panel in the block editor. Editors can create a new byline for someone not yet on the site without leaving the post screen, addressing the friction that has been tracked in #53 since 2012.

The change ships a new REST endpoint `POST /coauthors/v1/guest-authors` and a small inline modal in the existing sidebar component. The new endpoint reuses `CoAuthors_Guest_Authors::create()` so field validation, term creation, and failure cleanup stay in one place. The handler maps the create() error codes (`field-required`, `invalid-field`, `invalid-email`, `duplicate-field`) to REST HTTP statuses (400 / 409) so the modal can show meaningful messages instead of a generic 500.

Scope is block editor only. The classic editor sidebar is left untouched. Creating a full WordPress user account (login, role, password, email notification) is intentionally out of scope for this PR.

Fixes #53

## Deploy Notes

No new dependencies. No build, schema, or deployment changes required. The endpoint is namespaced under the existing `coauthors/v1` namespace and is gated by `current_user_can_set_authors()` via the same permission callback as the search and save routes.

### Automated

```sh
npx @wordpress/env start
docker exec wp-env-co-authors-plus-<id>-tests-cli-1 \
  /bin/bash -c 'cd wp-content/plugins/co-authors-plus && \
    ./vendor/bin/phpunit --testsuite integration --exclude=ms-required \
      --no-coverage --filter EndpointsTest'
```

(Replace `<id>` with your local wp-env hash. `npx @wordpress-env status` prints it.)

The new tests cover the happy path, blank display name, duplicate login, and HTTP status mapping.

Linters:

```sh
composer lint
composer lint-ci
npm run lint:js
npm run lint:css
npm run test:unit
```

### Manual

1. Open any post in the block editor.
2. In the Co-Authors sidebar, type a name that does not exist (e.g. `zzz_nonexistent_author`) in the `Select An Author` combobox.
3. Click `+ Create new guest author` directly below the combobox.
4. The modal opens with `Display name` prefilled to `zzz_nonexistent_author`. Leave email blank, click `Create`.
5. The modal closes. The new guest author appears as a chip in the Co-Authors panel.
6. Save the post and reload. The byline persists.
7. Open Users -> Guests. The new author is listed with slug `cap-zzz_nonexistent_author`.
8. Open another post, click `+ Create new guest author` again with a name that already exists. The modal stays open and shows an error like `That login already exists.` (HTTP 409).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
 			"devDependencies": {
 				"@playwright/test": "^1.62.1",
 				"@testing-library/dom": "^10.4.1",
+				"@testing-library/jest-dom": "^6.9.1",
 				"@testing-library/react": "^16.3.3",
 				"@wordpress/e2e-test-utils-playwright": "^1.54.0",
 				"@wordpress/icons": "^15.5.0",
@@ -26,6 +27,13 @@
 				"prettier": "npm:wp-prettier@^2.2.1-beta-1",
 				"version-bump-prompt": "^6.1.0"
 			}
+		},
+		"node_modules/@adobe/css-tools": {
+			"version": "4.5.0",
+			"resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.5.0.tgz",
+			"integrity": "sha512-6OzddxPio9UiWTCemp4N8cYLV2ZN1ncRnV1cVGtve7dhPOtRkleRyx32GQCYSwDYgaHU3USMm84tNsvKzRCa1Q==",
+			"dev": true,
+			"license": "MIT"
 		},
 		"node_modules/@ariakit/components": {
 			"version": "0.1.11",
@@ -6028,6 +6036,33 @@
 			"dev": true,
 			"license": "MIT"
 		},
+		"node_modules/@testing-library/jest-dom": {
+			"version": "6.9.1",
+			"resolved": "https://registry.npmjs.org/@testing-library/jest-dom/-/jest-dom-6.9.1.tgz",
+			"integrity": "sha512-zIcONa+hVtVSSep9UT3jZ5rizo2BsxgyDYU7WFD5eICBE7no3881HGeb/QkGfsJs6JTkY1aQhT7rIPC7e+0nnA==",
+			"dev": true,
+			"license": "MIT",
+			"dependencies": {
+				"@adobe/css-tools": "^4.4.0",
+				"aria-query": "^5.0.0",
+				"css.escape": "^1.5.1",
+				"dom-accessibility-api": "^0.6.3",
+				"picocolors": "^1.1.1",
+				"redent": "^3.0.0"
+			},
+			"engines": {
+				"node": ">=14",
+				"npm": ">=6",
+				"yarn": ">=1"
+			}
+		},
+		"node_modules/@testing-library/jest-dom/node_modules/dom-accessibility-api": {
+			"version": "0.6.3",
+			"resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.6.3.tgz",
+			"integrity": "sha512-7ZgogeTnjuHbo+ct10G9Ffp0mif17idi0IyWNVA/wcwcm7NPOD/WEHVP3n7n3MhXqxoIYm8d6MuZohYWIZ4T3w==",
+			"dev": true,
+			"license": "MIT"
+		},
 		"node_modules/@testing-library/react": {
 			"version": "16.3.3",
 			"resolved": "https://registry.npmjs.org/@testing-library/react/-/react-16.3.3.tgz",
@@ -10870,6 +10905,13 @@
 			"funding": {
 				"url": "https://github.com/sponsors/fb55"
 			}
+		},
+		"node_modules/css.escape": {
+			"version": "1.5.1",
+			"resolved": "https://registry.npmjs.org/css.escape/-/css.escape-1.5.1.tgz",
+			"integrity": "sha512-YUifsXXuknHlUsmlgyY0PKzgPOr7/FjCePfHNt0jxm83wHZi44VDMQ7/fGNkjY3/jV1MC+1CmZbaHzugyeRtpg==",
+			"dev": true,
+			"license": "MIT"
 		},
 		"node_modules/cssesc": {
 			"version": "3.0.0",
@@ -23659,6 +23701,12 @@
 		}
 	},
 	"dependencies": {
+		"@adobe/css-tools": {
+			"version": "4.5.0",
+			"resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.5.0.tgz",
+			"integrity": "sha512-6OzddxPio9UiWTCemp4N8cYLV2ZN1ncRnV1cVGtve7dhPOtRkleRyx32GQCYSwDYgaHU3USMm84tNsvKzRCa1Q==",
+			"dev": true
+		},
 		"@ariakit/components": {
 			"version": "0.1.11",
 			"resolved": "https://registry.npmjs.org/@ariakit/components/-/components-0.1.11.tgz",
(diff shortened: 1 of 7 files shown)


faisalahammad (commented):
## CI Fix: PHPCS NoPaging Sniff

CI "Basic CS and QA checks" was failing on `WordPressVIPMinimum.Performance.NoPaging.posts_per_page_numberposts`. The test used `get_posts( numberposts => -1 )` purely to count rows, which is the wrong tool for counting and trips the sniff.

### Fix
| File | Error | Fix |
|------|-------|-----|
| `tests/Integration/GuestAuthors/CreateGuestAuthorTest.php` | `NoPaging.posts_per_page_numberposts` on `numberposts => -1` (2 sites) | Replaced both `get_posts()` calls with a new `count_guest_author_posts()` helper that uses `wp_count_posts()` and sums all statuses except `trash` (matching the previous `post_status => 'any'` semantic). |

### Verification
- `composer cs -- tests/Integration/GuestAuthors/CreateGuestAuthorTest.php` → 0 errors
- `composer test:integration -- --filter CreateGuestAuthorTest` → 4/4 pass, 10 assertions
- PHP 7.4+ / WP 6.4+ compatible (no new APIs beyond `wp_count_posts`, which is stable since WP 2.5)

Test intent unchanged: still asserts no orphan guest author post is left behind when `wp_insert_term` fails during `CoAuthors_Guest_Authors::create()` (refs #1135).

GaryJones (commented):
Thanks for picking this one up, Faisal. #53 has been open since 2012, so it is good to see it moving. The overall shape is right: reusing `CoAuthors_Guest_Authors::create()` rather than reimplementing validation and cleanup is the correct instinct, the operation is properly a POST, and mapping the `create()` error codes onto sensible HTTP statuses is a nice touch.

I have verified a few things locally that I think need addressing before this goes in. Two of them are simply consequences of the branch having aged.

## 1. The branch is 37 commits behind develop, and one of the conflicts is semantic

`develop` has since landed 9a92830, "Fix author search never reaching the server in the block editor", which rewrote `onFilterValueChange` to read `fetchAuthors` and `threshold` from a `latest` ref instead of capturing them in the debounced closure.

This PR adds `lastQuery.current = query` into the older, closure capturing version of that handler. If the conflict is resolved in favour of this branch, it will quietly reintroduce the bug that 9a92830 fixed.

Could you rebase rather than merge, and place the `lastQuery` assignment by hand inside the new debounced callback that destructures from `latest.current`?

## 2. Commit 24c919a is now a duplicate and can be dropped

`develop` already absorbed the VIPCS NoPaging drift in 3ddbebd, using a `phpcs:ignore` on the existing `get_posts()` call. This branch solves the same sniff a second way, via `wp_count_posts()`.

Worth dropping during the rebase. Separately, `wp_count_posts()` is object cached, which is not ideal in a test that counts posts it has just created and deleted within the same request.

Between points 1 and 2, the green CI on this PR reflects a base that no longer exists, so it is worth rerunning once rebased.

## 3. `display_name` is stored without sanitisation

This is the one I would most like to see addressed.

The classic admin path (`manage_guest_author_save_meta_fields`) runs every field through its `sanitize_function`, falling back to `sanitize_text_field()`. `create()` writes the meta raw:

```php
update_post_meta( $post_id, $pm_key, $args[ $key ] );
```

and the new route registers no `sanitize_callback`. Dispatching a real request through `rest_do_request()` as an Editor:

```
sent           : "Evil <script>alert(1)</script>\nName\tWith\0Control"
stored in meta : "Evil <script>alert(1)</script>\nName\tWith\0Control"   (HTTP 200)
classic path   : "Evil Name WithControl"
```

Script tags, newlines, tabs and a NUL byte all persist into `cap-display_name` and `post_title`. `update_post_meta()` is not kses filtered, so this holds on multisite too, irrespective of `unfiltered_html`.

To be fair about severity: this is not a demonstrated XSS in CAP itself, since `_format_author_data()` applies `esc_html` and our own templates escape. But it is a real divergence from the classic path, it stores a NUL byte, and it matters here because this PR is the first thing to make `create()` reachable over HTTP. The existing production callers are only WP-CLI and `bin/seed-demo-data.php`. Third party byline templates that echo `cap-display_name` unescaped are common enough to be worth defending against.

A `sanitize_callback` on each registered arg should cover it.

## Smaller points

**No JS tests.** The PHP coverage is genuinely good, but nothing exercises `onSubmitCreate`, the disabled button guard, error surfacing, or the `lastQuery` prefill, despite `src/__tests__/` already existing.

**Test fidelity.** The new tests pass parameters via `set_url_params()` on a POST. `get_param()` happens to find them, so the tests pass, but the registered `args` are never exercised, including the `user_email` `validate_callback`. I confirmed `set_body_params()` with `rest_do_request()` works fine here, so this is an easy upgrade that would give you real coverage of the arg schema.

**Capability gating.** I checked this properly rather than assuming. The endpoint uses `current_user_can_set_authors()`, which is native `edit_others_posts`, so Editor and above. For reference, `post-new.php?post_type=guest-author` is reachable by a Contributor, because the post type registers no `capability_type` and core maps `create_posts` onto `edit_posts`. So the endpoint is stricter than the existing classic route, which is fine. The only observation is that it does not honour the documented `add_guest_author_cap` filter, so a site that has deliberately tightened guest author creation will find this route ignores that. Worth a conscious decision either way.

**The trigger renders unconditionally** below the combobox, rather than appearing when a search returns nothing. #53 describes the latter. Not wrong, but worth confirming it is intentional.

## Not caused by this PR, but noted while reviewing

`_format_author_data()` calls `update_author_term()`, which runs `wp_insert_term` / `wp_update_term`. That means the existing GET routes write on read. It predates this PR and I will raise it separately, but it is also why `create_guest_author()` ends up touching the term twice in a single request.

Happy to help with the rebase if that would be useful.

GaryJones (changes requested):

faisalahammad (commented):
Addressed the review feedback:

- Rebased onto current develop and preserved the latest debounced search implementation; lastQuery is assigned inside the latest-ref callback.
- Dropped the duplicate NoPaging workaround commit.
- Added REST sanitization for display_name, user_email, and user_login.
- Added a disabled-guest-authors guard with a controlled 404 response.
- Updated endpoint tests to use set_body_params() and rest_do_request(), including invalid email and control-character sanitization coverage.
- Added block-editor modal tests for query prefill, disabled Create state, successful creation, and error display.

GaryJones (commented):
Thanks Faisal, this is a good round of changes. I have checked each point against the diff rather than just the summary, and everything substantive from my review has been addressed properly.

- The `lastQuery` assignment now sits inside the current debounced callback and reads `minLength` from `latest.current`, so the 9a92830 fix is preserved. This was the one I was most worried about, and it is resolved correctly.
- The duplicate NoPaging commit is gone; the branch no longer touches `CreateGuestAuthorTest.php` at all.
- The sanitisation is right, and I want to call out one detail approvingly: the custom callback stripping `\x00–\x1F` before `sanitize_text_field()` is not redundant, since `sanitize_text_field()` misses NUL, and excluding tab, LF and CR from the regex (which the whitespace collapse already handles) is correct. Using the probe payload from my review as the test fixture gives us a proper regression anchor.
- The PHP tests now dispatch through `set_body_params()` and `rest_do_request()`, so the registered args and the `validate_callback` are genuinely exercised, and the JS tests cover the prefill, the disabled Create guard, success and error paths. The `is_guest_authors_enabled()` guard and the removal of the unreachable `invalid-email` mapping are both sensible additions.

Two small things remain before this can go in:

1. **One more rebase, sorry.** develop has moved again (81 commits), and there is a single conflict in `src/__tests__/co-authors.test.js`. It is mechanical this time: 8a7860c and c17daae added tests on develop that use `comboboxProps`, which this branch renamed to `mockComboboxProps`. Renaming the new occurrences during the rebase is all that is needed.

2. **A whitespace glitch from the conflict resolution** at `src/components/co-authors/index.jsx` around line 240: `const onFilterValueChange` is indented with two tabs instead of one, and the `useCallback(` line below it is missing one. CI did not flag it, so noting it here.

On the two open questions from my first review: I am happy to let the `add_guest_author_cap` point go, since the endpoint gate is defensibly stricter than the classic `post-new.php` path, and the always visible trigger can stay if that placement is intentional. A one line confirmation on the latter is all I would ask.

Once the rebase and the indentation fix are up, this looks ready.

faisalahammad (commented):
Rebased onto current develop and fixed the whitespace glitch. Both changes are up.

- Rebase: the historic conflict is gone. The two dropdown tests develop added (8a7860c, c17daae) are kept, with their `comboboxProps` renamed to `mockComboboxProps` to match the rest of this file. The PR modal tests sit alongside them, both describe blocks intact.
- Indentation: `const onFilterValueChange = useDebounce(` in `src/components/co-authors/index.jsx` is back to a single tab.

The always visible `+ Create new guest author` trigger below the combobox is intentional. It keeps the action discoverable even when the user has not typed a search, and the modal prefill already handles the common case. Happy to change it to only appear on an empty search result if you would rather match the issue wording.

faisalahammad:
@GaryJones The rebase onto current develop and the indentation fix are up. My note above also confirms the always visible trigger is intentional. Ready for another look when you have time.

Issue #1313, opened 2026-06-23
Cache hardening: extract a cache layer, close invalidation gaps, and reduce uncached hot paths


GaryJones:
Captured from an audit of Co-Authors Plus's caching.

## Shape of the current caching

- **17 production call sites** (4 `get`, 5 `set`, 8 `delete`), all `wp_cache_*`, across three files. No transients, no `wpcom_vip_*` helpers, no multi-key APIs.
- **Two ad-hoc groups**: `co-authors-plus` and `coauthors-plus-guest-authors`. No global-group or non-persistent-group registration; keys are not blog-prefixed (relies on the object cache's per-blog salting).
- **Mostly idiomatic** — reads sit on top of core functions that carry their own caching (`wp_get_object_terms`, `get_term_by`, `get_user_by`); the only bespoke layer is the guest-author postmeta lookup (reasonable, core can't cache that shape). The guest-author negative-cache (storing `'0'` for misses, read back with an `is_object()` guard) is correct and intentional.
- **No cache abstraction**: groups and key formats are hard-coded string literals scattered through business-logic methods (only the guest-author side has a partial `get_cache_key()`). This is the structural root cause of invalidation drift.

## Work items (prioritised)

- [ ] **Extract a cache infrastructure layer** (structural keystone). A small helper owning `group()` / `key()` / `forget()` so reads, writes and invalidation can't drift apart. The lack of this is how #849 happened.
- [ ] **Re-key the author-term cache on the stable `user_id`** instead of `user_nicename`, and invalidate on `set_user_role` / `user_register`. #1312 added a `profile_update` refresh, but a *nicename* change still orphans the old entry (the term slug is `cap-<nicename>`). `php/class-coauthors-plus.php` `get_author_term()` (~2362).
- [ ] **Cache `search_authors()`** — currently recomputed on every AJAX author-search keystroke (`get_users()` + `get_terms()` + a per-term `get_coauthor_by()` loop). `php/class-coauthors-plus.php`.
- [ ] **Fix `get_all_linked_accounts()` cold-path N+1** — it loops `get_user_by('login', …)` over every linked account on a cache miss; collapse to one `get_users(['login__in' => …])` (or `wp_cache_get_multiple`). Also invalidate it when a linked user is deleted outside the guest-author flow. `php/class-coauthors-guest-authors.php` (~1285).
[...]

Issue #849, opened 2022-05-17
Potential cache issue with ajax search on admin


jonathanstegall:
Lately I've had issues with the admin search for co authors on posts. Only a few (maybe ten out of 2,000+) authors are appearing in the autosuggest. When I search for authors with known names that we've used, even very frequently, they don't show up. At first I thought it might be related to Elasticsearch because I had just updated our index before I noticed it (we host on WordPress VIP) and it was behaving differently across the different environments. It is broken in all three (develop, preprod, production) although the results are different across all three. I've yet to see any issues locally.

If my debugging is correct, I no longer think it's related to Elasticsearch because according to [this method](https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-coauthors-guest-authors.php#L891-L975) it's a SQL query that never runs through Elasticsearch. The results are stored in memcache though, so it could be a cache issue and that would also explain why it behaves differently between different environments. I've also noticed that sometimes it is different based on what user is logged in or what role they have, which I don't really have an explanation for.

The first few days I was able to cause an author record to show up in the results by editing its first or last name, but I'm having trouble getting even this to work after adding a new author. The newly added author does show up in the results.

I first noticed this all happening just last Friday, so I can't really see how it would be related to version 3.5.1 or anything like that. My first thought was that it had to be a hosting issue with VIP, maybe with the cache infrastructure, and maybe it is, but I'm also curious if there's anything I could investigate within Co-Authors-Plus that could solve this.

lschuyler:
Hi @jonathanstegall,

Is this happening with the block editor? Or the classic editor? Or both?

jonathanstegall:
I haven't used the block editor very heavily, so it may only be the classic editor.

jonathanstegall:
Ultimately I've learned through VIP support that my issue is a duplicate of #622.

jonathanstegall:
After looking into the fix that's on #622, I might be experiencing something different as neither of the filters mentioned as solutions caused any change for me.

jonathanstegall:
In case it's helpful, I've found a way that does allow me to get results for the author Ajax search:

```php
if ( ! function_exists( 'minnpost_coauthors_search_authors_get_terms_args' ) ) :
	add_filter( 'coauthors_search_authors_get_terms_args', 'minnpost_coauthors_search_authors_get_terms_args', 10, 1 );
	function minnpost_coauthors_search_authors_get_terms_args( $args ) {
[...]

PR #1342, opened 2026-08-09
refactor(cache): centralize cache key builders

faisalahammad:
## Description

Extracts a `Automattic\CoAuthorsPlus\Cache\Keys` helper so the cache group string and key templates live in one place. Author-term cache identity stays on `user_nicename` so a `user_nicename` change can no longer orphan the cached entry, and any future re-key attempt that reuses a `WP_User->ID` (which collides with the guest-author CPT `->ID`) is now blocked by a regression test.

This PR covers items 1 and 2 of the seven hardening items in #1313. The other five (caching `search_authors()`, fixing the N+1 in `get_all_linked_accounts()`, group and persistence docs, the WP-CLI cache delete group bug at `class-wp-cli.php:896`, and `get_coauthors()` term batching) ship as follow-up PRs against the same helper so the structural keystone lands first.

### What changed

- New `Automattic\CoAuthorsPlus\Cache\Keys` helper (`php/cache/class-keys.php`) with the unified `GROUP` constant and pure per-purpose key builders (`author_term_key()`, `coauthors_post_key()`, `all_linked_accounts_key()`, `guest_author_key()`). `author_term_key()` is a pure builder, no `get` / `set` / `delete` wrappers — call sites use the standard `wp_cache_*` functions against `Keys::GROUP`. The `guest_author_key()` builder absorbs the `post_name` / `login` / `user_login` field normalization that used to live in `CoAuthors_Guest_Authors::get_cache_key()`.
- Unify both cache groups under the existing `'co-authors-plus'` literal. `CoAuthors_Guest_Authors::$cache_group` and `get_cache_key()` are kept as deprecated one-line aliases pointing at the helper, so any external integration that still touches them keeps working until a future major.
- Keep `get_author_term()` keyed on `user_nicename`. The on-disk term slug is unchanged, so `get_term_by()` lookups stay on the nicename. Added an `empty( $coauthor->ID )` early return so a missing ID cannot pass the object guard, plus a new `\stdClass` non-object guard assertion in `AuthorTermsTest`.
- Add `user_register` and `set_user_role` action handlers in `CoAuthors_Plus` that resolve the user ID via `get_userdata()` and drop the nicename-keyed author-term entry. Modelled on the existing `profile_update` handler: cache delete only, no term creation.
- Add a regression test (`test_author_term_key_avoids_cross_table_id_collision`) that asserts a WP user and a guest author with equal numeric IDs but different nicenames still produce distinct cache keys, so any future re-key to `user_id` fails loud here.

### Why

The cache group string and key templates were duplicated as bare literals across 16 call sites in three files. That drift was the structural root cause of past invalidation bugs (for example #849), and the nicename-based key already matches the on-disk term slug, which is the only stable cross-table identity: `WP_User->ID` (from `wp_users`) and guest-author `->ID` (from the guest-author CPT) come from independent auto-increment sequences and can collide, which would silently return the wrong cached term. #1312 added the `profile_update` hook; `user_register` and `set_user_role` were still missing. Fixing the structural keystone first means the follow-up PRs can plug into the same helper instead of patching the same duplication again.

## Deploy Notes

No new dependencies, no schema changes, no upgrade routine needed. The deprecated BC aliases on `CoAuthors_Guest_Authors` keep third-party code working.

## Steps to Test

1. Check out PR.
2. Run `composer test:unit`, `composer test:integration`, `composer cs`, `composer lint`, `npm run build`. All should pass.
3. In `wp-env` shell, run the seven spot checks in the local `TESTING_INSTRUCTIONS.md` (not committed): confirm `Keys::GROUP === 'co-authors-plus'` and `CoAuthors_Guest_Authors::$cache_group === Keys::GROUP`, confirm `get_author_term()` keys on `user_nicename`, confirm `do_action( 'user_register', $id )` drops the nicename-keyed entry, confirm `$user->set_role( 'editor' )` drops it, confirm `wp_update_user` keeps the entry warm (handler re-warms via `update_author_term`), confirm `Keys::guest_author_key( 'post_name', 'cap-jane' ) === Keys::guest_author_key( 'user_nicename', 'jane' )`, confirm `Keys::all_linked_accounts_key() === 'all-linked-accounts'`.

### Code refactor

- Introduce `Automattic\CoAuthorsPlus\Cache\Keys` as the single source of cache group constant and key templates.
- Drop the old `CoAuthors\Cache\Store` helper. `CoAuthors_Guest_Authors::$cache_group` and `get_cache_key()` remain as BC aliases that delegate to `Keys`.

### Bug fixes

- Drop the author-term cache entry on `user_register` and `set_user_role` (was missing for issue #1313).

### Tests

- Add `KeysTest` covering group, key templates, guest-author field normalization, BC alias equivalence, and a cross-table ID collision regression.
- Cover `user_register` and `set_user_role` invalidation in `AuthorTermInvalidationTest`, and the `\stdClass` non-object guard in `AuthorTermsTest`.


faisalahammad:
## CI fix: PHPCS NoPaging violation in `CreateGuestAuthorTest`

The "Basic CS and QA checks" job was failing on this PR:

```
FILE: tests/Integration/GuestAuthors/CreateGuestAuthorTest.php
108 | ERROR | Setting `numberposts` to `-1` disables pagination and is
140 |       | prohibited in VIP context.
      |       | (WordPressVIPMinimum.Performance.NoPaging.posts_per_page_numberposts)
```

### Fix
Replaced both `get_posts( array( 'numberposts' => -1 ) )` count calls with an exact, VIP-compliant count using `wp_count_posts()`:

```php
$count_before = array_sum( (array) wp_count_posts( $guest_author_obj->post_type ) );
```

`wp_count_posts()` returns per-status counts; `array_sum` over them equals `get_posts` with `post_status => 'any'`, with no pagination. Same for `$count_after`. The test intent (assert no orphaned guest-author post remains after term-creation failure) is unchanged.

### Verification
- `composer cs -- tests/` — 0 errors (matches the failing CI command)
- `composer test:integration --filter CreateGuestAuthorTest` — 4 tests pass
- All other CI checks (Unit Tests, Behat, Playwright) already green

Test-only change; no production code touched.

GaryJones (changes requested):
Thanks for this, Faisal. The core idea is sound and I would like to see it land. The group string and key templates genuinely were duplicated across three files, and centralising them is worth doing. Two things need to change before I can approve, and the first is a blocker.

## 1. Re-keying `get_author_term()` on the ID is a regression

`get_author_term()` is called with a co-author object that is either a `WP_User` **or** a guest author. For a guest author, `->ID` is the guest-author CPT **post** ID (`php/class-coauthors-guest-authors.php:1115`), not a user ID. `wp_posts` and `wp_users` are independent auto-increment sequences, so equal integers occur naturally on any site.

Both now build the same key (`author-term-{ID}`) in the same cache group. So an unrelated guest author and WP user that happen to share an integer collide, and whichever reads second gets the other's term. Their nicenames differ, so these really are different terms.

I reproduced this rather than just reading it:

```php
$guest_id = (int) $this->create_guest_author( 'collide_guest2' );

// Walk the user sequence up until a user lands on the guest author's post ID.
$max_user_id = (int) $wpdb->get_var( "SELECT MAX(ID) FROM $wpdb->users" );
for ( $i = $max_user_id + 1; $i <= $guest_id; $i++ ) {
    $user = $this->create_author( 'collider_' . $i );
}
$this->assertSame( $guest_id, (int) $user->ID );

$guest_author = $coauthors_plus->get_coauthor_by( 'user_login', 'collide_guest2' );
$coauthors_plus->update_author_term( $user );

$user_term  = $coauthors_plus->get_author_term( $user );
$guest_term = $coauthors_plus->get_author_term( $guest_author );

$this->assertNotSame( $user_term->term_id, $guest_term->term_id );
```

On this branch it fails, with the guest author read returning the WP user's term:

```
✘ Guest author and user with equal ids collide
  Failed asserting that 4 is not identical to 4.
```

On `develop` the same test passes. So this is a regression the PR introduces, not something pre-existing. The failure is silent and returns wrong data, and the term feeds byline filtering, post counts, `user_has_cap` and the searchable description.

The wider point is that the nicename key was not an accident. Term identity *is* the nicename slug (`cap-{nicename}`), so keying on the nicename kept the cache key collinear with the thing being cached: a guest author and its linked user correctly shared one entry, and unrelated co-authors always got distinct keys. A bare integer breaks that alignment.

The signature is where it shows. `author_term_key( int $user_id )` is named for a user ID, but every call site passes `$coauthor->ID`, which may be either kind. The type cannot express the distinction, so the collision was invisible at design time. The design issue and the bug are the same issue.

I would also push back gently on the scoping. The PR is framed as the structural keystone for the follow-ups, but items 1 and 2 of #1313 do not require changing the key basis at all. Centralising the group and key templates, and adding the missing `user_register` / `set_user_role` invalidation, can be done with no behavioural change whatsoever. That would be a genuinely safe refactor and a straightforward approve.

So, in order of preference:

1. Drop the re-key from this PR and keep the builder on the nicename. Land the centralisation and the new hooks as pure structure.
2. If the nicename-orphan-on-rename problem is worth solving now, do it as a separate PR that puts the namespace in the key *and* in the type signature: separate builders for users and guest authors, or a required type argument, plus a regression test covering the collision.

Worth weighing the trade too. The orphan problem this was meant to fix is a cache miss followed by a rebuild, whereas the collision returns silently wrong data. As it stands we would be trading a benign failure for a harmful one.

## 2. Shape of the new class

No objection to the extraction itself, and static is the right call for pure key builders. A few changes to how it is packaged.

**Drop `get()` / `set()` / `delete()`.** They have no production callers on this branch. All of the call sites still do `wp_cache_*( $key, Store::GROUP )` directly, and the only consumers are the tests written to cover them. They also do not add the abstraction they appear to: WordPress already gives us a swappable object cache through the `object-cache.php` drop-in, so a static passthrough over a global static function is a no-op layer. Better not to ship public surface we then have to support.

**Rename to `Keys`.** Once the wrappers are gone the class does not store anything; it holds a constant and builds keys. I suspect `Store` is what made the wrappers feel necessary in the first place.

**Make it `final` with a private constructor.** The docblock says it holds no instance state, but nothing enforces that, so `new Store()` works and it can be subclassed. For a shipped utility that is BC surface we did not mean to open.

**Namespace and location.** Please use `Automattic\CoAuthorsPlus\Cache` rather than `CoAuthors\Cache`. Both roots exist in the tree for historical reasons, but `Automattic\CoAuthorsPlus\` is the direction of travel, so new files should start there. Also move the file to `php/cache/class-keys.php` so the directory mirrors the namespace, consistent with `php/blocks/`, `php/api/endpoints/` and `php/integrations/yoast/`.

**Strict standards.** Please keep `declare( strict_types=1 )`. Happy for new files to be strict from the outset.

Taken together the class becomes constants plus pure functions, which is what it should be, and static then becomes unarguably correct.

Thanks again, this is a worthwhile problem to be attacking and I am keen to get the structural part merged.

faisalahammad (commented):
Thanks Gary — both blocker and shape feedback are now addressed in bf08833.

**1. Re-key regression (blocker)** — took your option 1. `author_term_key()` is back on `user_nicename` and the key template stays `author-term-{nicename}`, so the key stays collinear with the term slug (`cap-{nicename}`) and user/guest authors sharing a nicename still share one entry. No behavioural change to cache identity. The collision you reproduced is now blocked by a regression test rather than by memory: `KeysTest::test_author_term_key_avoids_cross_table_id_collision` asserts equal numeric IDs with different nicenames produce distinct keys, and the class docblock on `php/cache/class-keys.php` records why numeric IDs can never become the key basis. The `empty( $coauthor->ID )` early return stays so a missing ID can't pass the object guard, plus a `\stdClass` guard assertion in `AuthorTermsTest`.

**2. Class shape** — all five points done:
- `get()` / `set()` / `delete()` wrappers deleted; every call site uses `wp_cache_*` directly against `Keys::GROUP`.
- Renamed `Store` → `Keys`, moved to `php/cache/class-keys.php`, namespace `Automattic\CoAuthorsPlus\Cache`.
- `final` class, private constructor, `declare( strict_types=1 )` kept.
- `CoAuthors_Guest_Authors::$cache_group` and `get_cache_key()` remain as deprecated one-line BC aliases delegating to `Keys`.

**Hooks** — `user_register` and `set_user_role` handlers resolve the ID via `get_userdata()` and delete `author-term-{user_nicename}`; no term creation.

**Verification** — unit suite 38 tests/61 assertions OK; integration suite 280 tests/966 assertions OK in wp-env (PHP 8.4); PHPCS on changed files, PHP lint, `git diff --check`, `npm run build` all clean. `composer test:integration-ms` could not be reproduced locally: the multisite run uses the core `multisite.xml` bootstrap where the plugin global is not populated, so the plugin-targeted subset needs the plugin loaded the normal way; no behaviour depends on multisite in this diff.

faisalahammad:
Updated and rebased on current `develop`.

- Centralized cache key builders in `Cache\Keys`.
- Preserved nicename-based author-term keys to avoid user/CPT ID collisions.
- Retained registration and role-change invalidation.
- Added collision regression coverage.

Unit tests, integration tests, lint, PHPCS, and build passed. Multisite verification remains limited by the local core multisite bootstrap.

faisalahammad (commented):
Updated in `506ad67` and merged with current `develop`: nicename-based keys preserved, cache builders centralized, registration/role invalidation retained, and collision coverage added. Checks passed; local multisite bootstrap remains unavailable.

Issue #1056, opened 2024-08-13
Cannot find a way to query for all posts where a specific user is a (co)author)


roland-jungwirth:
I need to set up a custom query (for an Elementor loop grid) that constraints the query to all posts from a specific author and/or co-author. Unfortunately I did not find anything in the code. Obviously this won't work

` $query->set('author', $author_id);`

so how can I achieve this?

faisalahammad:
Co-Authors Plus rewrites author queries to include co-authored posts, but only when it recognizes the query as an author query. It does that in two cases: when `WP_Query::is_author()` is true (author set at parse time, e.g. `new WP_Query( array( 'author' => $id ) )`), or when `author__in` is a non-empty array.

In an Elementor custom-query callback the query is already parsed, so setting `author` there does not set `is_author()` and the rewrite is skipped. Use `author__in` instead:

```php
function my_coauthor_query( $query ) {
	$query->set( 'author__in', array( 123 ) );
}
add_action( 'elementor/query/my_coauthor_query', 'my_coauthor_query' );
```

Co-Authors Plus reads `author__in` directly from the query vars, so it detects the query regardless of when you set it, and the returned posts include those where the user is either the primary author or a co-author.

For a standalone query built from scratch, the plain `author` argument also works because it is present at parse time:

```php
$query = new WP_Query(
	array(
		'author' => $author_id,
	)
);
```

Documentation for this is on the way in PR #1345 (`docs/querying-posts.md`), and the behavior is covered by `tests/Integration/Queries/AuthorQueriesTest.php`, including the exact "user is a co-author, not the post author" case.

PR #1345, opened 2026-08-11
Docs: add guide for querying posts by author or co-author

faisalahammad:
## Description

Addresses the discoverability gap behind issue #1056, which asks how to run a `WP_Query` that returns all posts where a specific user is the primary author or a Co-Authors Plus co-author.

The behavior already works. CAP registers `posts_where`, `posts_join`, and `posts_groupby` filters that rewrite author queries to include the `author` taxonomy, and `tests/Integration/Queries/AuthorQueriesTest.php` covers every query form. What is missing is in-repository documentation showing developers how to use it.

This PR:

- Adds `docs/querying-posts.md`, a developer guide covering `author`, `author__in`, comma-separated `author`, `author_name`, guest authors, supported post types, and the two opt-out filters.
- Links the new page from the README developer-reference list.
- Documents the `coauthors_plus_is_author_query` filter in `docs/filters.md`.
- Tightens the `coauthors_plus_should_query_post_author` wording.
- Corrects the documented `coauthors_posts_link` argument keys (the filter passes an associative array with `before_html`, `href`, `rel`, `title`, `class`, `text`, `after_html`, not `text`, `title`, `url`).

The Elementor example uses `author__in` because it is the explicit form that CAP reads directly from the query vars. A correction to my earlier wording: setting `author` late in a `pre_get_posts` style hook does still trigger the rewrite. `is_author()` stays false because it is only set during `parse_query()`, but `WP_Query::get_posts()` expands a non-empty `author` into `author__in` before the SQL filters run, so CAP picks it up through the multi-author path. Integration tests now cover this (added in 5b2faec): late `author` and late `author__in` both return the co-authored post. The guide now explains this instead of claiming that a late `author` misses co-authored posts.

## Deploy Notes

No new dependencies. Documentation-only change; no runtime code touched.

## Steps to Test

1. Check out the branch.
1. Confirm the following pass:
   - `composer lint` (PHP syntax)
   - `npx wp-scripts lint-md-docs docs/querying-posts.md docs/filters.md` (Markdown)
1. Run the integration suite to confirm the documented query forms behave as described:
   - `composer test:integration -- --filter AuthorQueriesTest` (16 tests, all pass)
1. Open `docs/querying-posts.md` and verify each documented query form matches the behavior asserted in `tests/Integration/Queries/AuthorQueriesTest.php`.
1. Optional: verify the Elementor paragraph by adding a `pre_get_posts` callback that calls `$query->set( 'author', $coauthor_id )` on a fresh `WP_Query` and confirm it returns posts where that user is only a co-author.


GaryJones (approved):
Approving in principle — the content is accurate and the code clearly informed it.

I verified each substantive claim against the source rather than the description:

- `coauthors_plus_is_author_query` exists at `php/class-coauthors-plus.php:928` with the documented `bool $is_author, WP_Query $query` signature.
- `coauthors_plus_should_query_post_author` is real, and the reworded "keeps the rewrite but switches to taxonomy-only" description matches the branches at lines 278/1069/1248.
- The `coauthors_posts_link` correction is right: the actual keys in `template-tags.php:286-293` are `before_html`, `href`, `rel`, `title`, `class`, `text`, `after_html`. The old docs' `url` never existed, it's `href`.
- The Elementor `author__in` reasoning holds up against `is_author_query()` (lines 909-916): a late `$query->set('author', …)` runs after `parse_query()`, so `is_author()` is already false and only `author__in` will trigger the rewrite. That's a genuinely non-obvious point and it's documented correctly.
- The query forms line up with `AuthorQueriesTest.php`, and this maps directly onto issue #1056's Elementor loop-grid case.

`wp-scripts lint-md-docs` passes clean on both files.

One non-blocking point before merge:

**One clarity trap in the guide itself.** We tell readers a plain `'author' => 123` query returns co-authored posts (true, and the tests confirm it), then tell them the Elementor case *must* use `author__in`. Both are correct, but the guide never states *why* the same argument behaves differently in the two places. A single sentence on parse-time versus setting the var late would stop a reader concluding `author` is simply unreliable.

faisalahammad (commented):
Thanks for the review. I added the parse-time explanation in 2198f53, but while checking it I found that my earlier reasoning (and the reasoning in the PR body) was wrong, so I want to be upfront about it.

A late `$query->set( 'author', 123 )` in a `pre_get_posts` style hook does still trigger the CAP rewrite. `is_author()` stays false because it is only set in `parse_query()`, but `WP_Query::get_posts()` expands a non-empty `author` into `author__in` before `posts_where` fires. `is_author_query()` then sees a non-empty `author__in` and the query takes the multi-author path. The plugin's own comments around lines 942 and 1026 already describe this expansion.

I confirmed it with a throwaway integration test (not committed): a `pre_get_posts` callback setting `author` to a user who is only a co-author returned the post, with `is_author()` false. Setting `author__in` late returned it too.

So the new paragraph in `docs/querying-posts.md` now says both forms work, explains the flag timing, and recommends `author__in` because it is the explicit form CAP reads directly and must be an array of integer IDs. I also updated the PR description to remove the wrong claim.

One more small change: the `coauthors_plus_is_author_query` example now targets a single query and returns the original value for all others, instead of returning `false` for everything.

Please take another look when you have time.

GaryJones (changes requested):
Thanks Faisal — this addresses my review point, and I've verified the new paragraph empirically rather than taking it on trust. Test setup: a post primary-authored by user A with user B as co-author only, and a `pre_get_posts` callback setting the author vars late. Results: a late plain `author` **does** trigger the rewrite via the multi-author path, a string `author__in` silently degrades to primary-author-only matching, and an array `author__in` works. All three match what the paragraph says, so I'll happily note my original review had this backwards — WordPress's expansion of `author` into `author__in` before the SQL filters run is the piece I missed, and it's exactly the right thing to document.

Two changes requested before merge:

1. The PR description still says a late `author` "does not trigger the rewrite and would miss co-authored posts", which the guide now (correctly) contradicts. Could you update the description to match?
2. The guide now asserts behaviour that has no test coverage: `AuthorQueriesTest` only passes author vars via the `WP_Query` constructor, never via `pre_get_posts`. A small test that sets `author` and `author__in` late (mirroring the Elementor path from #1056) would pin down the behaviour this page documents.

Markdown lint passes on both updated files. The scoped `coauthors_plus_is_author_query` example is a nice improvement over the blanket opt-out.

faisalahammad:
Thanks for the second review, and for verifying the paragraph empirically.

Both requested changes are done:

1. The PR description no longer contains the outdated claim about a late `author` query. It was corrected before your review landed, so you may have been reading a cached version. The description now says the opposite: setting `author` late in a `pre_get_posts` style hook does still trigger the rewrite, because `WP_Query::get_posts()` expands a non-empty `author` into `author__in` before the SQL filters run.

2. I added the requested integration tests in 5b2faec. `tests/Integration/Queries/AuthorQueriesTest.php` now has a section that mirrors the Elementor path from #1056: a `pre_get_posts` callback sets `author` late on a running query in one test and `author__in` in the other, and both assert the co-authored post is returned. The `author` test also asserts `is_author()` stays false. The class now runs 16 tests.

One detail worth noting: `new WP_Query()` with no arguments never runs a query at all, so a callback attached to it never fires. The tests construct the query with `post_type` set, which is how a real Elementor custom query behaves.

PR #1346, opened 2026-08-11
fix(coauthors): preserve input order in add_coauthors()

### LINKED ISSUE #1052: Migrating from custom system, and not able to keep the order in authors. Wrong sorting.
Hello,

We are currently migrating from another custom system to co-authors.
We are having a lot of trouble with the migration since the ordering of authors seems to be completely random. 
Sometimes (most of the time) respects the old order, but sometimes it doesn't. 
Maybe the way we are doing it it's wrong (?)

How are we doing it? 
Passing an array of authors (with user_nicename) to add_coauthors function.

```
global $coauthors_plus;
$new_co_authors = [ 'Jhon Doe', 'Test User'];
$coauthors_plus->add_coauthors( $post_id, array_map( 'sanitize_title', $new_co_authors ) );
```

Sometimes we will get co-authors in the right other, sometimes inverse.

Why the order is sometimes not respected? I tried other combinations, like calling add_coauthors with just one author, n times. Nothing seems to work, and all the time seems random.

Any clues? Are we doing it wrong?

Thanks,

faisalahammad:
## Description

Co-Authors Plus reads the co-authors for a post ordered by `wp_term_relationships.term_order ASC`, but `add_coauthors()` never wrote that column. After `wp_set_post_terms()` every relationship row was left at the default `term_order = 0`, so the read path fell back to an arbitrary order based on `term_taxonomy_id`. Programmatic callers (e.g. `$coauthors_plus->add_coauthors( $post_id, $authors )`) saw a random author order, and reordering in the editor drifted too, because the editor also funnels through `add_coauthors()` via `sync_coauthors_on_rest_save`.

This fix writes `term_order` for each author inside `add_coauthors()` so the stored order matches the input array, then clears the `coauthors_post_{ID}` cache so the next read sees the new order. It reuses the `term_taxonomy_id` already returned by `update_author_term()` instead of making an extra `get_term_by()` query per author.

Fixes #1052

## Deploy Notes

No new dependencies. Database schema unchanged; this only writes the existing `term_order` column on `wp_term_relationships`.

## Steps to Test

1. Check out PR.
2. Create a post with three existing users and assign them in a specific order via `$coauthors_plus->add_coauthors()`.
3. Confirm `get_coauthors()` returns them in the exact order passed in.
4. Call `add_coauthors()` again with a different order and confirm the order updates.
5. Reorder authors in the block editor, save, and confirm the order persists on reload.
6. Run the integration test suite with `composer test:integration` (all 270 tests should pass, including the new `AddCoauthorsTest::test_add_coauthors_preserves_input_order`).

### DIFF
--- a/php/class-coauthors-plus.php
+++ b/php/class-coauthors-plus.php
@@ -1662,6 +1662,7 @@ class CoAuthors_Plus {
 		// Set the co-authors
 		$coauthors        = array_unique( array_merge( $existing_coauthors, $coauthors ) );
 		$coauthor_objects = array();
+		$coauthor_terms   = array();
 		foreach ( $coauthors as &$author_name ) {
 			if ( $this->is_rest_save && has_filter( 'coauthors_post_get_coauthor_by_field' ) ) {
 				_deprecated_hook(
@@ -1676,11 +1677,39 @@ class CoAuthors_Plus {
 			$author             = $this->get_coauthor_by( $field, $author_name );
 			$coauthor_objects[] = $author;
 			$term               = $this->update_author_term( $author );
-			if ( is_object( $term ) ) {
+			$has_term           = is_object( $term ) && ! is_wp_error( $term );
+			$coauthor_terms[]   = $has_term ? $term : null;
+			if ( $has_term ) {
 				$author_name = $term->slug;
 			}
 		}
+		unset( $author_name );
 		wp_set_post_terms( $post_id, $coauthors, $this->coauthor_taxonomy );
+
+		// Persist the requested order in wp_term_relationships.term_order so that
+		// get_coauthor_terms_for_post() (which orders by term_order ASC) returns
+		// the authors in the same order they were passed in. Without this, every
+		// row gets term_order = 0 and the read path falls back to an arbitrary
+		// order based on term_taxonomy_id. See issue #1052.
+		foreach ( $coauthors as $order => $slug ) {
+			$term_taxonomy_id = isset( $coauthor_terms[ $order ] ) && $coauthor_terms[ $order ]
+				? (int) $coauthor_terms[ $order ]->term_taxonomy_id
+				: 0;
+			if ( ! $term_taxonomy_id ) {
+				continue;
+			}
+			$wpdb->update(
+				$wpdb->term_relationships,
+				array( 'term_order' => (int) $order ),
+				array(
+					'object_id'        => $post_id,
+					'term_taxonomy_id' => $term_taxonomy_id,
+				),
+				array( '%d' ),
+				array( '%d', '%d' )
+			);
+		}
+		$this->clear_cache( $post_id );
 
 		// If the original post_author is no longer assigned,
 		// update to the first WP_User $coauthor
--- a/tests/Integration/Bylines/AddCoauthorsTest.php
+++ b/tests/Integration/Bylines/AddCoauthorsTest.php
@@ -873,6 +873,61 @@ class AddCoauthorsTest extends TestCase {
 	}
 
 	/**
+	 * Confirms that add_coauthors() persists the order of the input array via
+	 * wp_term_relationships.term_order, and that get_coauthors() (which orders
+	 * by term_order ASC) returns the authors in that same order. See issue #1052.
+	 *
+	 * Also covers a re-call with a different order to prove the term_order
+	 * write actually overwrites the previously persisted value, rather than
+	 * only writing on the first call.
+	 *
+	 * @covers ::add_coauthors
+	 */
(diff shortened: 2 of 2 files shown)


GaryJones (commented):
Thanks for this, and sorry it has sat for a while. I have rebased it locally onto `develop` and dug into it properly. The rebase is clean and does fix CI — the only red check was `Basic CS and QA checks`, failing on two pre-existing `WordPressVIPMinimum.Performance.NoPaging` errors in `CreateGuestAuthorTest.php` that were unrelated to your change and have since been fixed on `develop` by #1364.

Unfortunately I do not think the fix itself is needed, and I would like to lay out the evidence rather than just assert it.

### WordPress core already writes `term_order` here

`wp_set_object_terms()` writes `term_order` in input order for any taxonomy registered with `'sort' => true`:

```php
// wp-includes/taxonomy.php (WP 6.4, our minimum supported version)
if ( ! $append && isset( $t->sort ) && $t->sort ) {
    ...
    $values[] = $wpdb->prepare( '(%d, %d, %d)', $object_id, $tt_id, ++$term_order );
    ...
    'INSERT INTO ... ON DUPLICATE KEY UPDATE term_order = VALUES(term_order)'
}
```

The `author` taxonomy is registered with `'sort' => true` in `action_init_late()` (in place since September 2023), and `add_coauthors()` always calls `wp_set_post_terms()` in **non-append** mode — it does its own merging of the existing and incoming lists first — so core's `! $append` guard is always satisfied. Every other place the plugin sets author terms is non-append too.

I verified this rather than trusting the reading: I reverted the production hunk from this branch, kept your new test, and ran it against plain `develop`. **`test_add_coauthors_preserves_input_order` passes without the fix.** Dumping `wp_term_relationships` directly showed `term_order` values of `1, 2, 3` in input order, already written by core. The new loop then overwrites them with `0, 1, 2` — different numbering, identical sort semantics.

### The loop has a latent bug, which is itself the clue

`array_unique()` preserves keys, so the merged list can have gaps:

```php
$coauthors = array_unique( array_merge( $existing_coauthors, $coauthors ) ); // keys 0, 1, 3
foreach ( $coauthors as &$author_name ) { $coauthor_terms[] = ...; }         // keys 0, 1, 2
foreach ( $coauthors as $order => $slug ) {
    if ( ! isset( $coauthor_terms[ $order ] ) ) { continue; }                // $order 3 → skipped
```

Any mid-list duplicate — for example `add_coauthors( $id, [ 'b', 'd' ], true )` where `b` is already assigned — leaves a key gap, and the final author never gets its `term_order` written. `array_values()` would fix it.

The telling part is that I wrote a test for exactly that case and it passed, because core had already written the correct value the loop failed to write. That is fairly strong evidence the loop is not doing any load-bearing work.

For completeness, if this were reworked: the raw `$wpdb->update()` also bypasses core's term caching. `clear_cache()` only deletes the plugin's own `coauthors_post_{ID}` key, not `wp_cache_set_terms_last_changed()`, so on a persistent object cache any `WP_Term_Query` entry primed between `wp_set_post_terms()` and the raw update could go stale with nothing to invalidate it.

### What I have done with it

Your work was not wasted, and it prompted a genuinely useful change. I have opened #1368, which:

- Keeps your `test_add_coauthors_preserves_input_order` as a characterisation test. It passes on `develop` today, but it pins a contract we rely on and do not control, living entirely in two registration arguments that look inert and could easily be tidied away. I confirmed it fails if `'sort' => true` is removed, so it earns its place.
- Adds a second test for the append case above, so the `array_unique()` trap is documented rather than rediscovered.
- Keeps your `is_wp_error()` guard, which is a real fix. `update_author_term()` returns a `WP_Error` on `wp_insert_term()` failure, a `WP_Error` is an object, and reading `->slug` off it yields `null`, which `wp_set_post_terms()` then drops — silently losing the author from the byline.
- Keeps your `unset( $author_name )` after the by-reference loop.

### On #1052

The issue is from June 2024, well after `'sort' => true` landed, so the reported random ordering has some other cause and the write path can now be ruled out. I will follow up on the issue itself with what we have established and where I would look next. It should stay open.

Given the above, I would suggest we close this PR in favour of #1368, but I wanted to show the working first rather than close it out from under you. Happy to hear if you think I have got any of this wrong — and thank you for the careful write-up in the description, which is what made it quick to check.

faisalahammad:
Thanks for the deep check on this. You are right, `'sort' => true` on the author taxonomy already makes core write term_order, so the loop here is not needed. Good catch on the array_unique key gap too, that is a real bug in my loop.

Closing this in favor of #1368, it keeps the useful part (the is_wp_error guard) and adds the right tests.

Leaving #1052 open since the write path is not the cause.

PR #1347, opened 2026-08-11
Add rest_coauthors_prepare_items filter to Co-Authors REST endpoint

### LINKED ISSUE #1051: Override "Co-Authors" blocks to only render primary author
This isn't really an "issue" so much as it is a question, but I didn't know where else to post this.

I would like to use the "Co-Authors" block to display author information under a post - but not for all co-authors, only for the primary author, i.e. the first one. I don't know which functions I need to hook filters or actions in to change the result from all co-authors to only one. I could directly change the `render_block()` functions of the blocks or the `Blocks::get_authors_with_api_schema()` function which gets the author data. But then my changes will be overridden with the next update, so I would like to do it with an external override.

I'm new to WP so if there is an obvious answer for how to find out which functions can be hooked into, please let me know.

faisalahammad:
## Description

The Co-Authors block reads co-authors through the `coauthors/v1/coauthors` REST endpoint. There was no filter on the resolved list at that layer, so a site that wanted to render only the primary author had to edit plugin code, which does not survive updates.

This PR adds a new filter, `rest_coauthors_prepare_items`, between `get_coauthors()` and the REST response formatting. Consumers can use it to limit, reorder, or replace the list of co-authors returned by the endpoint. The Co-Authors block's server render and its editor preview both read from the same endpoint, so both follow whatever the filter produces; headless clients and other plugins that query the endpoint see the same list.

Default behavior is unchanged when no callback is registered. The filter also receives `$post_id` and the full `WP_REST_Request`, so callbacks can branch on request context (params, headers, route) the same way core's `rest_prepare_*` filters do.

This is presentation filtering, not access control: co-authors removed here still exist on the post and remain visible through `get_coauthors()`, the byline template tags, the post's `coauthors` REST field, and the Authors panel in the editor sidebar. Use the existing `get_coauthors` filter if the change should apply everywhere the plugin resolves co-authors, not just this REST endpoint.

Closes #1051

### Incidental fix

The new guard at the filter call site reindexes the array with `array_values()` before serialization. `get_coauthors()` returns `array_unique( $coauthors, SORT_REGULAR )`, which can leave non-sequential keys when the same co-author appears under more than one term, and `array_map()` preserves keys, so a post with duplicate co-author objects previously serialised as a JSON object rather than a JSON array. A new test (`test_duplicate_coauthors_response_is_reindexed`) covers the fix.

## Deploy Notes

No new dependencies. No changes to existing REST schema, permission checks, or block attributes. Existing consumers of the endpoint see identical responses unless they register the new filter.

## Steps to Test

1. Check out this branch on a WordPress site with Co-Authors Plus active and a post that has at least two co-authors.
2. Confirm the Co-Authors block still renders every co-author in the existing order (no filter registered, default path).
3. Add a small mu-plugin that registers the filter:
   ```php
   add_filter( 'rest_coauthors_prepare_items', static function ( $coauthors ) {
       return array_slice( $coauthors, 0, 1 );
   } );
   ```
4. Reload the post on the front end. Only the first co-author renders.
5. Open the same post in the block editor and confirm the block preview also shows only the first co-author.
6. Deactivate the mu-plugin, reload, and confirm every co-author renders again.
7. Run `composer test:integration` against this branch. The new file `tests/Integration/Rest/RestCoauthorsPrepareItemsFilterTest.php` adds 9 tests covering the default path, the limit-to-primary case, the `$post_id` and `WP_REST_Request` arguments, non-array return coercion, the `is_coauthor()` guard dropping invalid objects, the real `Blocks::get_authors_with_api_schema()` block render path (with pretty permalinks), and the duplicate-key reindexing behaviour.

Expected result: steps 1-2 unchanged, steps 3-5 limited to primary author, step 6 reverted, step 7 green.

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -46,6 +46,13 @@ The resolved co-authors for a post. Runs after the plugin's lookup but before an
 - **Parameters:** `array $coauthors`, `int $post_id`
 - **File:** `template-tags.php`
 
+### `coauthors_block_authors`
+
+The resolved co-authors for a post, evaluated inside the `coauthors/v1/coauthors` REST endpoint. Use this to limit, reorder, or replace the list of co-authors that the Co-Authors block renders on the front end. The list is also exposed to the block editor via the same endpoint, so editor previews stay in sync. Returning a non-array from the callback yields an empty response.
+
+- **Parameters:** `array $coauthors`, `int $post_id`, `string $context`
+- **File:** `php/api/endpoints/class-coauthors-controller.php`
+
 ### `coauthors_open_graph_tags`
 
 The Open Graph tags the plugin contributes to the document head.
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -46,13 +46,6 @@ The resolved co-authors for a post. Runs after the plugin's lookup but before an
 - **Parameters:** `array $coauthors`, `int $post_id`
 - **File:** `template-tags.php`
 
-### `coauthors_block_authors`
-
-The resolved co-authors for a post, evaluated inside the `coauthors/v1/coauthors` REST endpoint. Use this to limit, reorder, or replace the list of co-authors that the Co-Authors block renders on the front end. The list is also exposed to the block editor via the same endpoint, so editor previews stay in sync. Returning a non-array from the callback yields an empty response.
-
-- **Parameters:** `array $coauthors`, `int $post_id`, `string $context`
-- **File:** `php/api/endpoints/class-coauthors-controller.php`
-
 ### `coauthors_open_graph_tags`
 
 The Open Graph tags the plugin contributes to the document head.
@@ -201,6 +194,19 @@ Schema for a single co-author resource returned by the REST controller.
 The `WP_REST_Response` for a single co-author, after the default preparation. Use this to add custom fields.
 
 - **Parameters:** `WP_REST_Response $response`, `object $author`, `WP_REST_Request $request`
+
+### `rest_coauthors_prepare_items`
+
+The resolved co-authors for a post, before the `coauthors/v1/coauthors` collection endpoint formats them for a REST response. Use this to limit, reorder, or replace the list that endpoint returns. The Co-Authors block's server render and the block editor both read from this endpoint, so their output follows the filtered list; headless clients and other plugins that query the same endpoint see it too.
+
+This is presentation filtering only: co-authors removed here still exist on the post and remain visible through `get_coauthors()`, the byline template tags, the post's `coauthors` REST field, and the Authors panel in the editor sidebar. Do not use this filter for access control.
+
+When you need to change the co-authors everywhere the plugin resolves them, filter `get_coauthors` instead — it runs earlier and applies to template tags and archives as well. Reach for `rest_coauthors_prepare_items` only when the change should affect the REST collection response.
+
+Returning a non-array from the callback yields an empty response. Entries that are not valid co-author objects are dropped, and the surviving entries are reindexed so the response is always a well-formed JSON array.
+
+- **Parameters:** `array $coauthors`, `int $post_id`, `WP_REST_Request $request`
+- **File:** `php/api/endpoints/class-coauthors-controller.php`
 
 ## JavaScript filters (`@wordpress/hooks`)
 
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -56,6 +56,8 @@ Returns rich co-author data (display name, avatar, email, etc.) for a given post
 
 Available to anonymous requesters when the post is publicly viewable (`is_post_publicly_viewable()`). Otherwise the requester must have `read_post` permission for that post ID. Tightened in 4.0 to stop draft/private/scheduled posts from leaking author data.
 
+The response array passes through the `coauthors_block_authors` filter (added in 4.2.0) before serialization, which is the supported extension point for limiting, reordering, or replacing the list rendered by the Co-Authors block. See [filters.md](./filters.md).
+
 **Defined in** `php/api/endpoints/class-coauthors-controller.php`.
 
 ### `GET /coauthors/v1/coauthors/{user_nicename}`
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -56,7 +56,7 @@ Returns rich co-author data (display name, avatar, email, etc.) for a given post
 
 Available to anonymous requesters when the post is publicly viewable (`is_post_publicly_viewable()`). Otherwise the requester must have `read_post` permission for that post ID. Tightened in 4.0 to stop draft/private/scheduled posts from leaking author data.
 
-The response array passes through the `coauthors_block_authors` filter (added in 4.2.0) before serialization, which is the supported extension point for limiting, reordering, or replacing the list rendered by the Co-Authors block. See [filters.md](./filters.md).
+The response array passes through the `rest_coauthors_prepare_items` filter (added in 4.2.0) before serialization, which is the supported extension point for limiting, reordering, or replacing the returned list. The Co-Authors block is the main consumer of this endpoint, but the filter applies to every response from it. See [filters.md](./filters.md).
 
 **Defined in** `php/api/endpoints/class-coauthors-controller.php`.
 
--- a/php/api/endpoints/class-coauthors-controller.php
+++ b/php/api/endpoints/class-coauthors-controller.php
@@ -314,17 +314,18 @@ class CoAuthors_Controller extends WP_REST_Controller {
 		 * Filters the resolved co-authors before they are formatted for the REST response.
 		 *
 		 * Allows consumers to limit, reorder, or replace the list returned by the
-		 * coauthors/v1/coauthors endpoint, which is the data source for the Co-Authors
-		 * block's server render. Returning a non-array short-circuits to an empty list.
-		 * Invalid (non-object) entries and sparse keys are dropped/reindexed so the
-		 * response stays a well-formed JSON array.
+		 * coauthors/v1/coauthors collection endpoint. This endpoint is the data source
+		 * for the Co-Authors block's server render and its editor preview, but it is a
+		 * public endpoint also consumed by headless clients and other plugins. Returning
+		 * a non-array short-circuits to an empty list. Invalid entries are dropped and
+		 * sparse keys are reindexed so the response stays a well-formed JSON array.
 		 *
 		 * @since 4.2.0
-		 * @param array  $coauthors Resolved co-author objects (WP_User|stdClass).
-		 * @param int    $post_id  Post ID from the request.
-		 * @param string $context  Always 'block' for now; reserved for future endpoints.
+		 * @param array           $coauthors Resolved co-author objects (WP_User|stdClass).
+		 * @param int             $post_id   Post ID from the request.
+		 * @param WP_REST_Request $request   The current REST request.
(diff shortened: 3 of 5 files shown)


GaryJones (commented):
Thanks for this, @faisalahammad — and sorry for the slow first response. The idea is sound and the approach is the one @douglas-johnson suggested on #1051 ("a filter here in the API controller after the authors are found but before the data is prepared for a REST response"), so we're on the right track. The tests are real and they pass.

I do want a few changes before this goes in, though. Two of them are permanent public API that we can't take back once released, and one is a bug I was able to reproduce.

## CI

The failure is not yours. `composer cs -- tests/` is tripping over pre-existing `WordPressVIPMinimum.Performance.NoPaging` errors at `tests/Integration/GuestAuthors/CreateGuestAuthorTest.php:108` and `:140`, which were fixed on `develop` by #1364. A rebase clears it.

I cherry-picked your commit onto current `develop` to check: `phpcs tests/` is clean, and the full integration suite is **276 passed, 1 skipped**. The branch is 37 commits behind with no conflicts, so the rebase should be uneventful.

## Blocking

### 1. `is_object()` is too weak a guard — `self::is_coauthor()` already exists

The PR sets out to harden this path, but the guard doesn't quite get there. `prepare_item_for_response()` reads `$author->ID`, `->description`, `->display_name` and `->user_nicename` unguarded, so any object survives `is_object()` and then falls over inside. With a filter returning `(object) [ 'foo' => 'bar' ]`:

```
PHP ERRORS: ["Undefined property: stdClass::$ID" x4, "...::$description" x2,
"...::$display_name", "...::$ID", "...::$user_nicename",
"rawurlencode(): Passing null to parameter #1 ($string) of type string is deprecated",
"...::$user_nicename"]
DATA: [{"id":0,"display_name":"","link":"http://...?author_name","user_nicename":"", ...}]
```

Eleven warnings, plus a junk empty author in a 200 response.

`get_item()` on the sibling route already guards with `self::is_coauthor()`, so there's an existing helper to reuse and a consistency argument for using it. Changing the closure to:

```php
return is_object( $author ) && self::is_coauthor( $author );
```

clears every warning and drops the junk entry, with all seven of your tests still green — I verified that locally. Worth noting the `rawurlencode()` deprecation would also trip `convertDeprecationsToExceptions="true"` for anyone who later writes a test against that case.

### 2. The name says "block", but the hook fires for every consumer

This sits on the public `GET /coauthors/v1/coauthors` collection endpoint. Headless front ends, other plugins, the block's server render, and `apiFetch` in `src/blocks/block-coauthors/edit.js` all receive the filtered list. Calling it `coauthors_block_authors` and documenting it as what "the Co-Authors block renders" understates its reach.

`docs/filters.md` already has a **REST API** section containing `rest_coauthors_item_schema` and `rest_prepare_coauthor`, so there's both an established home and a naming convention here — the new entry currently sits under "Display and formatting" instead. Something like `rest_coauthors_prepare_items` would fit, with the block kept as the motivating example in the docs.

I realise #1051 asked specifically about the block, and the block is the main consumer. But that's an argument for documenting the block use case prominently, not for naming the hook after one of its callers. Filter names are forever.

### 3. `$context` is speculative API, and its value isn't accurate

It's hardcoded to `'block'` and documented as "Always 'block' for now; reserved for future endpoints". Two problems: we can never remove it, and the value is simply wrong whenever the caller isn't the block.

If the goal is to give callbacks something to branch on, pass `WP_REST_Request $request` as the third argument instead. It's genuinely useful (params, headers, route), and it matches the shape of core's `rest_prepare_*` filters.

## Should fix

### 4. `test_block_render_path_sees_filtered_list` doesn't test the block render path

It performs the same dispatch as `test_filter_can_limit_to_primary_author`, with a docblock explaining why it isn't testing what its name claims. A test whose own comment says it doesn't test the thing gives false confidence — I'd rather have no test than that.

The stated reason isn't quite right either. `WP_REST_Request::from_url()` returns `false` because under the default plain permalinks `rest_url()` is `?rest_route=`, so `home_url( '/wp-json/...' )` never matches a REST route. It's not the spy server rejecting it.

Add this at the top of the test:

```php
$this->set_permalink_structure( '/%postname%/' );
```

and `Blocks::get_authors_with_api_schema()` dispatches fine. I checked, and it returns the filtered single author. That turns the placeholder into the real test the PR describes.

### 5. The overlap with the existing `get_coauthors` filter isn't addressed

`get_items()` calls `get_coauthors()`, which already fires `apply_filters( 'get_coauthors', $coauthors, $post_id )` at `template-tags.php:54` — the same two arguments, one statement earlier. The new hook is the very next line.

The REST-only versus everywhere distinction is a real one, but neither the PR description nor the docs state it, and for the reporter of #1051 (who only uses the block) `get_coauthors` would already have worked. Could you add a sentence to `docs/filters.md` on when to reach for which? Otherwise we'll see the same question again.

### 6. Make clear this is presentation, not access control

Co-authors removed by this filter are still present in the post's `coauthors` REST field, in `get_coauthors()`, in the byline template tags, and in the Authors panel in the editor sidebar. Someone will read "limit the co-authors" as "hide the co-author". One sentence in the docs would head that off.

### 7. Test annotations don't match the repo's conventions

`@coversNothing` on the class combined with `@covers ::get_items` on the methods is self-contradictory, and `::get_items` without a `@coversDefaultClass` resolves to a global function that doesn't exist.

`tests/Integration/Rest/CoAuthorsControllerTest.php`, in the same directory, already has `@coversDefaultClass \CoAuthors\API\Endpoints\CoAuthors_Controller` plus a `make_request()` helper. These tests would arguably be at home in that file; failing that, please match its annotation style.

## Minor

- `(int) $request->get_param( 'post_id' )` — the route's `sanitize_callback` already returns `absint()`, so the cast is redundant.
- `remove_filter( ..., 10, 2 )` passes a fourth argument that `remove_filter()` doesn't accept; it's silently ignored. `WP_UnitTestCase` also backs up and restores `$wp_filter` between tests, so the removals are belt and braces regardless.
- Silently coercing a non-array return to `array()` is defensible, but a `_doing_it_wrong()` would save someone a long stare at an empty block. Only a mild preference — I can see the argument that it's noisy.

## One thing worth crediting

The `array_values()` call incidentally fixes a latent bug that nobody has mentioned. `get_coauthors()` returns `array_unique( $coauthors, SORT_REGULAR )`, which can leave non-sequential keys, and `array_map()` preserves keys — so a post with duplicate co-author objects currently serialises as a JSON *object* rather than an array. That's worth calling out in the PR description, and a test using a genuine duplicate (rather than a filter that returns a sparse array) would lock the fix in.

---

To summarise what I'd like next: rebase on `develop`, swap `is_object()` for `self::is_coauthor()`, settle the hook name and drop or repurpose `$context`, and make the block-render test real. The rest is documentation polish. Happy to talk through the naming if you'd rather land on it together before you rework anything.

GaryJones (changes requested):
Formally marking this as *Request changes* so it's clear where the PR stands — the detail is in my [previous review](https://github.com/Automattic/co-authors-plus/pull/1347#pullrequestreview-5092150623).

To recap what's needed before merge:

1. **Rebase on `develop`.** The CI failure isn't yours — it's pre-existing `NoPaging` drift fixed by #1364.
2. **Swap `is_object()` for `self::is_coauthor()`** in the filter's guard. Reproduced: a filter returning a bare `stdClass` yields eleven PHP warnings and a junk empty author in a 200 response. The helper already exists and `get_item()` uses it.
3. **Settle the hook name.** It fires for every consumer of the public collection endpoint, not just the block, and `docs/filters.md` already has a REST API section with an established `rest_*` convention.
4. **Drop or repurpose `$context`.** Hardcoded to `'block'` and documented as reserved, which we can never remove. `WP_REST_Request $request` would be more useful and matches core.
5. **Make `test_block_render_path_sees_filtered_list` real.** Adding `$this->set_permalink_structure( '/%postname%/' )` makes `Blocks::get_authors_with_api_schema()` dispatch correctly — verified.

The remaining points (documenting the overlap with `get_coauthors`, the presentation-not-access-control note, test annotations, and the minor items) are polish and won't hold anything up on their own.

No rush on any of this, and do push back if you disagree with the naming call in particular — happy to settle it in discussion before you rework the code.

faisalahammad:
Thanks for the detailed review, Gary. All five blocking items and the polish items are addressed in b4e6e0f, rebased on current develop.

1. **Rebase**: done, clean, no conflicts.
2. **Guard**: the filter now validates with `is_object( $author ) && self::is_coauthor( $author )`, the same predicate `get_item()` uses. I kept the `is_object()` pre-check because `is_coauthor()` calls `property_exists()`, which would throw on a non-object under PHP 8. A new test covers the bare stdClass case.
3. **Hook name**: renamed to `rest_coauthors_prepare_items`, moved to the REST API section of `docs/filters.md`, and documented as endpoint-wide rather than block-specific.
4. **`$context`**: dropped. The third argument is now the `WP_REST_Request` itself, with a test asserting the instance and its `post_id` param.
5. **Block render test**: real now. `set_permalink_structure( '/%postname%/' )` plus a direct call to `Blocks::get_authors_with_api_schema()`, with a comment explaining the plain-permalink caveat.

Also done: the entry now sits in the REST API section with the presentation-not-access-control note and a sentence on when to use `get_coauthors` instead; annotations follow the `@coversDefaultClass` convention; the invalid `remove_filter` fourth argument is gone; and a new test locks in the duplicate co-author reindexing you flagged as worth crediting. The redundant `(int)` cast stayed since the sanitize callback makes it harmless and it documents the expected shape at the call site.

Integration suite: 328 passed, 1 skipped, on PHP 8.4. `composer cs -- tests/` is clean.

faisalahammad (commented):
Rebased on develop, swapped the guard to `is_object() && self::is_coauthor()`, renamed the hook to `rest_coauthors_prepare_items` and moved its docs to the REST API section, replaced the `$context` argument with the `WP_REST_Request`, and made the block render test real via pretty permalinks plus `Blocks::get_authors_with_api_schema()`. Docs now cover the `get_coauthors` overlap and the presentation-not-access-control boundary, annotations follow the repo convention, and a test locks in the duplicate-key reindexing. Details in the PR comment.

GaryJones (approved):
Thanks @faisalahammad — this addresses everything, and thoroughly. Verified locally on top of current `develop`:

- `self::is_coauthor()` guard in place; my bare-`stdClass` reproduction now yields zero warnings, and `test_non_coauthor_object_is_dropped` locks it in.
- `rest_coauthors_prepare_items` with `WP_REST_Request $request` — right name, right section of `docs/filters.md`, and the `get_coauthors` overlap and presentation-only caveats are exactly what I was after.
- The block render test now genuinely exercises `Blocks::get_authors_with_api_schema()`, with an accurate explanation of the permalink requirement.
- `test_duplicate_coauthors_response_is_reindexed` covering the latent sparse-key serialisation bug with a real duplicate is a nice bonus.

`phpcs tests/` clean, full integration suite 328 passed / 1 skipped (the expected build-files skip). CI is green.

One nit: the file is still named `BlockAuthorsFilterTest.php` while the filter it tests is now `rest_coauthors_prepare_items` — something like `RestCoauthorsPrepareItemsFilterTest.php` would age better. 

Approving (pending file name change). :shipit:

faisalahammad:
Done. The test file is renamed to `tests/Integration/Rest/RestCoauthorsPrepareItemsFilterTest.php`, and the class name matches. No production code changed. c7dbdb0

PR #1348, opened 2026-08-11
fix(tests): replace numberposts => -1 to clear cs-lint regression

faisalahammad:
## Description

The `cs-lint` workflow runs PHPCS on `tests/` and blocks on any failure. Commit `a5d8f506` (Jun 24) introduced `numberposts => -1` in `test_create_guest_author_returns_error_and_cleans_up_on_term_failure`, which trips the VIP rule `WordPressVIPMinimum.Performance.NoPaging.posts_per_page_numberposts`. Because the workflow's `paths:` filter only triggers on `**.php` changes, the failure stayed hidden until any later PR touched PHP — including unrelated PRs like #1346.

This PR replaces both `get_posts([ 'post_status' => 'any', 'numberposts' => -1 ])` count blocks with a private `count_guest_author_posts()` helper that sums every status from `wp_count_posts()`. Semantic behaviour for the test is unchanged; the rule no longer trips.

Refs #1346 (blocked on this).

## Deploy Notes

No new dependencies. Test-only change.

## Steps to Test

1. Check out this PR.
2. Run `./vendor/bin/phpcs --standard=.phpcs.xml.dist tests/Integration/GuestAuthors/CreateGuestAuthorTest.php` — expect 0 errors (was 2).
3. Run `./vendor/bin/phpcs --standard=.phpcs.xml.dist tests/` — expect all files clean.
4. Run `composer test:integration -- --filter CreateGuestAuthor` — expect `test_create_guest_author_returns_error_and_cleans_up_on_term_failure` to pass.
5. Confirm the `Basic CS and QA checks` check on this PR turns green in CI.
6. After merge, re-run CI on #1346 — its `Basic CS and QA checks` failure was this same regression; it should now be green.

### DIFF
--- a/tests/Integration/GuestAuthors/CreateGuestAuthorTest.php
+++ b/tests/Integration/GuestAuthors/CreateGuestAuthorTest.php
@@ -101,14 +101,10 @@ class CreateGuestAuthorTest extends TestCase {
 		$guest_author_obj = $coauthors_plus->guest_authors;
 
 		// Count posts before attempting to create the guest author.
-		$posts_before = get_posts(
-			array(
-				'post_type'   => $guest_author_obj->post_type,
-				'post_status' => 'any',
-				'numberposts' => -1,
-			)
-		);
-		$count_before = count( $posts_before );
+		// Sum every status key on the wp_count_posts() result so we count
+		// across all statuses without using numberposts => -1, which is
+		// prohibited by the VIP coding standard.
+		$count_before = $this->count_guest_author_posts( $guest_author_obj->post_type );
 
 		// Force wp_insert_term to fail by using a filter.
 		add_filter(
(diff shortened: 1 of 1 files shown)


GaryJones:
Closing in favour of #1364, which has now been merged into `develop`.

Both PRs target the same failing `Basic CS and QA checks` job, but they diagnose and fix it differently, and after review I went with #1364's approach:

**Diagnosis.** The test file last changed on 24 June and the check was green on 26 June, so the code didn't regress. What changed is the tooling: CI installs `automattic/vipwpcs: ^3` without a committed `composer.lock`, so a VIPCS release since then promoted `WordPressVIPMinimum.Performance.NoPaging` to an error on those two lines. It's the PHP-side twin of the Prettier drift absorbed in #1361, and it broke every open PR at once rather than being introduced by any one commit.

**Why the `phpcs:ignore` over the `wp_count_posts()` rewrite.** The unbounded `get_posts()` count is deliberate and its assertion is exact. `wp_count_posts()` isn't a like-for-like substitute: it sums *all* statuses (including `trash`/`auto-draft`, which `post_status => 'any'` excludes) and it's cached. It happens to stay correct here only because cleanup force-deletes and the intervening `wp_insert_post` invalidates the counts cache — two coincidences that a future change to the cleanup path could quietly break, weakening the test without any visible failure. `NoPaging` is a production-runtime performance guard that doesn't apply to a one-shot count against an isolated test database, so suppressing it with a justification comment (the pattern the suite already uses) is the more honest and robust fix.

Thanks very much for digging into this, @faisalahammad — the analysis was genuinely useful in pinning down what was going on.

PR #1349, opened 2026-08-11
fix: raise author search limit from 10 to 100 (#1050)

### LINKED ISSUE #1050: Gutenberg widget fails with more than 50 guest authors
When the site has more than 50 guest authors, the Gutenberg widget may malfunction and display an infinite spinner after loading. This issue seems to be caused by the API returning only the first 50 records in alphabetical order. If an author outside of this list is selected, the plugin fails to retrieve the necessary value.

**Steps to Reproduce:**
Add more than 50 guest authors to the site.
Open the Gutenberg editor and try to use the widget.
Select a guest author that is alphabetically beyond the first 50 in the list.

**Expected Result:**
The widget should function correctly and display the selected guest author.

**Actual Result:**
The widget shows an infinite spinner and does not load properly.

**Additional Information:**
It appears that the API call is limited to 50 records. The plugin needs to handle pagination or request more records to include all guest authors.
https://github.com/Automattic/Co-Authors-Plus/blob/1c29766eeda11276bcf6f20e07171b3efcbc7c03/src/components/co-authors/index.jsx#L73

image

faisalahammad:
## Description

The co-author search endpoint and search_authors() used a hard-coded limit of 10 results. Sites with more than ~50 guest authors could not see or select authors past the first page in the Gutenberg sidebar combobox, causing an infinite spinner or failed adds.

Raised the default cap to 100 and wired a per_page parameter through the REST handler. Existing callers and the filter still work.

Fixes #1050

## Deploy Notes

None.

## Steps to Test

1. Install the build zip on a site with 15+ guest authors.
2. Open the block editor and the Co-Authors panel.
3. Confirm the "Select An Author" dropdown shows more results than the old limit of 10.
4. Select a guest author that would have been excluded before.
5. Confirm the selection succeeds and persists after save.

### DIFF
--- a/php/class-coauthors-endpoint.php
+++ b/php/class-coauthors-endpoint.php
@@ -80,6 +80,19 @@ class Endpoints {
 							'type'        => 'string',
 							'required'    => false,
 						),
+						'per_page'         => array(
+							'description'       => __( 'Maximum number of authors to return.', 'co-authors-plus' ),
+							'required'          => false,
+							'type'              => 'integer',
+							'default'           => 100,
+							'sanitize_callback' => function( $value ): int {
+								return absint( $value );
+							},
+							'validate_callback' => function( $value ): bool {
+								$value = (int) $value;
+								return $value >= 1 && $value <= 200;
+							},
+						),
 					),
 				),
 			)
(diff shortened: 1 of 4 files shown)


GaryJones:
Thanks for this, @faisalahammad, and for the work digging into #1050.

We're closing this in favour of #1356, which tackles the same issue from a different angle. Raising the limit from 10 to 100 lifts the ceiling, but it turned out the ceiling was never the real problem: in the block editor the search request was being cancelled by a re-render before it ever reached the server, so only the initial preloaded batch was ever shown (filtered client-side by `ComboboxControl`). #1356 fixes that root cause, so search works at any scale without preloading a larger batch on every editor load.

Really appreciate the effort here — it helped pin down what was actually going on. Closing as superseded by #1356 (now merged).

PR #1350, opened 2026-08-12
fix: fire comment_notification_recipients filter in wp_notify_postauthor

### LINKED ISSUE #1026: comment_notification_recipients hook does not work with co authors plus plugin
I use this hook in a theme: comment_notification_recipients to add some extra emails, but it does not work, if your plugin is active. 

Could I use another hook? Any ideas?

faisalahammad:
## Description

Co-Authors Plus overrides the pluggable `wp_notify_postauthor()` with a fork that predates WordPress core's `comment_notification_recipients` filter. As a result, themes and plugins hooking `comment_notification_recipients` to add comment-notification email addresses see no effect while the plugin is active.

This rewrites the override so it still notifies every co-author assigned to the post (linked accounts and guest authors), but first runs the recipient list through the core `comment_notification_recipients` filter. From there it keeps to the core pipeline: the `comment_notification_notify_author` skip, per-recipient `switch_to_user_locale()`, and `comment_notification_headers` / `comment_notification_text` / `comment_notification_subject` for the message.

The plugin's reason for overriding the function (notifying all co-authors, not just the post author) is preserved: when no co-author terms are assigned the post author is used as a fallback, matching the current behavior.

Fixes #1026

## Deploy Notes

No new dependencies or deployment steps. Standard WordPress.org plugin release process applies.

## Steps to Test

1. Check out this branch.
1. Install Co-Authors Plus on a site, create a post with two co-authors assigned.
1. Add the following to a theme's `functions.php`:

   ```php
   add_filter( 'comment_notification_recipients', function ( $emails ) {
       $emails[] = 'extra@example.com';
       return $emails;
   } );
   ```

1. Leave a comment on the post.
1. Confirm all co-authors receive the notification email, and that `extra@example.com` also receives one.

Automated integration tests in `tests/Integration/Comments/CommentNotificationsTest.php` cover the multi-author case, guest authors, the own-comment skip, filter-added addresses, and the `comment_notification_notify_author` re-include path.

### DIFF
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -133,6 +133,8 @@ if ( ! function_exists( 'wp_notify_postauthor' ) ) :
 		$notify_author = apply_filters( 'comment_notification_notify_author', false, $comment->comment_ID );
 
 		// The comment was left by one of the co-authors.
+		// Note: for a guest author, $coauthor->ID is the guest-author CPT's post ID, not a WP user ID —
+		// it could theoretically collide with $comment->user_id or the current user's ID for an unrelated post/user pair.
 		foreach ( $coauthors as $coauthor ) {
 			if ( ! $notify_author && (int) $comment->user_id === (int) $coauthor->ID ) {
 				unset( $emails[ $coauthor->user_email ] );
@@ -230,7 +232,7 @@ if ( ! function_exists( 'wp_notify_postauthor' ) ) :
 					$subject = sprintf( __( '[%1$s] Pingback: "%2$s"', 'co-authors-plus' ), $blogname, $post->post_title );
 					break;
 
-				default: 
+				default:
 					// Comments.
 					/* translators: %s: Post title. */
 					$notify_message = sprintf( __( 'New comment on your post "%s"', 'co-authors-plus' ), $post->post_title ) . "\r\n";
@@ -253,7 +255,7 @@ if ( ! function_exists( 'wp_notify_postauthor' ) ) :
 			/* translators: %s: Comment URL. */
 			$notify_message .= sprintf( __( 'Permalink: %s', 'co-authors-plus' ), get_comment_link( $comment ) ) . "\r\n";
 
-			if ( user_can( $post->post_author, 'edit_comment', $comment->comment_ID ) ) {
+			if ( $user && user_can( $user, 'edit_comment', $comment->comment_ID ) ) {
 				if ( EMPTY_TRASH_DAYS ) {
 					/* translators: Comment moderation. %s: Comment action URL. */
 					$notify_message .= sprintf( __( 'Trash it: %s', 'co-authors-plus' ), admin_url( "comment.php?action=trash&c={$comment->comment_ID}#wpbody-content" ) ) . "\r\n";
--- a/co-authors-plus.php
+++ b/co-authors-plus.php
@@ -63,125 +63,230 @@ CoAuthors\Blocks::run();
 if ( ! function_exists( 'wp_notify_postauthor' ) ) :
 	/**
 	 * Notify a co-author of a comment/trackback/pingback to one of their posts.
-	 * This is a modified version of the core function in wp-includes/pluggable.php that
-	 * supports notifs to multiple co-authors. Unfortunately, this is the best way to do it :(
+	 *
+	 * This mirrors the core function in wp-includes/pluggable.php but pre-populates
+	 * the recipient list with all co-authors of the post, so every co-author gets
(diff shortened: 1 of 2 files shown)


GaryJones (changes requested):
## Review

Thanks for this — the core change is sound, and the failing CI here is not caused by your code.

### The red CI is stale and environmental

This CI run is from 12 August, and the PR is now in a merge-conflict state, so it has never run against the current `develop`.

- **Behat** — all 20 scenarios fail identically with `✖ Environment not initialized. Run \`wp-env start\` first.`, ending in the generic `Failed to activate plugin: (RuntimeException)`. The wp-env container never started; the plugin code is never exercised. Behat is green on `develop`.
- **Playwright** — dies during setup with `GitError: fatal: couldn't find remote ref 7.0.4` (it tried to check out a non-existent WordPress tag). Pure infrastructure.

Unit tests, Basic CS, and the full WP 6.4 / WP-master integration matrices all passed — which they wouldn't if the rewrite broke plugin load. A rebase onto `develop` will retrigger fresh CI, which is the real next step.

### The merge conflict / redundant commit

The second commit ("satisfy VIP NoPaging in CreateGuestAuthorTest") rewrites the `get_posts()` calls to `posts_per_page => 100, fields => ids`. `develop` already fixed the *same lines* independently in the merged #1364, keeping `numberposts => -1` behind a `phpcs:ignore`. That's almost certainly the conflict. Please rebase and **drop that second commit**, taking `develop`'s resolution — it's now redundant and unrelated to this fix.

### The core change

Faithful port of modern core `wp_notify_postauthor()`, pre-seeded with all co-authors and run through `comment_notification_recipients`. Good calls: dedup via `array_flip`, per-recipient `switch_to_user_locale()`, `WP_Http::is_ip_address()` over `@gethostbyaddr`, and `get_comment_link()`. The post-author fallback preserves the original override reason, and the integration coverage is a real improvement.

### Points worth addressing before merge

1. **Moderation links are gated by the post author's caps, not each recipient's.** `user_can( $post->post_author, 'edit_comment', … )` decides the Trash/Delete/Spam links for *every* co-author. Core evaluates this per recipient, so a co-author who can moderate may get no links (or vice versa). Easy to move inside the recipient loop.
2. **Weak assertion in `test_comment_author_is_not_notified_for_their_own_comment`.** The post author is the only co-author and also the commenter, so `$emails` empties out and the function returns early — `sent_to` is empty and `assertNotContains` passes trivially. Add a second co-author so the test proves the commenter is skipped *while others are still notified*.

### Optional / low priority

3. Guest-author `ID` (a post ID) is compared against `$comment->user_id` (a user ID) in the skip loop — a theoretical collision could unset the wrong recipient. Worth a comment at most.
4. Trailing whitespace after `default:` in the switch.

Bottom line: correct, well-tested fix; the red CI is stale infrastructure. Rebase on `develop`, drop the redundant NoPaging commit, and address 1–2 above.

faisalahammad:
Thanks for the review. Rebased on develop and pushed the fixes.

1. Rebase: dropped the second commit (VIP NoPaging change in `CreateGuestAuthorTest.php`). `develop` already fixed those same lines differently in #1364, so I took develop's version instead of mine. Now this branch is one commit on top of current develop.

2. Moderation links: changed the check to use each recipient's own capability, `$user && user_can( $user, 'edit_comment', $comment->comment_ID )`, instead of the post author's. One honest note: I checked WordPress core's current `wp_notify_postauthor()` and core still uses the post author's capability for this, not per recipient. But the fix is still right here, since this plugin notifies multiple real co-authors who can have different capabilities. Core only ever has one real recipient (the post author), so it never needed a per-recipient check.

3. Test: added a second co-author to `test_comment_author_is_not_notified_for_their_own_comment`. It now checks the commenter is skipped AND the other co-author still gets notified, instead of passing trivially on an empty send list.

4. Added a comment above the skip loop about the guest-author ID vs user ID collision risk you flagged. No logic change, as you said.

5. Removed the trailing space after `default:`.

Also found and fixed one more weak assertion while at it (CodeRabbit caught it, same class of issue as point 3): `test_comment_notification_recipients_includes_guest_author` now also checks the guest author's email reaches `sent_to`, not just the pre-filter recipients list.

Full integration suite passes (280 tests, 960 assertions), no regressions. Re-requesting your review.

GaryJones (approved):
All review points addressed: per-recipient moderation-link gating, strengthened self-comment and guest-author tests, collision note, whitespace. Rebased cleanly and all CI checks green. Thanks Faisal!

Issue #129, opened 2013-07-25
Add support for Atom feeds


nickdaugherty:
Currently, Co Authors Plus includes a hook for adding coauthors to RSS feeds as `<dc:creator />` elements.

Let's add support for ATOM entries in the same way, using `<atom:author>` or `<atom:contributor>`.

AJB99:
What is the hook for adding coauthors to RSS feeds?

I've tried the following, but it doesn't work:
https://wordpress.org/support/topic/coauthor-in-rss-feed-fix/

PR #1352, opened 2026-08-13
fix: apply co-author bylines to feeds without requiring coauthors_auto_apply_template_tags

### LINKED ISSUE #736: How to display co-authors within RSS feed <dc:creator> tags?
Currently all the <dc:creator> tags in my RSS feed are the WP User that originally created the post, and not the co-author assigned as the author for display.

How do I get the co-author to display within the <dc:creator> tags of my RSS feed?

I've already attempted the following, but it doesn't work:
```
function db_coauthors_in_rss( $the_author ) {

    if ( !is_feed() || !function_exists( 'coauthors') ){
        return $the_author;
    }
    else{
        return coauthors( null, null, null, null, false );
    }

}
add_filter( 'the_author', 'db_coauthors_in_rss');
```

chris920:
## Description

Fixes #736.

Co-Authors Plus ships working RSS feed support that never runs on a default install. This PR
registers the feed filters independently of the theme template-tag opt-in, so `<dc:creator>`
reflects the assigned co-author out of the box.

### The problem

`CoAuthors_Template_Filters::register_hooks()` bundles four hooks:

```php
add_filter( 'the_author', array( $this, 'filter_the_author' ) );
add_filter( 'the_author_posts_link', array( $this, 'filter_the_author_posts_link' ) );

// Add support for Guest Authors in RSS feeds.
add_filter( 'the_author', array( $this, 'filter_the_author_rss' ), 15 );
add_action( 'rss2_item', array( $this, 'action_add_rss_guest_authors' ) );
```

`CoAuthors_Plus::action_init()` only calls that method when `coauthors_auto_apply_template_tags`
returns true, and it defaults to false. So on a default install `wp-includes/feed-rss2.php` calls
`the_author()`, no CAP filter is attached, and the value resolves from `post_author`.

Result: every feed item is credited to whichever WordPress user hit publish, not to the assigned
co-author. For a guest author with no `wp_users` row, there is nothing correct to fall back to.

The front end stays correct throughout, because the documented integration path is for themes to
call `coauthors()` directly. Nothing surfaces the discrepancy, so sites can run this way for years
without noticing.

### Why this is a bug rather than a configuration choice

The two hook groups are different concerns behind one switch.

`the_author` and `the_author_posts_link` in theme templates are legitimately opt-in. CAP cannot know
whether a theme already calls `coauthors()`, and filtering unconditionally risks unexpected output.
Default false is correct there.

Feed output is not a theming concern. `feed-rss2.php` is core-generated and no theme controls it.
There is no coherent configuration that wants CAP bylines on the page and `post_author` in the feed.

CAP already treats one feed surface as non-optional. `CoAuthors_Plus::fix_author_page()` pins the
global `$authordata` to the guest author, guarded only by `is_author()`, with no template-tag opt-in.
So on a default install:

| Feed | Byline on a default install |
|---|---|
| `/author/{guest-author}/feed/` | correct co-author |
| `/feed/` | publishing WordPress user |
| `/category/{slug}/feed/` | publishing WordPress user |

Author feeds work because of a global that happens to persist through the loop. Everything else does
not. That inconsistency reads as an oversight rather than a design decision.

This is directly observable in the test suite added here: with the new registration disabled, the
author-archive-feed test still passes while the site-feed tests fail.

### Real-world impact

Found on a trade publishing site running four titles on one install. Of the last 300 published posts,
105 had a feed byline that did not match the assigned co-author.

The severity is not evenly distributed. Editorial staff publish on behalf of freelancers and
contributors, so the wrong name is almost always a staff name. On that site every sponsored post was
syndicating under the Creative Director's byline, and contributed pieces from three outside
organizations were going out under the managing editor's name. That is an attribution problem with
more weight than a cosmetic one, and it was invisible because the article pages were correct.

### Changes

**1. New `CoAuthors_Feed_Filters` class** (`php/class-coauthors-feed-filters.php`)

`filter_the_author_rss()` and `action_add_rss_guest_authors()` move here, following the existing
composition-root pattern of a side-effect-free constructor with a separate `register_hooks()`. They
are removed from `CoAuthors_Template_Filters` so sites already opted into
`coauthors_auto_apply_template_tags` do not double-register and emit duplicate `dc:creator` elements.

The `require_once` sits alongside its sibling `class-coauthors-template-filters.php` at the top of the
plugin file rather than lazily inside `action_init()` the way `class-coauthors-guest-authors.php` does.
Guest authors are a large optional subsystem behind a stored setting; this is one small class split out
of a file that is already eagerly required, and eager loading keeps it available regardless of the
opt-out filter. Happy to move it if you would rather it be lazy.

**2. Registered independently, with an opt-out**

```php
if ( apply_filters( 'coauthors_filter_feed_authors', true ) ) {
    global $coauthors_plus_feed_filters;
    $coauthors_plus_feed_filters = new CoAuthors_Feed_Filters();
    $coauthors_plus_feed_filters->register_hooks();
}
```

Defaults on. The instance is held in a global mirroring the existing
`$coauthors_plus_template_filters`, so individual hooks can be unhooked by site code and tests.

Any site depending on current behavior restores it with:

```php
add_filter( 'coauthors_filter_feed_authors', '__return_false' );
```

**3. Fixed double-encoding inside CDATA**

`action_add_rss_guest_authors()` ran `esc_html()` on a value placed inside a CDATA section. Everything
between `<![CDATA[` and `]]>` is literal, so a display name like `Smith & Jones` was emitted as
`Smith &amp; Jones` and rendered with the entity text visible. Core's `feed-rss2.php` does not escape
inside CDATA. Now only the CDATA terminator is guarded. Indentation also switched from six spaces to
tabs to match the surrounding core template output.

**4. Restored core's entity normalization**

Core registers `add_filter( 'the_author', 'ent2ncr', 8 )`. `filter_the_author_rss` runs at 15 and
replaces the string wholesale, so the replacement never passed through it. Invisible in RSS2 where the
value sits in CDATA, but `feed-atom.php` renders `the_author()` inside a bare `<name>` element. The
return value now passes through `ent2ncr()`.

Alternative worth a maintainer opinion: registering at priority 5 and letting core's filter run after
would achieve the same thing more idiomatically, at the cost of changing hook ordering and possibly
surprising anyone hooked between 5 and 15. Happy to switch if preferred.

Scope note: this restores parity with core's handling. It does not fix raw `&` or `<` in the Atom
`<name>` element, which is a pre-existing core issue affecting non-CAP sites identically.

**5. Minor robustness**

Consistent array casting between the two methods, and co-authors with an empty `display_name` are
skipped rather than emitting an empty element.

### Backward compatibility

This changes published output. Sites that had not opted in will see feed bylines change on upgrade,
retroactively across every item currently in their feed window. That is the fix working as intended,
but it is visible, so it belongs in release notes rather than buried. The changelog entry states it
plainly and gives the opt-out.

No change to front-end output. The filter is gated on `is_feed()`, and template-tag filtering remains
behind `coauthors_auto_apply_template_tags` exactly as before.

### Related

- Closes the practical gap left by #92, which added this code in 2012 but only behind a manual opt-in
- Fixes #736, which was answered as a configuration question rather than as the bug underneath it
- Does **not** address #129 (Atom feeds). `rss2_item` does not fire on Atom, so additional co-authors
  still will not appear there. That is a feature rather than a bug fix, and deliberately left out of
  this PR. Happy to open a follow-up adding an `atom_entry` handler if this direction is acceptable.

### Open questions for maintainers

1. `coauthors_filter_feed_authors` as the filter name, or something closer to existing conventions?
2. `ent2ncr()` in the filter body, or move registration to priority 5?
3. Is a separate class the right shape, or would you rather keep one class with two registration
   methods?
4. Eager `require_once` at the top of the plugin file, or lazy inside `action_init()`?

## Deploy Notes

No new dependencies.

One behavior change to call out in the release notes: feed bylines change on upgrade for any site that
had not set `coauthors_auto_apply_template_tags` to true. The opt-out is a single filter, documented in
the changelog entry under a new `## [Unreleased]` heading.

## Steps to Test

Automated:

```sh
composer lint
composer test:integration
composer test:integration-ms
composer test:unit
composer cs -- tests/
```

Seven new integration tests in `tests/Integration/Feeds/FeedFiltersTest.php`:

- Guest author appears in `dc:creator` with no opt-in filter set (the regression test)
- Multiple co-authors emit one `dc:creator` each, in byline order
- `coauthors_filter_feed_authors` returning false restores core behavior
- The hooks are registered by default (the inverse of the above, exercised through `action_init()`)
- `the_author` outside a feed is unchanged when `coauthors_auto_apply_template_tags` is false
- A display name containing `&` appears once, unescaped, inside CDATA
- Author archive feeds still resolve correctly alongside `fix_author_page()`

Manual, in `wp-env`:

1. Confirm no `coauthors_auto_apply_template_tags` filter is set anywhere.
2. Assign a guest author to a post published by a different WordPress user.
3. Load `/?feed=rss2`. `dc:creator` shows the guest author, not the publishing user.
4. Load the post on the front end. The byline is unchanged.
5. Add `add_filter( 'coauthors_filter_feed_authors', '__return_false' );` in an mu-plugin and reload
   `/?feed=rss2`. The byline reverts to the publishing user.
6. Remove the mu-plugin and reload. The byline returns to the guest author.


GaryJones:
Thanks for this — it's a well-argued PR. The framing is the strongest part: separating theme-controlled template tags (legitimately opt-in) from core-generated feed markup (no theme controls it, so no coherent reason to gate it) is the right distinction, and the `fix_author_page()` precedent is a good supporting point. The class split follows the composition-root pattern, the CHANGELOG and AGENTS.md are updated properly, and the CDATA double-encoding and `ent2ncr` fixes are both correct and clearly justified.

## CI

Six of seven checks are green, including both integration matrices that run the seven new `FeedFiltersTest` cases, plus Behat, Playwright, and the unit jobs.

The one red check — **Basic CS and QA** — is unrelated to this PR. It fails on `tests/Integration/GuestAuthors/CreateGuestAuthorTest.php` (lines 108/140) for `WordPressVIPMinimum.Performance.NoPaging`, a file this PR doesn't touch. That drift was already fixed on `develop` in `3ddbebd`. This branch forked from `ce0e498` and is ~37 commits behind, so it predates the fix. **A rebase on `develop` clears it** with no code change, and gets you tested against current develop.

## Findings (all minor, none blocking)

1. **Empty first-author name isn't handled symmetrically.** `action_add_rss_guest_authors()` skips co-authors with an empty `display_name`, but `filter_the_author_rss()` doesn't: if `$coauthors[0]->display_name` is `''`, it returns `ent2ncr('')` and emits an empty `<dc:creator>` rather than falling back to `$the_author`. The "empty names are skipped" robustness note only holds for the additional authors. An edge case, but the two methods reading differently invites confusion — consider `! empty( $coauthors[0]->display_name )` in the guard.

2. **The `ent2ncr` asymmetry is defensible but worth a comment.** The first author passes through `ent2ncr()`; the extras don't. Your reasoning is sound (extras only ever appear in RSS2 CDATA, and `ent2ncr` exists for Atom's bare `<name>`, which extras never reach), so it's not a bug — but a reader diffing the two methods will notice. A one-line note on the extras loop would pre-empt the question.

3. **The gate test re-runs all of `action_init()`.** `reregister_feed_filters()` re-registers every hook that method wires, not just the feed filters. It's safe because `WP_UnitTestCase` backs up and restores `$wp_filter` per test, so there's no leakage — but it's a broad hammer for a narrow assertion. Extracting feed-filter registration into its own method the test could call directly would be cleaner. Optional.

## On your open questions

1. `coauthors_filter_feed_authors` is consistent enough with the existing `coauthors_*` naming.
2. Keep `ent2ncr` in the body. Moving to priority 5 would silently reorder anything hooked between 5 and 15, which isn't worth it.
3. A separate class is the right shape.
4. Eager `require_once` is fine given how small the class is and that its sibling is already eagerly required.

Approve-worthy after a rebase to clear the unrelated CS failure. The three findings above are polish rather than blockers.

chris920:
Thanks for the thorough read, Gary — and for answering all four open questions. That settles the design, so I've left the filter name, the `ent2ncr()` placement, the separate class, and the eager `require_once` exactly as they were.

Rebased onto `develop`. The branch is now a clean fast-forward, and the `WordPressVIPMinimum.Performance.NoPaging` drift in `CreateGuestAuthorTest.php` is gone with it, as you predicted — no code change needed.

## The three findings

**1. Empty first-author name** — fixed. The guard is now `! empty( $coauthors[0]->display_name )`, which also makes the redundant `count( $coauthors ) >= 1` unnecessary. You were right that the two methods reading differently was the real problem, so the docblock now states the shared rule: an empty name falls back rather than emitting an empty element, in both methods.

Added a test for it. It injects two nameless co-authors through the `get_coauthors` filter and asserts the feed still credits the publishing user rather than emitting `<dc:creator><![CDATA[]]></dc:creator>`. I checked it fails against the previous guard before keeping it:

```
- Expected: array( 'Publishing Editor' )
+ Actual:   array( '' )
```

**2. `ent2ncr` asymmetry** — documented. A four-line note on `action_add_rss_guest_authors()` explains that the normalization exists for the bare `<name>` element in `feed-atom.php`, and that `rss2_item` never fires on Atom, so the extras only ever reach RSS2 CDATA. Exactly the question you predicted a reader would have when diffing the two.

**3. The gate test's broad hammer** — fixed properly rather than worked around. Registration moved into `CoAuthors_Plus::maybe_register_feed_filters()`, and `reregister_feed_filters()` calls that instead of `action_init()`. The test now re-runs only the registration it asserts on.

## Test results

Run locally against the rebased branch:

| Suite | Result |
|---|---|
| `composer lint` | 139 files, no syntax errors |
| `composer cs -- tests/` | Pass |
| `composer test:unit` | 122 tests, 274 assertions |
| `composer test:integration` | 377 tests, 1172 assertions |
| `composer test:integration-ms` | 379 tests, 1183 assertions |
| `composer behat` | 185 scenarios, 2051 steps |
| `npm run test:unit` | 85 tests |
| `npm run lint:js` / `lint:css` / `build` | Pass |

`FeedFiltersTest` is now eight cases with the new one added.

Worth noting: the force-push reset the workflow approval, so all five checks are sitting at `action_required` until someone with write access approves the run.

## One unrelated thing I ran into

Getting Behat running locally surfaced a pre-existing bug on `develop`, in the `docker exec` fast path you have in `tests/Behat/FeatureContext.php`. `get_cli_container()` matches bind-mount sources with:

```php
docker inspect %s --format "{{range .Mounts}}{{.Source}}\n{{end}}"
```

In a single-quoted PHP string that `\n` stays a literal backslash-n, and Docker does not interpret escape sequences in `--format`. Verified:

```
$ docker inspect <container> --format '{{range .Mounts}}{{.Source}}\n{{end}}' | od -c
0000000  / h o s t _ m n t / U s e r s /
0000020  c h r i s l i u / p r o j e c t
0000040  s / c o - a u t h o r s - p l u
0000060  s  \  n  / U s e r s / c h r i s l
```

Every mount source comes back on one line, so `$source === $repo_root` never matches, `get_cli_container()` returns `''`, and every call takes the `parent::run_wp_cli_command()` fallback to `wp-env run` — the ~1.07s-per-invocation cost the override was written to avoid. The suite still passes, it just pays the full price. A `{{println .Source}}` or `{{.Source}}{{"\n"}}` would fix it.

Untouched here since it's outside this PR's scope. Happy to open a separate one if useful.

PR #1353, opened 2026-08-17
fix: guest author byline link points to guest author archive

### LINKED ISSUE #1351: `coauthors_auto_apply_template_tags` changes attribution name, but link points to old author
I have a site where we prefer using Guest Authors over regular users for attribution.  The site is using an old theme that does not support Co-Authors Plus template tags, and I would prefer not to have to create a child theme, so I reached for `coauthors_auto_apply_template_tags`.  

When setting `coauthors_auto_apply_template_tags` to true, on a post where the attribution is set to a single Guest Author, the author name updates properly on the frontend.  However, the byline link still points to the author page for the user who originally published the article.

faisalahammad:
## Description

Fixes #1351 for older themes that enable `coauthors_auto_apply_template_tags` but build the byline link themselves.

CAP filters `the_author()` so the guest author name is displayed, but an old theme can still call `get_author_posts_url( get_the_author_meta( 'ID' ) )` for the link. That ID comes from the post's original `post_author`, so the name and link can refer to different authors. The `author_link` filter now detects a singular post with one guest coauthor and uses that guest author's nicename when building the archive URL.

The earlier helper in `template-tags.php` was removed after confirming that `coauthors_posts_links()` already produced the correct canonical guest archive URL. Pretty-permalink links remain `/author/{guest-nicename}/`; plain-permalink links remain `?author_name={guest-nicename}`. The existing `esc_url()` hardening in `coauthors_wp_list_authors()` is retained.

The regression coverage includes the old-theme rendering pattern, both permalink styles for CAP's own template tag, and a publisher-absence assertion.

## Deploy Notes

No new dependencies.

## Steps to Test

1. Check out this PR and enable old-theme compatibility with `add_filter( 'coauthors_auto_apply_template_tags', '__return_true' );`.
2. Create a guest author named `jane-guest` and a separate WordPress user named `publisher-bob`.
3. Create a post published by `publisher-bob`, then set its only coauthor to `jane-guest`.
4. Use an old-theme byline pattern that calls `get_author_posts_url( get_the_author_meta( 'ID' ) )` for the link and `get_the_author()` for the label.
5. Confirm the label is `jane-guest` and the link goes to the guest archive, not the publisher archive.
6. Repeat with pretty and plain permalinks. Confirm CAP's `coauthors_posts_links()` keeps its canonical URL shape in both cases.

### DIFF
--- a/php/class-coauthors-guest-authors.php
+++ b/php/class-coauthors-guest-authors.php
@@ -1638,6 +1638,22 @@ class CoAuthors_Guest_Authors {
 	 * @since 3.0
 	 */
 	public function filter_author_link( $link, $author_id, $author_nicename ): ?string {
+		// Old themes can filter the author name but build the link from the
+		// post_author field. Use the single guest coauthor's slug in that case.
+		global $authordata;
+		$post = get_post();
+		if ( isset( $authordata->type ) && 'guest-author' === $authordata->type && (int) $author_id === (int) $authordata->ID ) {
+			$author_nicename = $authordata->user_nicename;
+		}
+		if ( $post && ! is_admin() && is_singular() ) {
+			$coauthors = get_coauthors( $post->ID );
+			if ( 1 === count( $coauthors ) && isset( $coauthors[0]->type ) && 'guest-author' === $coauthors[0]->type ) {
+				$guest_author = $coauthors[0];
+				if ( (int) $author_id === (int) $post->post_author || (int) $author_id === (int) $guest_author->ID ) {
+					$author_nicename = $guest_author->user_nicename;
+				}
+			}
+		}
 
 		// If we're using this at the top of the loop on author.php,
 		// our queried object should be set correctly
--- a/template-tags.php
+++ b/template-tags.php
@@ -267,6 +267,27 @@ function coauthors_posts_links( $between = null, $betweenLast = null, $before = 
 }
 
 /**
+ * Get the URL for a coauthor's post archive.
+ *
+ * Guest authors are stored as posts rather than WP users, so get_author_posts_url()
+ * falls back to a default link that is not their archive. Run the author_link filter
+ * instead, which the plugin uses to resolve guest author links correctly.
+ *
+ * @param object $author Coauthor object.
+ * @return string
+ */
+function cap_get_coauthor_posts_url( $author ) {
+	if ( isset( $author->type ) && 'guest-author' === $author->type ) {
+		$link = apply_filters( 'author_link', '', $author->ID, $author->user_nicename );
+		if ( ! empty( $link ) ) {
+			return $link;
+		}
+	}
+
+	return get_author_posts_url( $author->ID, $author->user_nicename );
+}
+
+/**
  * Outputs a single co-author linked to their post archive.
  *
  * @param object $author
@@ -284,7 +305,7 @@ function coauthors_posts_links_single( $author ) {
 	}
 	$args        = array(
 		'before_html' => '',
-		'href'        => get_author_posts_url( $author->ID, $author->user_nicename ),
+		'href'        => cap_get_coauthor_posts_url( $author ),
 		'rel'         => 'author',
 		/* translators: %s: author display name */
 		'title'       => sprintf( __( 'Posts by %s', 'co-authors-plus' ), apply_filters( 'the_author', $author->display_name ) ),
@@ -705,7 +726,7 @@ function coauthors_wp_list_authors( $args = array() ) {
 			}
 		} else {
 			/* translators: %s: author display name */
-			$link = '<a href="' . get_author_posts_url( $author->ID, $author->user_nicename ) . '" title="' . esc_attr( sprintf( __( 'Posts by %s', 'co-authors-plus' ), $name ) ) . '">' . esc_html( $name ) . '</a>';
+			$link = '<a href="' . esc_url( cap_get_coauthor_posts_url( $author ) ) . '" title="' . esc_attr( sprintf( __( 'Posts by %s', 'co-authors-plus' ), $name ) ) . '">' . esc_html( $name ) . '</a>';
 
 			if ( ( ! empty( $args['feed_image'] ) ) || ( ! empty( $args['feed'] ) ) ) {
 				$link .= ' ';
--- a/template-tags.php
+++ b/template-tags.php
@@ -273,10 +273,12 @@ function coauthors_posts_links( $between = null, $betweenLast = null, $before = 
  * falls back to a default link that is not their archive. Run the author_link filter
  * instead, which the plugin uses to resolve guest author links correctly.
  *
+ * Internal helper, not part of the public template tag API.
+ *
  * @param object $author Coauthor object.
  * @return string
  */
-function cap_get_coauthor_posts_url( $author ) {
+function coauthors__get_coauthor_posts_url( $author ) {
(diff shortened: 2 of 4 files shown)


GaryJones (changes requested):
Thanks for this — the core fix is well-targeted and it ships with tests. A few things before merge, one of them blocking.

## 1. Blocking — the second commit is unrelated and causes the merge conflict

Commit `5fa758e` rewrites `CreateGuestAuthorTest.php` to count posts via `wp_count_posts()` instead of `get_posts( … numberposts => -1 )`. That has nothing to do with the byline link bug.

More to the point, `develop` has **already** resolved the same VIPCS `NoPaging` drift, but differently — `3ddbebd` keeps `numberposts => -1` with a `phpcs:ignore`. Both touch the same lines, which is exactly why the PR shows as conflicting.

Please **drop that commit and rebase onto `develop`**. The CI concern it addressed no longer exists, so removing it clears the conflict and leaves a single, focused commit.

## 2. Naming — the `cap_` prefix is new to this file

`cap_get_coauthor_posts_url()` uses a prefix that appears nowhere else in the plugin; every public function in `template-tags.php` uses `coauthors_`, `get_`, `is_` or `the_`. Since this is a new public global function it becomes part of the plugin's API on release. Could we either rename it to fit convention (e.g. `coauthors_get_posts_url()`) or reconsider whether it needs to be public at all?

## 3. Please confirm the root cause on both permalink styles

`get_author_posts_url()` applies the `author_link` filter itself, and `filter_author_link()` rebuilds guest URLs in that path too — so on the face of it the original call ought to have been corrected by the same filter. I'd like to understand precisely why it wasn't (empty `user_nicename` on the coauthor object? filter not firing in that context?), because that tells us whether we're fixing the cause or a symptom. Could you confirm reproduction on both pretty and plain permalinks? Not a blocker — the regression test does show the behaviour change — but worth nailing down.

For what it's worth, passing an empty `''` into the filter takes the `?author_name=…` branch, which is consistent with how the plugin already links guest authors elsewhere (e.g. the admin profile links around `class-coauthors-guest-authors.php:170`), so the query-arg form is expected, not a regression.

## 4. Nit — the integration test is slightly self-referential

`test_coauthors_posts_links_for_single_guest_author()` computes `$expected` by running the same `author_link` filter the production code runs, so expected and actual share a source of truth. It's a valid regression test (without the fix the href would be the publisher's URL and differ), but it can't distinguish "correct guest archive URL" from "whatever the filter emits". A stronger assertion would pin the concrete guest nicename in the URL **and** assert the original author's nicename is absent — that encodes the actual bug in #1351.

## Things done well

- Adding `esc_url()` in `coauthors_wp_list_authors()` is a genuine improvement — the original output was unescaped.
- The `isset( $author->type )` guard is safe for real `WP_User` objects, so they correctly fall through to `get_author_posts_url()` and existing behaviour is preserved.
- Both a unit and an integration test included — appreciated.

Summary: happy to approve in principle once (1) is sorted; (2)–(4) are smaller follow-ups.

GaryJones (changes requested):
Thanks for the thorough follow-up, Faisal — points 1, 2 and 4 from my previous review have been properly addressed: the unrelated commit is gone, the helper now follows the `coauthors__` internal convention, and the integration test builds its expected URL independently and pins the nicenames. I've also verified your numbers locally: 48 unit tests and 321 integration tests pass on this branch, and PHPCS reports no new violations on the changed files.

However, I dug into point 3 empirically, and I'm afraid the results change the picture: **the evidence says this code path was never broken, so this PR doesn't fix #1351 — it changes the shape of an already-correct URL.**

## The experiment

I reverted `template-tags.php` to develop's version and ran your two new integration tests against the *old* code:

- **Plain permalinks: the test passes.** The old code already emitted `?author_name=jane-guest` — the plugin's `author_link` filter corrects the core-built URL before output. No bug.
- **Pretty permalinks: the old code emitted `http://site/author/jane-guest/`** — the guest author's *own* archive URL, built from the nicename we pass to `get_author_posts_url()`. The test fails only because that differs in shape from the expected `?author_name=jane-guest`; both URLs point at the guest author, not the publisher.

```
--- Expected
+++ Actual
-'<a href="http://localhost:9911?author_name=jane-guest" ...>jane-guest</a>'
+'<a href="http://localhost:9911/author/jane-guest/" ...>jane-guest</a>'
```

I then seeded a live wp-env site (pretty permalinks, unlinked guest author `jane-guest`, post published by `publisher-bob`) and requested both URLs. `/author/jane-guest/` and `/?author_name=jane-guest` **both return 200 and render the same guest archive** listing the post.

So on the root cause: core's `get_author_posts_url()` builds the pretty URL from the *nicename* we pass in (it only falls back to a user lookup when the nicename argument is empty), and then applies `author_link`, where `CoAuthors_Guest_Authors::filter_author_link()` rebuilds it — again from the nicename. Under plain permalinks the CPT-ID form does appear briefly, but the same filter corrects it to `?author_name=` before anything is output. At no point does the publisher's URL enter this path.

That means the practical effect of this change is a small regression: every guest byline on a pretty-permalink site (the common case) moves from the canonical `/author/jane-guest/` to the non-canonical `http://site?author_name=jane-guest` (note also the missing `/` before the `?`), in `coauthors_wp_list_authors()` too — and the new tests would enshrine that as expected behaviour.

## Where #1351 probably actually lives

Since `coauthors_posts_links()` emits correct links, the reporter's symptom — name updates, link still points at the original publisher — most plausibly comes from their theme building the link outside anything Co-Authors Plus filters. For example, `get_author_posts_url( get_the_author_meta( 'ID' ) )` called directly uses the publisher's ID from `$authordata` and never passes through the `the_author_posts_link` filter that `coauthors_auto_apply_template_tags` hooks. If that's the path, no change to `coauthors_posts_links_single()` can fix it.

## Requested changes

1. Let's first reproduce the reported symptom. Could you (or we can ask on #1351) find out how the reporter's theme renders the byline — which template tag or function? A failing test that shows the byline href actually pointing at the publisher is the bar here; the current tests demonstrate a URL-shape change, not the bug.
2. Until we have that reproduction, please hold off on changing the URL generation in `coauthors_posts_links_single()` / `coauthors_wp_list_authors()` — as it stands it trades a canonical URL for a non-canonical one on working sites.

I appreciate this is frustrating after a careful revision — the test infrastructure you've built here (the permalink-style matrix, the independent expected URL, the publisher-absence tripwire) is genuinely good and will be exactly what we need once the real path is identified. The experiment above was only possible because your tests are now rigorous enough to run against the old code.

faisalahammad (commented):
## CI Fix Summary — 2 errors resolved (11 PHPCS violations)

The "Basic CS and QA checks" failure was PHPCS only, in the new tests.

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | tests/Integration/TemplateTags/CoauthorsPostsLinksTest.php:76-152 | Generic.WhiteSpace.ScopeIndent.IncorrectExact (10x) | Re-indented the over-indented test methods and their bodies with phpcbf, matching the one-tab member style of the other methods in the file |
| 2 | tests/Integration/TemplateTags/CoauthorsPostsLinksTest.php:114 | Generic.Commenting.DocComment.LongNotCapital | Reworded the docblock paragraph to start with "The coauthors_auto_apply_template_tags filter is simulated..." |

No production code changed in this commit. `composer cs -- tests/` passes locally on PHP 7.4, the same step that failed in CI.

Tests: all other checks were already green (unit 7.4 + latest, integration, Behat, Playwright).

Issue #1252, opened 2026-04-28
Current user not unassigned in editor


MastaBaba:
In at least two Wordpress installations, with the introduction of CoAuthors+ 4, I have the following issue. I'm now on 4.0.1.

I add a post. 
In the authors box in the sidebar, I never show up as the author. There's a spinner that never goes away.
I can add another author which shows up in the list of authors when I select them.
The spinner then goes away, and I'm not visible as the author.
I publish the post, and all looks fine. 
In the list of posts in the backend, my selected author shows up as the only author.

Except that, when requesting the user for the post in the frontend through get_coauthors($post_id), I'm the first author that's returned.

When I 'quick edit' the post, and switch the post to some other author, then to my desired author, the 'hidden' author is removed, and get_coauthors() returns the correct author.

When I don't publish the post, but save the post, the author I added in the author box disappears, and the spinner returns.

MastaBaba:
In 4.02 this issue persists for me, with the difference that, when creating a new post, I see no spinner in the author-box. In the author-box, nothing happens, and no author is showing. Adding an author works, but saving as draft results in that author being removed. Publishing the post, while an author is visible, works, but then the first author retrieved through coauthors plus is me, not the author that was explicitly set.

PR #1431, opened 2026-09-06
fix: signal unresolved co-author names in add_coauthors()

### LINKED ISSUE #1369: add_coauthors() silently drops a co-author it cannot resolve
## Problem

`add_coauthors()` accepts a name it cannot resolve to any user or guest author, stores it as a stray taxonomy term, returns `true`, and then silently drops it at read time. The caller has no way to know anything went wrong.

Demonstrated with an integration probe, passing a middle entry that matches nobody:

```
INPUT:  author1, jhon-doe, author20
STORED: cap-author1, jhon-doe, cap-author20
READ:   author1, author20
```

The unresolvable name is persisted in the correct position as an unprefixed `author` term, but `get_coauthors()` cannot map it back to a co-author, so it is dropped. No warning, no error, and `add_coauthors()` still reports success.

## Why it matters

In `add_coauthors()`, `get_coauthor_by()` returning false leads to `update_author_term( false )` returning false, so `$author_name` is left as the caller's raw string and `wp_set_post_terms()` creates a fresh term from it. This is most damaging during bulk migrations, where a caller has no realistic way to spot that a fraction of its bylines have quietly lost an author — it just looks like the data came out wrong. It also leaves orphaned terms behind that count towards nothing and match no co-author.

See #1052, where this is the most likely explanation for the reported behaviour.

## Possible directions

Worth discussing before implementing, since the right answer depends on how much we can change a long-standing public API:

- Skip unresolvable names rather than creating a term from them, so at least no orphan term is left behind.
- Surface the failure to the caller, either via a `WP_Error`, a returned list of names that could not be resolved, or a `_doing_it_wrong()`.
- Leave runtime behaviour alone but add a WP-CLI command to find and clean up orphaned author terms.

The return type is the awkward part: `add_coauthors()` currently has a `bool` return type, so returning a `WP_Error` would be a breaking change. An action fired for each unresolved name may be the least disruptive route.

faisalahammad:
## Description

`add_coauthors()` resolves each name it receives to a WordPress user or guest author. When a name could not be resolved, the plugin stored it as an author term without the `cap-` prefix, a term no read path can match, and `get_coauthors()` dropped it when the byline was read back. Bulk imports and migrations lost bylines with no signal. Fixes #1369.

This PR:

- **Retries resolution before giving up.** The classic metabox and some WP-CLI commands pass `user_login` values, while the default lookup field is `user_nicename`. If the first lookup fails, `add_coauthors()` now retries by login and then by the sanitized title of the name, so input shapes that were previously misread as unresolved now resolve.
- **Excludes names that still cannot be resolved from the term write.** The stray, unprefixed term was dead weight: archive queries filter on the `cap-` prefix, and the assignment service round-trips the slug, so the stray term broke its idempotence check. A post whose submitted names are all unresolved keeps its existing byline terms instead of being left termless.
- **Fires a `coauthors_unresolved_coauthor` action** once per unresolved name, with the name, lookup field, post ID, and append flag. This is the only signal. An earlier iteration also raised a `_doing_it_wrong()` notice, but that fired on legitimate runtime paths, such as a REST save feeding back an orphan term after a guest author was deleted outside the plugin, and under `WP_DEBUG_DISPLAY` it corrupted the REST response.

The block editor's main write path is not covered by this signal: core drops invalid term IDs before `add_coauthors()` runs. That gap needs its own issue and is not fixed here.

## Deploy Notes

No new dependencies. One behavior change to be aware of: names that cannot be resolved are no longer written as author terms. Nothing read them back, so no byline changes for names that resolved before.

## Steps to Test

1. Check out this PR and hook the action:

```php
add_action( 'coauthors_unresolved_coauthor', function ( $name, $field, $post_id, $append ) {
    error_log( "Unresolved co-author '{$name}' ({$field}) on post {$post_id}" );
}, 10, 4 );
```

2. Assign a mix of a real author and a name that matches nobody:

```php
global $coauthors_plus;
$coauthors_plus->add_coauthors( $post_id, array( 'admin', 'no-such-author' ) );
```

3. Confirm the debug log has one line for `no-such-author` only, the post stores just the real author's term, and the byline keeps its order.
4. Assign a user whose login and nicename differ (for example login `Jane Doe`, nicename `jane-doe-the-second`) the way the metabox does. Confirm no action fires; the name resolves through the login retry.
5. Delete a guest author post directly (leaving its orphan author term on a post), then save that post in the block editor. Confirm the save succeeds, the action fires once for the orphan slug, and the valid co-authors stay on the byline.

Automated: `composer test:integration`. `AddCoauthorsTest` covers the action args, exclusion from the term write, the login-shaped input, append mode, and the orphan-term REST save. `CoauthorsLinksSingleTest` passes `user_login` inputs again, which the retry now resolves.


GaryJones (changes requested):
Thanks for taking this on — surfacing silent byline loss is absolutely worth doing, and the action + tests are a good shape for it. I've reviewed the branch in some depth, though, and I don't think the PR is mergeable in its current form: the central premise ("a name that fails `get_coauthor_by()` will not appear in the byline") turns out not to hold, which makes the new notice fire falsely — including on the plugin's own internal call paths. Details below, with a suggested path through.

## The core problem: the notice fires for names that *do* resolve

The author term's **name** is the co-author's `user_login` (`wp_insert_term( $coauthor->user_login, … )` at `php/class-coauthors-plus.php:2465`), and `wp_set_post_terms()` falls back to a `term_exists()` **name** match. So passing a guest author's `user_login` (e.g. `Jane Doe`) fails the `user_nicename` lookup, fires the new action and `_doing_it_wrong()` — and then attaches the correct `cap-jane-doe` term by name and renders the byline perfectly.

That's why `CoauthorsLinksSingleTest` passed on `develop` and broke under this PR. The description frames that as the notice catching a bug in the test, but I'd argue the opposite: the test was covering a live input shape (the classic metabox submits `name="coauthors[]"` with `value="{$coauthor->user_login}"` at `php/class-coauthors-plus.php:591`, and both CLI commands pass `user_login` too), and switching it to nicenames removes the only end-to-end coverage of that path while hiding the false positive.

## Append mode reports the plugin's own data as caller error

In append mode, existing co-authors are plucked by `user_login` (line 1669) but re-resolved via `$query_type` (default `user_nicename`) at line 1704. Any existing co-author whose login differs from their nicename — every guest author with a spaced login, any WP user with a truncated or `-2`-suffixed nicename — is reported as an unresolved *caller* error for a name the caller never passed. `delete_user_action()` (line 1782, `append = true`) hits this on every reassigned post, as does `wp co-authors-plus assign-user-to-coauthor` (`php/cli/class-assign-user-to-coauthor-command.php:131` passes `user_login`) and `swap-coauthors` (`php/cli/class-swap-coauthors-command.php:262` plucks `user_login`). On a bulk migration that's one `_doing_it_wrong` per post while every assignment actually succeeds via the term-name match.

The new append-mode test can't catch this because the factory users all have `user_login === user_nicename` — a fixture where they differ would fail today.

## `_doing_it_wrong()` is the wrong tool for the runtime paths

`_doing_it_wrong()` implies developer misuse, but this fires on data conditions nobody miscoded:

- `sync_coauthors_on_rest_save()` (line 1589) feeds every existing author term slug back into `add_coauthors()`, so a post carrying an orphan term (guest author deleted outside `CoAuthors_Guest_Authors::delete()`) emits the notice on **every block-editor save, forever** — the non-append write re-persists the same slug. With `WP_DEBUG_DISPLAY` on, the notice lands inside the REST response body and the editor shows "The response is not a valid JSON response" on an otherwise successful save.
- The classic-editor save path (line 1627) and the `coauthors/v1` endpoint (`php/class-coauthors-endpoint.php:207`, client-supplied names) reach it too, where it can fire before `wp_redirect()`.

## The back-compat rationale for keeping the stray term only half-holds

Keeping the stray term so a later-created guest author can heal the byline works for the *read* path, but: the unprefixed stray term never matches archive queries (`posts_where_filter()` builds `having_terms` from the `cap-`-prefixed term), and the assignment service round-trips the stray slug so its idempotence check reports the name "already assigned" and never attaches the real co-author. So the natural remediation — create the missing guest author, re-run the import — is a no-op, and the post stays off the new author's archive until someone re-saves it. Excluding unresolved names from the `wp_set_post_terms()` list (they're known at the point of detection) would be a smaller diff that fixes the data rather than only reporting it. If keeping the write is deliberate, the description should defend it against those two paths, not just the byline one.

## Smaller items

- The message builds names via `array_keys( $unresolved )` and fields via `array_unique( $unresolved )` as two independent lists, so when `coauthors_post_get_coauthor_by_field` varies the field per name the correspondence is lost (and a numeric-string name like `'2024'` is coerced to an integer key). Emitting per name inside the loop — the action already fires per name — deletes the accumulator and this problem together.
- `coauthors_unresolved_coauthor` isn't added to `docs/filters.md`, even though the notice text points developers at it.
- The action docblock's "The name is still stored as an author term" isn't always true (with `$query_type` `email` the slug is the sanitised name; a name that trims to empty is skipped entirely).
- The hook name is baked into the translatable string; pass it as a `sprintf` placeholder instead (as `class-coauthors-controller.php:412` does) so translators can't alter it.
- Test hygiene: the four new tests inline the same post fixture where `TestCase::create_post()` already exists, and the three hand-rolled action spies capture three different arg subsets — one shared spy (or core's `MockAction`) would cover all four args uniformly.

## Suggested path

1. **Fix resolution tolerance first.** Retry `get_coauthor_by( 'user_nicename', $value )` with `sanitize_title( $value )` before declaring failure. That alone makes `Jane Doe` resolve, silences the append-mode, delete-user and CLI noise, and lets `CoauthorsLinksSingleTest` revert to its original `user_login` inputs. While there, have the two CLI commands pass `'user_login'` as the query type (as `bin/seed-demo-data.php:153` already does).
2. **Decide the storage question explicitly** — I'd exclude unresolved names from the term write, per above.
3. **Reshape the signal.** Keep the action (it's the valuable part); restrict `_doing_it_wrong()` to names the caller actually passed, or drop it in favour of the action plus core's `doing_it_wrong_run`.
4. **Rebuild the tests to discriminate:** restore the reverted template test, add a fixture where `user_login ≠ user_nicename`, add an orphan-term-on-REST-save case, and cover whichever storage decision is made.
5. **Paperwork:** `docs/filters.md`, the docblock, the placeholder.

One last note for the issue rather than this PR: even done perfectly, this signal never fires on the block editor's main write path, because invalid term IDs are dropped by core before `add_coauthors()` runs (see the `sync_coauthors_on_rest_save()` docblock). If #1369 is the target, that gap probably deserves its own issue so it isn't silently half-closed here.

Happy to talk through any of this — the direction is right, it's the resolution mismatch underneath that needs fixing first so the signal can be trusted.

faisalahammad (commented):
Thanks for the detailed review. Pushed 5917f5b with the rework:

- Resolution retry: `get_coauthor_by()` first tries the filtered field, then `user_login`, then `user_nicename` with `sanitize_title()`. The metabox path (login values against the nicename default) and the append-mode re-resolution of plucked logins now resolve instead of firing the signal. `CoauthorsLinksSingleTest` is back to its original `user_login` inputs.
- The two CLI commands that pass logins (`assign-user-to-coauthor`, `swap-coauthors`) now pass `'user_login'` as the query type, like `bin/seed-demo-data.php`.
- Storage: unresolved names are excluded from the `wp_set_post_terms()` list. If that empties the list, the write is skipped so a post keeps its existing terms instead of ending up termless (the #1252 fallback in `sync_coauthors_on_rest_save()` runs before the call, so all-orphan REST saves keep the prior terms).
- `_doing_it_wrong()` is removed. The `coauthors_unresolved_coauthor` action is the only signal, so no notice can corrupt a REST response. The action is documented in `docs/filters.md`, and the docblock no longer claims the name is stored as a term.
- Tests: fixtures for login != nicename (assign and append), an orphan-term REST save case, and assertions that unresolved names leave no term behind.

One note on the action arg: on the REST save path the reported name is the term slug (for example `cap-orphan-guest`), since that is what `sync_coauthors_on_rest_save()` feeds in. 

The block editor write-path gap you flagged (invalid term IDs dropped by core before `add_coauthors()` runs) will get its own issue so it is not lost.

PR #1432, opened 2026-09-06
fix: re-index the Yoast schema graph after filtering Person nodes

### LINKED ISSUE #1360: Yoast integration: filter_graph() can emit @graph as a JSON object instead of an array (array_filter without array_values)
### Summary

`CoAuthors\Integrations\Yoast::filter_graph()` removes every `Person` node from the Yoast schema graph with `array_filter()`, which preserves array keys. When the removed node is not the last element, the PHP array is left with a gap and `json_encode()` serializes `@graph` as a JSON **object** instead of an array:

```json
"@graph": { "0": {...}, "1": {...}, "2": {...}, "3": {...}, "4": {...}, "5": {...}, "7": {...} }
```

A JSON-LD consumer expanding that object discards **every** node in it, because `"0"`, `"1"`, … are neither JSON-LD keywords nor IRIs. The page then has no structured data at all — Article, Recipe/HowTo, Breadcrumbs, Organization, images, all gone — while the markup still looks complete in the source.

### Affected code

https://github.com/Automattic/Co-Authors-Plus/blob/trunk/php/integrations/yoast.php — the `array_filter()` inside `if ( $add_to_graph )`:

```php
$data = array_filter(
    $data,
    function( $piece ) {
        return empty( $piece['@type'] ) || $piece['@type'] !== 'Person';
    }
);
```

Unchanged from 3.7.0 through 4.1.1 (checked against the 4.1.1 release build).

### When it triggers

`$add_to_graph` is set as soon as an article-type node is found, even when `get_coauthors()` produced no resolvable author and `$ids` is empty. In that case:

1. the article node gets `"author": []`,
2. the Yoast-generated `Person` node is removed,
3. `$author_data` is empty, so the `array_merge( $data, $authors )` below — which would re-index as a side effect — never runs.

Yoast's piece order on a post is `0 Article · 1 WebPage · 2 Main_Image · 3 Breadcrumb · 4 Website · 5 Organization · 6 Author (Person) · 7 HowTo/Recipe/block schema`. Removing index 6 with anything behind it produces the gap; if the `Person` node happens to be last, the keys stay contiguous and nothing is noticed. That is why this can sit unnoticed for a long time.

One easy way to end up with an empty author list: a post carries an `author` term whose slug no longer matches any `user_nicename` or guest author — e.g. after a user's nicename was edited — so `get_coauthors()` returns `[]`. The `post_author` fallback does not apply, because it only runs for posts with no author term at all.

### Impact

Measured on two production sites (WordPress multisite, Yoast SEO + CAP): 6,557 published articles resolve to an empty author list, and at least 4,104 of them ship `@graph` as an object and are therefore invisible to structured-data consumers. The front end still renders an author, so nothing looks broken.

### Steps to reproduce

1. Take a post that produces at least one extra graph node after the author — a Yoast HowTo/FAQ block, or a recipe plugin that appends a `Recipe` node.
2. Make `get_coauthors()` return an empty list for it — e.g. attach an `author` term whose slug matches no user and no guest author.
3. View source: `@graph` is a JSON object with a missing numeric key instead of an array.

### Suggested fix

Re-index after filtering:

```php
$data = array_values(
    array_filter(
        $data,
        function( $piece ) {
            return empty( $piece['@type'] ) || $piece['@type'] !== 'Person';
        }
    )
);
```

I'm happy to open a PR for this.

(Somewhat related, same function: #1123.)

faisalahammad:
## Description

`filter_graph()` removed Person nodes from Yoast's schema graph with `array_filter()`, which preserves numeric keys. When the removed node was not the last element, the leftover key gap made `json_encode()` serialize `@graph` as a JSON object instead of an array. JSON-LD consumers expanding that object discard every node in it, because `"0"`, `"1"`, and so on are neither JSON-LD keywords nor IRIs. The page then ships no structured data at all, while the source still looks complete.

Wrapping the `array_filter()` in `array_values()` re-indexes the graph unconditionally, so `@graph` always comes back as a list. This closes both trigger paths from the issue: an empty author list (the `array_merge()` that used to re-index as a side effect only ran when `get_coauthors()` resolved at least one author) and the site-represents-person path that skips the same merge.

Fixes #1360

## Deploy Notes

None. No new dependencies. The Yoast integration's contract with `wpseo_schema_graph` is unchanged: the filter expects an array of schema pieces, and re-indexing restores exactly the list shape Yoast builds the graph with.

## Steps to Test

1. Install Yoast SEO alongside the patched Co-Authors Plus.
2. Pick a published post that produces at least one schema node after the author, such as a Yoast HowTo or FAQ block.
3. Make `get_coauthors()` return an empty list for it. For example, give the post an `author` term whose slug matches no user and no guest author, which happens after a user nicename is edited. The `post_author` fallback does not apply, because it only runs when the post has no author term at all.
4. On the unpatched build, view source and search for `"@graph"`. It is a JSON object with a missing numeric key, for example `"@graph":{"0":{...},"1":{...},"3":{...}}`. Rich Results Test reports no structured data.
5. Apply the patch and reload. `"@graph"` is now a JSON array, and Rich Results Test detects the Article and Person nodes.
6. Confirm a healthy post with working authors is unchanged: same nodes, same order, still a JSON array.

Automated coverage: the new unit test in `tests/Unit/Integrations/YoastFilterGraphTest.php` feeds a graph with a Person in the middle and an empty author list, then asserts the result is a contiguous list and that a JSON round-trip stays an array. It fails on the unpatched build and passes with the patch. Run with `composer test:unit -- --filter YoastFilterGraphTest`.

### DIFF
--- a/php/integrations/yoast.php
+++ b/php/integrations/yoast.php
@@ -214,11 +214,14 @@ class Yoast {
 
 		if ( $add_to_graph ) {
 			// Clean all Persons from the schema, as the user stored as post owner might be incorrectly added if the post has only guest authors as authors.
-			$data = array_filter(
-				$data,
-				function( $piece ) {
-					return empty( $piece['@type'] ) || $piece['@type'] !== 'Person';
-				}
+			// Re-index the result so json_encode() keeps @graph as a JSON array even when a Person node is removed from the middle. See #1360.
+			$data = array_values(
+				array_filter(
+					$data,
+					function( $piece ) {
+						return empty( $piece['@type'] ) || $piece['@type'] !== 'Person';
+					}
+				)
 			);
 
 			if ( ! empty( $author_data ) ) {
--- a/tests/Unit/Integrations/YoastFilterGraphTest.php
+++ b/tests/Unit/Integrations/YoastFilterGraphTest.php
@@ -53,6 +53,8 @@ final class YoastFilterGraphTest extends TestCase {
 
 	const PRESENTATION_CLASS = 'Yoast\\WP\\SEO\\Presentations\\Indexable_Author_Archive_Presentation';
 
+	const SCHEMA_TYPES_CLASS = 'Yoast\\WP\\SEO\\Config\\Schema_Types';
+
 	/**
 	 * Regression coverage for issue #1113.
 	 *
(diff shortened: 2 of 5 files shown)


GaryJones (commented):
Thanks for turning the issue report into a PR so quickly — the fix is exactly what we discussed on #1360, and I've verified it holds up:

- The only consumer of `wpseo_schema_graph` in this plugin registers at priority 11 and the code after the re-index (`array_merge`) is index-agnostic, so re-indexing breaks no caller. Key preservation was never a contract of this filter — the merge path already re-indexed as a side effect.
- The full unit suite passes in both default and reverse order, and the new file is clean against the repo phpcs ruleset.
- The `assertSame( array_values( $result ), $result )` assertion is a correct stand-in for the `array_is_list()` check I suggested, and it does discriminate (the `array_column` assertion alone would not, since `array_column` ignores key gaps).

Three asks on the test file before merge:

1. **The `overload:` mock of `Schema_Types` leaks process-wide** (`mock_schema_types()`). Mockery's `overload:` defines the class for the remainder of the PHP process, and the only protection is the docblock's "only one test in a run may use this". The next test that drives `filter_graph()` past its guards without re-registering the overload gets a mock bound to a closed container — `get_article_type_options()` returns `null` and `array_column()` throws a `TypeError` on PHP 8, but only in full-suite runs, i.e. order-dependent flakiness. Please either add `@runInSeparateProcess` / `@preserveGlobalState disabled` to the test, or replace the overload with a `class_exists()`-guarded plain stub class (the same pattern as the function stubs at the top of the file), which also lifts the one-test-per-run restriction.

2. **Drop the JSON round-trip assertion.** `json_decode( json_encode( $result ) )` being an array is logically equivalent to the `assertSame( array_values( $result ), $result )` check three lines above — both hold iff the keys are a contiguous `0..n-1` list, so the two can only fail together. Keep the `array_values` form; its failure message already names the JSON consequence.

3. **Hoist the duplicated helpers.** `stub_get_coauthors_dependencies()` and the `tear_down()` global cleanup are byte-identical copies of the private helpers in `YoastPresentationPostTest.php`. Please move the pair into the shared `tests/Unit` `TestCase` (or a small trait) and call it from both files — the stub encodes assumptions about `get_coauthors()` internals that must now be updated in lockstep, and the seed/cleanup pair only travels together by convention.

Two things I'm noting for follow-up issues rather than this PR:

- The re-index fixes the serialisation shape, which is the right scope here, but the zero-resolved-author path still blanks the Article's `author` and strips Yoast's valid `Person` node, leaving dangling `@id` references (`WebPage`/`publisher` still point at the removed `#/schema/person/…`). The deeper fix is probably an early return of `$data` untouched when `$ids` is empty, before any mutation — worth knowing that the `assertSame( array(), $result[0]['author'] )` assertion in this test would need revisiting then, as it documents today's degraded output rather than desired behaviour.
- Pre-existing and adjacent: the merge gate reads loop leftovers — `! empty( $author_data )` and `$author->ID` reflect only the last co-author, and `$author_data` is never reset per iteration — so a multi-author post whose last co-author fails to resolve (or matches the site user on a site-represents-person site) drops or duplicates author nodes. Not this PR's doing; I'll raise it separately.

Issue #1434, opened 2026-09-07
REST author formatting writes terms on every read


GaryJones:
While reviewing #1341 I said I would raise this separately.

`Endpoints::_format_author_data()` (`php/class-coauthors-endpoint.php:279-284`) calls `CoAuthors_Plus::update_author_term()` as its first step. That method runs `wp_insert_term()` / `wp_update_term()` as needed, so every route that formats an author — the GET search and authors routes among them — can write to the terms table during what a caller reasonably assumes is a read.

Two things make this more than a tidy-up:

- **The write is load-bearing.** The formatter uses the returned term to decide whether the row is included at all (`if ( ! $term || is_wp_error( $term ) ) return null;`) and to populate `termId`. Simply deleting the call changes which authors appear in results — an author with no term yet would vanish where today the read path quietly creates their term and includes them. Any fix has to decide whether read routes should backfill missing terms (current behaviour, surprising), skip such authors (a behaviour change), or read without repairing (`get_author_term()` and tolerate a null `termId`).
- **The newer read-only controller already made the opposite choice.** `php/api/endpoints/class-coauthors-controller.php` deliberately reads without writing, so the plugin currently answers the question both ways depending on which endpoint you hit.

All these routes sit behind an `edit_posts`-level `permission_callback`, so this is not unauthenticated write exposure — it is a consistency, performance (a write plus cache invalidation per formatted author per request) and least-surprise problem.

PR #1436, opened 2026-09-07
Stop refreshing author terms on REST GET requests

GaryJones:
## Summary

Addresses the steady-state part of #1434.

`Endpoints::_format_author_data()` called `CoAuthors_Plus::update_author_term()` for every author it formatted, so the GET `search`, `authors` and `authors-by-term-ids` routes wrote to the terms table during what callers reasonably assume is a read. Even when the term description was already up to date, the call still deleted and re-primed the author term cache for every author on every request. The refresh is also redundant these days: term description freshness has been owned by the profile-update hook (`update_author_term_on_profile_update()`) since 4.1.0.

The formatter is now a pure read via `get_author_term()` — it can neither create nor refresh a term, so no current or future route can inherit write-on-read behaviour from it. That's safe because on almost every path a term exists by construction: search results are found *via* `get_terms` on the co-author taxonomy, the by-term-ids route resolves the term before the author, and the POST route creates terms in `add_coauthors()`.

Exactly one path can produce a term-less author: the `post_author` fallback in `get_coauthors()` on a post saved before Co-Authors Plus. That author cannot simply be skipped — the editor persists selections by term id (`set_post_author_for_rest_save()` resolves them with `get_term()`), so omitting them from the response means the next editor save silently drops the original author. `_build_authors_response()` therefore backfills a missing term explicitly, at route level, scoped to that single case: a one-time write per legacy author, after which every read is pure. `search_authors()` needed no change, as it already follows the read-first pattern.

Three integration tests pin the behaviour: formatting an author whose term has a stale description leaves the description untouched, the formatter creates no term for a term-less author (both fail against the previous code), and the authors route backfills a term for a legacy post's `post_author` and includes them in the response.

This does not yet make the read routes literally write-free — the legacy backfill remains, and `search_authors()` still creates terms for users who merely appear in results. Moving those writes fully out of GET (selection-time creation, CLI healing for legacy posts) is the remaining scope on #1434.

## Test plan

- [ ] `wp-env run tests-cli --env-cwd=wp-content/plugins/co-authors-plus ./vendor/bin/phpunit --testsuite integration --filter EndpointsTest` passes (16 tests)
- [ ] With an author whose term description is stale, a GET to `coauthors/v1/search` no longer rewrites the term description

### DIFF
--- a/php/class-coauthors-endpoint.php
+++ b/php/class-coauthors-endpoint.php
@@ -273,11 +273,22 @@ class Endpoints {
 	 * so we exclude it from the response rather than feed the editor data it
 	 * can't round-trip.
 	 *
+	 * Reads the existing author term rather than refreshing it: the editor
+	 * saves selections by term id, so an author must have a term to be
+	 * selectable at all, and only a missing term is backfilled (a one-time
+	 * write per author, e.g. the post_author fallback on a pre-CAP post).
+	 * Description freshness is owned by the profile-update hook, not by GET
+	 * requests.
+	 *
 	 * @param object $author The result from co-authors methods.
 	 * @return array|null
 	 */
 	public function _format_author_data( $author ): ?array {
-		$term = $this->coauthors->update_author_term( $author );
+		$term = $this->coauthors->get_author_term( $author );
+
+		if ( ! $term ) {
+			$term = $this->coauthors->update_author_term( $author );
+		}
 
 		if ( ! $term || is_wp_error( $term ) ) {
 			return null;
--- a/php/class-coauthors-endpoint.php
+++ b/php/class-coauthors-endpoint.php
@@ -273,22 +273,16 @@ class Endpoints {
 	 * so we exclude it from the response rather than feed the editor data it
 	 * can't round-trip.
 	 *
-	 * Reads the existing author term rather than refreshing it: the editor
-	 * saves selections by term id, so an author must have a term to be
-	 * selectable at all, and only a missing term is backfilled (a one-time
-	 * write per author, e.g. the post_author fallback on a pre-CAP post).
-	 * Description freshness is owned by the profile-update hook, not by GET
-	 * requests.
+	 * This is a pure read: it never creates or refreshes the term. Description
+	 * freshness is owned by the profile-update hook, and the one caller that
+	 * can legitimately see a term-less author backfills before formatting
+	 * (see _build_authors_response()).
 	 *
 	 * @param object $author The result from co-authors methods.
 	 * @return array|null
 	 */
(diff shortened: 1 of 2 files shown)


faisalahammad (commented):
Reviewed against the issue and the surrounding code. The core move is right: the formatter no longer writes, the legacy backfill is explicit and at route level, and the three new tests genuinely fail against the previous code. CI is green. A few notes, none blocking.

**1. The title and summary overclaim the scope.** The title says "Stop refreshing author terms on REST GET requests", but the GET `search` route still refreshes them. `search_authors()` (`php/class-coauthors-plus.php:2181-2186`) calls `update_author_term()` for any result whose term is missing or has an empty description, and that is a write plus a cache flush per matching user. The remaining-scope paragraph acknowledges the creation case, but the summary line above it says `search_authors()` "already follows the read-first pattern", which reads as if it is write-free. Read-first is not read-only here, because the miss branch inserts and the empty-description branch updates. Suggest tightening that one sentence so the description matches what the diff actually does.

**2. `is_wp_error()` in the formatter is now dead code.** `_format_author_data()` now takes its term from `get_author_term()`, which can only return a `WP_Term`, `false`, or `null` (from `get_term_by()`). The `WP_Error` path belonged to `update_author_term()` and can no longer reach that guard. Harmless, but the condition is misleading now. Could drop it, or keep it with a short comment if you want the guard to survive a future return-type change.

**3. Losing read-time description healing is the real trade-off, and it looks acceptable.** After this change a drifted term description is only corrected by `profile_update`. That covers the normal paths, since `wp_update_user()` and `edit_user()` both fire `profile_update` and `update_author_term_on_profile_update()` refreshes an existing term. But the formatter docblock says freshness is "owned by the profile-update hook" without saying what happens when a description drifts some other way (a direct `wp_update_term()`, a plugin writing user meta, an import). Those now stay stale until the profile is next saved. If that is intended, fine, though it is the one behaviour change beyond the write-on-read fix and it may be worth naming in the comment.

**4. Test coverage.** The three new tests are well aimed. Two small gaps: nothing pins the `search` route's continued writes, so if `search_authors()` is ever made read-only the "remaining scope" claim has no test to flip; and the by-term-ids route now runs through the pure formatter but that connection is only visible in `RestCoauthorSyncTest::test_authors_by_term_ids_skips_terms_with_no_resolvable_author`, which is easy to miss.

**5. Minor.** The backfill in `_build_authors_response()` also runs on the POST path. That is harmless (terms exist there via `add_coauthors()`, and the `get_author_term()` guard makes it a cached no-op), but the comment says "the one term-less case", which is about reachability rather than where the code runs.

Item 1 is the one I would change before merge. The rest are optional.

